PR #11, opened 2016-03-19
Added #getFriends()

mc-frank:
Returns List<User> of people the user is friends with


BtoBastian:
Hey,
thank you for contributing, but there are some problems with your pull request:
1. I'm still not sure if I will add friend support. Discord devs said they don't want us to add client-only features
2. You create a new user object every time. I'll try to make most object unique so you can compare two users using the '==' operator
3. Empty catch block
4. I've got some strict code conventions. I haven't published them so this is my fault. They are very similar to Google's Java Conventions but with max line length 120 characters instead of 80.

Issue #16, opened 2016-09-06
'Timestamp' attribute on messages ignores the timezone.


me4502:
Currently the timestamp attribute on messages completely ignores the timezone, which causes major issues when comparing the time. As no TZ is supplied, Java assumes it wants to use the system time, which means unless the system is set to use UTC, the time will be wrong.

This can be remedied in one of two ways:
* Just set the calendar instance to UTC
* Or better, implement proper timestamp handling that includes the timezone.

BtoBastian:
https://github.com/BtoBastian/Javacord/commit/a472914c847a45dcc40b268a763fcd1322f6ad09
Thank you very much!

PR #17, opened 2016-09-06
Time Zone Fix for Messages

Bourne-ID:
Issue #16 fix, now including TimeZone parsing. Using Regexp and Google Joiner to remove the nano seconds and make it milli seconds (for Java 7) . Test completed on quick tests + my bot.


Bourne-ID:
I should also add I've enhanced the way it parses the timestamp - before a synchronous block would cause potential slowdown and locking issues. Given the multithreaded nature, I've added in ThreadLocals, so the simpledateformats are created as needed.

PR #24, opened 2016-09-26
Fix ImplServer#banUser(String, int)

Mysterypotatoguy:
Fixed an issue where banUser wasn't getting the guild id properly


BtoBastian:
Thanks :)

PR #26, opened 2016-10-05
Double message delete event

Dunciboy:
The deleted code caused a double delete event, meaning that the same message got 2 delete events. In the event you would want the bot to delete a message after for example posting a command it will cause a double call on the delete event causing the message signalled as deleted 2 times.

Compiled and tested on a bot that logs deleted messages. After removing said code the double logging of the deleted message no longer occurred.

I suggest you make something that listens for the deletion of said message the current code just triggers a delete event by itself.


Dunciboy:
Update: Added check to see if the message was actually deleted.
I plan on doing the same thing for the for the function edit since it had the same issues as the delete function. So please wait with merging until my next commit.

Dunciboy:
Update: Checks are now in place ready for merge if you agree with the checks for the changes in the buffer.

PR #27, opened 2016-10-05
Fixed double call on listener

Dunciboy:
The removed script parts caused double triggers when a message was
deleted, if for example you would have used delete to delete a message
the listener would be called twice for the same message.
Also changed variable logger to LOGGER since by convention rules a
static final should be in full caps.

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/message/impl/ImplMessage.java
+++ b/src/main/java/de/btobastian/javacord/entities/message/impl/ImplMessage.java
@@ -32,8 +32,6 @@ import de.btobastian.javacord.entities.impl.ImplUser;
 import de.btobastian.javacord.entities.message.Message;
 import de.btobastian.javacord.entities.message.MessageAttachment;
 import de.btobastian.javacord.entities.message.MessageReceiver;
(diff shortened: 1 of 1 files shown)


BtoBastian:
Hello, 
thank you for the pull request.

Nevertheless the variable logger was intended to be in "Lower Camel Case". Javacord is using the Google Style Guide which says that only "real" constants should be in "Upper Case".
To quote the Guide: " Every constant is a static final field, but not all static final fields are constants. "
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

Dunciboy:
So you want me to revert the variable name change correct?

BtoBastian:
Would be nice if you could. :)
Otherwise I would merge the pull request and change it afterwards.

BtoBastian:
Hey,

this commit fixed the double call. It didn't remove the listener call in the delete() method but now checks in the handler if the message was already deleted.
https://github.com/BtoBastian/Javacord/commit/4e7bea2c9adecdf9415b88abec57d21ca0c4d1a9

PR #46, opened 2017-01-29
Audio Support

567legodude:
Basic structure for audio.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,20 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.22</version>
         </dependency>
+
+        <!-- Opus -->
+        <dependency>
+            <groupId>net.tomp2p</groupId>
+            <artifactId>opus-wrapper</artifactId>
+            <version>1.3</version>
+        </dependency>
+
+        <!-- Tritonus -->
+        <dependency>
+            <groupId>com.googlecode.soundlibs</groupId>
+            <artifactId>tritonus-all</artifactId>
+            <version>0.3.7.2</version>
+        </dependency>
     </dependencies>
 
     <build>
@@ -75,6 +89,14 @@
                         </goals>
                     </execution>
                 </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                </configuration>
             </plugin>
         </plugins>
 
--- a/src/main/java/de/btobastian/javacord/DiscordAPI.java
+++ b/src/main/java/de/btobastian/javacord/DiscordAPI.java
@@ -26,6 +26,7 @@ import de.btobastian.javacord.entities.permissions.PermissionsBuilder;
 import de.btobastian.javacord.listener.Listener;
 import de.btobastian.javacord.utils.ThreadPool;
 import de.btobastian.javacord.utils.ratelimits.RateLimitManager;
+import org.apache.http.HttpHost;
 
 import java.awt.image.BufferedImage;
 import java.util.Collection;
@@ -144,6 +145,8 @@ public interface DiscordAPI {
      * @return The voice channel with the given id. <code>Null</code> if no channel with the id was found.
      */
     public VoiceChannel getVoiceChannelById(String id);
+    
+    public boolean isAudioEnabled();
 
     /**
      * Gets an user by its id. It first will check if the user is in the cache. If no user was found in the cache it
@@ -498,5 +501,7 @@ public interface DiscordAPI {
      * After disconnecting you should NOT use this instance again.
      */
     public void disconnect();
+    
+    public void setAudioProxy(HttpHost host);
 
 }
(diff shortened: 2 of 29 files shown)


DV8FromTheWorld:
Yeah, I just want to point out that almost every line of code in this is copy-pasted from JDA with a few changes to make it work with Javacord. While JDA is an open source project, I don't really appreciate the blatent copying, especially considering you ripped the copyright headers off of all of these files.

567legodude:
I did use the same setup as JDA and that's where I learned how to connect to the voice servers. I did notice that it was looking oddly the same while writing it, but I never copy/pasted any code.

austinv11:
I'm also curious about your implementation of pojos for interfacing with the Discord api. Javacord doesn't do it anywhere else but you implemented it for this PR. Discord4J does it for voice [here](https://github.com/austinv11/Discord4J/tree/master/src/main/java/sx/blah/discord/api/internal/json/requests/voice) and yours seem [very similar](https://github.com/567legodude/Javacord/tree/057a171b9172f8c6e07be8aecb8854e45462e5ea/src/main/java/de/btobastian/javacord/utils/payload) to mine. Now you could argue that D4J also "copied" JDA too but the difference is that we got explicit permission and we [still kept license headers for assets we used](https://github.com/austinv11/Discord4J/blob/master/src/main/java/sx/blah/discord/api/internal/ServiceUtil.java#L1-L15), plus we are rewriting the internals anyway.

DV8FromTheWorld:
Please, don't lie to me.
There are significant portions of this code which are identical to subsystems in JDA, many of which only exist to provide specialized functionality.
One of which is the `PacketProvider`. I did extensive reworking of the audio sending systems to abstract the UDP sending portion of the audio system to allow for custom implementations. One such example of an implementation would be https://github.com/sedmelluq/jda-nas. Sedmelluq specifically asked me to rework this system and PacketProvider is an object provided to an AudioSendFactory to construct AudioSendSystems (You've renamed to AudioSender). It makes no sense that you would "come up with" this implementation as it is not the straightforward implementation.

567legodude:
I never said I made it myself. I already said that JDA's structure is what I based mine off of. But I'm not lying when I say I never directly copied anything, if I did that then there's no challenge.

and austinv11
I don't know what pojos is. Are you talking about the Payload object for sending JSON? The only reason I made that was because I could just send(new PayloadType(params)) rather than writing out the entire JSON structure.

567legodude:
I would be perfectly fine with reworking the structure if that's what you want. I didn't intend for it to be so similar.

DV8FromTheWorld:
Say what you want, but it is quite obvious the level of copying here.

567legodude:
But you did the good thing, you brought up the fact that it looks similar, and I realize this and I am currently working on completely redoing it.

shawishi:
I guess the real problem here is that you copied their code and didn't give them credit...
Reworking it would just be silly.
Why not just give DV8 their due credit and get it over with?

567legodude:
If that's a solution that works, then let's do that.

I did already get most of the way making a system that uses a FrameIterator instead of a provider service, and used a different setup where I can connect with one class instead of two. If that works I'll just replace it with that.

BtoBastian:
Hey, I already talked to the author of the PR yesterday about this. I don't won't to use any source code of JDA, Discord4J or any other library without their permission and without crediting them. However, I really appreciate 567legodude making the pull requests as it gives me a good starting point working on voice. I didn't compare the source yet, so I don't know how similar it is to JDA.
@DV8FromTheWorld Would you be ok with it, if I use this as a basis? I'm sure I'll change a lot of the current code and ofc give credit to you/JDA. Is there any particular class you don't want to be used by me? Or are you completely against this PR?
@austinv11 Beside the requests split up into different classes, is there anything else "inspired" by Disvord4J? I don't think I will use this system in the final code anyway, cause that's a completely different style then the rest of the Javacord source. (That's how I did it ~1 year ago when I started with voice but never finished it: https://github.com/BtoBastian/Javacord/blob/voice/src/main/java/de/btobastian/javacord/utils/DiscordVoiceWebsocket.java#L129). I pretty certain, that it will look similar to my old code when finished. However, I would be glad to have your permission to accept this PR, too. If there is any part you don't like in this PR or anything you want credit for I will replace it/credit you.

austinv11:
@BtoBastian There are some various similarities with the abstraction with D4J, however not enough for me to have the right to definitively say that its copied from D4J. And @DV8FromTheWorld's opinion is way more important than mine.

DV8FromTheWorld:
I would be much more okay with using the audio system in JDA 2.x (legacy) as apposed to this PR. Legacy's audio system is much more straight forward in its implementation and is what D4J's audio system is based on. 3.x's modified audio system came about after extensive redesign providing automatic reconnects, connection monitoring, and the ability to provide custom implementations of certain parts of the audio system.
This PR removes most if not all of the forward facing features introduced in 3.x but keeps the internal objects used in 3.x's redesign. As such, I'd much prefer to see an audio system based off of (read: not copied) from JDA 2.x.

I would also be happy to field questions related to the legacy system.

In the end, it is up to you, but I feel you will find the legacy system to be more straightforward in its implementation and is a better place to start at for your own customization.

BtoBastian:
Thank you for your answer. I'll take a closer look at the audio system in JDA 2 and this pull request next week. :-)

567legodude:
Don't worry DV8, I kept my promise and have remade the entire thing. I went simple, just a single class for all things related to connections (I think 3 classes is a bit much) and made a FrameIterator to iterate the audio frames. Nothing fancy like reconnect or receiving audio.

Camouflage100:
~8 months later~

PR #49, opened 2017-04-02
Add support for editing a roles mentionable and managed values

kenwiel:
I'm not sure if I did everything correctly since I'm still a beginner, but this should add support for making a Role mentionable.
I'm not sure what the managed value is used for/by but since it's also a boolean I decided to add that in as well.

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/permissions/Role.java
+++ b/src/main/java/de/btobastian/javacord/entities/permissions/Role.java
@@ -128,7 +128,7 @@ public interface Role {
     /**
      * Updates the permissions of this role.
      * If you want to update other things like name, color, etc. use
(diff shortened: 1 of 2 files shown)


BtoBastian:
Thank you very much. 👍

williamwebb:
This MR seems to have broken the build. Add a reference to a `RoleChangeManagedListener` wich does not seem to exist anywhere, as well as a few other compile errors.

BtoBastian:
Ohh, I didn't want to merge it into the master branch.
Overlooked it, my bad ._.

BtoBastian:
@kenwiel Can you please take a look at the problem mentioned by @williamwebb ?
Like he said, you use a RoleChangeManagedListener, which doesn't even exist.

PR #53, opened 2017-05-25
Implemented text and voice channels updatePosition() calls.

NXTk:
Implemented text and voice channels updatePosition() calls, added and changed update() accordingly.
Channels with same position value sorted by id, depends on client.


BtoBastian:
Great, thank you! :-)

PR #60, opened 2017-07-21
Pinning messages.

GamingAndStuffs:
Couldn't find a way to pin messages, so I pretty much mashed together the delete and edit methods to make a pin method.

Tested with 1 bot on 1 server so there may be issues with it, but none that I know of.


BtoBastian:
Thank you!

PR #62, opened 2017-07-29
Fix icon related handling, update some documentation

Mysterypotatoguy:
Old code wasn't handling the case of icon being null, leading to issues such as [this](https://pastebin.com/J6PT6WzU) .


BtoBastian:
Thank you!

Issue #66, opened 2017-08-05
API for getting users in a voice channel


MedicOP:
Could API be added for getting users in a voice channel?

There is ImplVoiceChannel.getConnectedUsers() but it's not in the API

BtoBastian:
Thank's to TheDiamondPicks, this will be part of the next version!

PR #68, opened 2017-08-05
Added API method to get connected users to a VoiceChannel

TheDiamondPicks:
Pretty simple - just added the existing getConnectedUsers from ImplVoiceChannel method into the VoiceChannel interface.

Fixes Issue #66 

Tested this with a simple for each loop - seems to work perfectly.


BtoBastian:
Thank you

PR #78, opened 2017-08-20
Fix User#isYourself

shawishi:
User#isYourself doesn't always work as it compares references rather than calling Object#equals

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/impl/ImplUser.java
+++ b/src/main/java/de/btobastian/javacord/entities/impl/ImplUser.java
@@ -155,7 +155,7 @@ public class ImplUser implements User {
 
     @Override
     public boolean isYourself() {
-        return api.getYourself() == this;
+        return this.equals(api.getYourself());
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


BtoBastian:
In Javacord user objects are unique. There should never exist two user objects for the same user, therefore the method should work fine.

shawishi:
Except when loaded externally for when the library fails to update certain fields/attributes/properties of the object which is a problem I'd actually faced (i.e. I had to manually request an up-to-date user object from Discord's API and rebuild Javacord's object instead of fetching the old out-of-date cached one).
One could argue that starting a pull request that patches up all these inconsistencies that could occur between the cached object and the actual user object as is on Discord itself would make more sense, but this for now ensures that if in case Javacord falls behind, it won't cause the user's application to experience unexpected behavior when attempting to call said method with what is expected to be (and in fact is) the application/bot's own user object.

Regardless, generally speaking, what's currently implemented is considered bad practice, especially so in this case because it only guarantees proper functionality under specific usage conditions.

BtoBastian:
If you're using internal methods/constructors you can expect unexpected behavior. The library was designed with unique objects in mind and this is most likely not the only problem it's creating. If there are inconsistencies in the cache, please create an issue and I'll fix it, but I can't fix errors I don't know of. Having a wrong cache is a way bigger problem imo.

I would also disagree, that using the == operator is bad practice. It just depends on what you are trying to achieve and if both `==` and `#equals()` fullfil the same purpose, I would always choose `==`.

PR #89, opened 2017-09-11
Added listeners for all currently implemented events.

kenwiel:
Added listeners for all currently implemented events.

PR #90, opened 2017-09-14
Added api#getYourself()

GamingAndStuffs:
Self-Explanatory
Tested in Javacord server.

### DIFF
--- a/src/main/java/de/btobastian/javacord/DiscordApi.java
+++ b/src/main/java/de/btobastian/javacord/DiscordApi.java
@@ -452,4 +452,16 @@ public interface DiscordApi {
      */
     List<UserStartTypingListener> getUserStartTypingListeners();
 
+    /**
+     * Gets yourself
+     *
+     * @return Yourself.
+     */
+    Optional<User> getYourself();
+
+    /**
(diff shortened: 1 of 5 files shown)


BtoBastian:
Hey, thank you. :-)
There are just two thing which need changes:
- Why does `getYourself()` return an `Optional<User>` instead of just the user? It's set in the Ready packet and should therefore never be not present
- `setYourself()` is an internal method and should not be part of the `DiscordApi` interface, only in the implementation

BtoBastian:
Please also notice the order of the methods, e.g. the listener methods (`addXyzListener` and `getXyzListeners`) are always the last methods.
As there's no contribution guide at the moment, I can't really recommend PR's to the v_3. The rewrite is still in a very early stage.
I really appreciate everyone trying to help, but right now it's better to ask me where I need help (e.g. like kenwiel with #88 and #89) instead of just creating PR's for random missing features.

GamingAndStuffs:
I'll fix it as soon as I can, and sorry for the seemingly random feature, it's just the last one that was needed for sdcf4j to be fully supported.

PR #93, opened 2017-09-22
Changed return null to return Optional.empty()

kenwiel:
IDE was complaining about it ¯\\\_(ツ)_/¯


BtoBastian:
I was unsure what to use, as this is something which in theory should never happen.
Can you please change it in `Server#getIcon()` and `GroupChannel#getIcon()`, too. :-)
Ty

PR #110, opened 2017-12-19
Added overloaded methods in Messageable for sending BufferedImages

BoneyIsSpooky:
I added overloaded methods in `Messageable` for converting `BufferedImage` to `InputStream`, and sending them as in `Messageable#sendMessage`. This allows for more seamless functionality with the rest of the library. For example, methods such as `Icon#asBufferedImage` return `BufferedImage` and it is now possible to directly pass this output to `Messageable#sendMessage`.

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/message/Messageable.java
+++ b/src/main/java/de/btobastian/javacord/entities/message/Messageable.java
@@ -2,10 +2,10 @@ package de.btobastian.javacord.entities.message;
 
 import de.btobastian.javacord.entities.message.embed.EmbedBuilder;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


BtoBastian:
Hey,
thank you for the PR and sorry for the late answer, I wanted to take some time to think about it.
I'm not sure, if it's such a good idea to accept `BufferedImage` as a parameter, too.
However, what seems like a good idea to me, is to extend the `Icon` class with a `getInputStream()` method, which can be used for the `sendMessage(...)` method. What do you think about this solution?

BoneyIsSpooky:
I think that is a better solution than my original idea. I will write this and make a new PR.

BtoBastian:
`I will write this and make a new PR.`
I've already created a `Icon#asInputStream()` method, however, it seems like the method to send files is broken atm (most likely since the change to okHttp).

PR #112, opened 2017-12-21
Added support for managing roles

GamingAndStuffs:
Added support for updating a user's roles, and added a builder to add and remove certain roles from users.


BtoBastian:
Great, thank you!

PR #116, opened 2017-12-28
Renamed DiscordWebsocketAdapter to DiscordWebSocketAdapter

Vampire:
This is more consistent with official naming in the spec, with the namings in the used WS lib and even with other parts in Javacord like `getWebSocketAdapter()`


BtoBastian:
👍

PR #118, opened 2017-12-28
Added trace-level logging for OkHttpClient calls and WebSocket events

Vampire:
With the old http lib there already was logging of this kind included on debug level.
With this PR the OkHttpClient calls are logged on trace level.
The WebSocket events are also logged on trace level.
With this logging enabled it is much easier to see what is going on and diagnose deeper-level problems.


BtoBastian:
Looks fine :)

PR #119, opened 2017-12-28
Make HTTP responses usable when retrieved from an exception

Vampire:
OkHttp responses (or response bodies) should be closed to prevent memory leaks.
This is done already implicitly as always `new RestRequestResult(this, response)` is called which calls `response.body().string()` which in turn implicitly closes the response body and thus also the response.
This though also makes it impossible to read the body of the result again, e. g. in an exception handler that wants to react differently according to what discord sent back exactly.

With this PR the `AutoClosable` property of the response is used to ensure explicitly that all requests are properly closed.
Besides that, the custom exceptions that previously had an optional `Request` and an optional `Response`, now have an optional `RestRequestResult` instead of the `Response` where the body of the response can be retrieved as well as the response itself to be checked for other properties like status code and so on.

### DIFF
--- a/src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
+++ b/src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
@@ -324,15 +324,17 @@ public class RestRequest<T> {
         logger.debug("Trying to send {} request to {}{}",
                 method.name(), endpoint.getFullUrl(urlParameters), body != null ? " with body " + body : "");
 
-        try(Response response = getApi().getHttpClient().newCall(requestBuilder.build()).execute()) {
+        try (Response response = getApi().getHttpClient().newCall(requestBuilder.build()).execute()) {
             RestRequestResult result = new RestRequestResult(this, response);
             logger.debug("Sent {} request to {} and received status code {} with{} body{}",
                     method.name(), endpoint.getFullUrl(urlParameters), response.code(),
-                    result.getBody().map(b -> "").orElse(" empty"), result.getStringBody().map(s -> " " + s).orElse(""));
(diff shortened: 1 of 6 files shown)


src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
```diff
@@ -324,38 +324,39 @@ public RestRequestResult executeBlocking() throws Exception {
...
-                result.getBody().map(b -> "").orElse(" empty"), result.getStringBody().map(s -> " " + s).orElse(""));
-        if (response.code() >= 300 || response.code() < 200) {
-            if (!result.getJsonBody().isNull() && result.getJsonBody().has("code")) {
-                int code = result.getJsonBody().get("code").asInt();
-                String message = result.getJsonBody().has("message") ? result.getJsonBody().get("message").asText() : null;
-                switch (code) {
-                    case 50007:
-                        throw new CannotMessageUserException(origin,
-                                message == null ? "Cannot send message to this user" : message, response, this);
+        try(Response response = getApi().getHttpClient().newCall(requestBuilder.build()).execute()) {
```
BtoBastian: The missing space between `try(...)` is bothering me more than it honestly should.. ^^

BtoBastian:
Thank you. 👍

src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
Vampire: Very understandable for me, don't worry. :-) Maybe you should check in idea project files that are meant for being checked in (most stuff except workspace.xml) or create an `.editorconfig` file if that supports enough of your rules, I don't know what it supports exactly, but it is cross IDE code style settings.

PR #121, opened 2017-12-31
Fixed the ratelimit handling for non-ratelimited endpoints

Vampire:
At non-ratelimited endpoints like the `/gateway` one (**not** `/gateway/bot`), there are no ratelimiting headers set, but the ratelimit manager assumed that `X-RateLimit-Reset` is always set if no hard-coded ratelimit is set. Now `0` is used as default value so that there is never a delay or wait time.


BtoBastian:
Was this causing a NPE?

BtoBastian:
Ok, nevermind.
https://github.com/BtoBastian/Javacord/pull/120 explained it

PR #122, opened 2017-12-31
Improved the Gateway URL handling so that the non-ratelimited endpoint is used and the URL is only retrieved once per process or if (re-)connection failed

Vampire:
This PR improves the Gateway URL handling in three points:

1. The non-ratelimited `/gateway` endpoint is used. `/gateway/bot` is only necessary if you want to get the recommended shard count, the URL will always be the same in both requests. This detail was just lost in the API docs during a docu rewrite. `/gateway/bot` is limited to twice in a certain amount of time (not sure which amount exactly), without any major parameters, so being effective over all authenticated calls to this endpoint. If you e. g. start three separate processes to handle three shards and each calls the `/gateway/bot` endpoint, at least the third will currently get a ratelimited reply and needs to wait. Using the `/gateway` endpoint does not have this problem, as it is not ratelimited. It is not even authenticated, so a ratelimiting would not be possible easily anyway.

2. & 3. Another fact that was lost during the docu rewrite is, that the `/gateway` or `/gateway/bot` endpoint should only be called once per process that probably manages multiple shards and the value for the gateway URL should be cached (2.) as long as no connection error happens. If a connection error happens, the URL might be reretrieved (3.) to make sure it did not change. This PR also takes care about these two points which were not done before. Up to now each shard (instance of `ImplDiscordApi`) did retrieve the gateway URL itself and never updated it if a connection error happened.

**Note:** When in future auto-sharding will be possible, then the auto-sharding functionality of course has to call the `/gateway/bot` endpoint to get the recommended shard count, but then also the URL should be taken from this response and given to the single shards to be used until a connection failure happens. At that point, the URL should be reretrieved which can be done via the `/gateway` endpoint of course if not the whole shard-farm is rebooted.

### DIFF
--- a/src/main/java/de/btobastian/javacord/ImplDiscordApi.java
+++ b/src/main/java/de/btobastian/javacord/ImplDiscordApi.java
@@ -37,9 +37,6 @@ import de.btobastian.javacord.utils.ListenerManager;
 import de.btobastian.javacord.utils.ThreadPool;
 import de.btobastian.javacord.utils.logging.LoggerUtil;
 import de.btobastian.javacord.utils.ratelimits.RatelimitManager;
-import de.btobastian.javacord.utils.rest.RestEndpoint;
-import de.btobastian.javacord.utils.rest.RestMethod;
-import de.btobastian.javacord.utils.rest.RestRequest;
 import okhttp3.OkHttpClient;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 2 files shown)


Vampire:
This of course even works if you start multiple separate bots from one process (whyever one wants to do that, as the URL should be the same for all).

src/main/java/de/btobastian/javacord/ImplDiscordApi.java
```diff
@@ -231,46 +231,31 @@ public ImplDiscordApi(
...
-                                    } else {
-                                        clientId = applicationInfo.getClientId();
-                                        ownerId = applicationInfo.getOwnerId();
-                                    }
-                                    ready.complete(this);
-                                });
-                            } else {
-                                ready.complete(this);
-                            }
+        this.websocketAdapter = new DiscordWebSocketAdapter(this);
```
BtoBastian: With your changes, the constructor of DiscordWebSocketAdapter is now blocking, causing the login method also being blocking as it waits for the gateway being received.

Vampire: Hm, you're right, that is bad.
What would fit best into Javacord design?
Running a new thread with the code?
Using `getThreadPool().getExecutorService().submit()` to run the code?
Keeping the constructor synchronously but using `websocket.connectAsynchronously()` and changing the `onConnectError()` handler to also do the retry logic instead of just logging?

Vampire: Ok, for now I used the executor service of the thread pool.

PR #123, opened 2018-01-07
Removed unused private field

Vampire:
I didn't see this field used anywhere, just updated, so I removed it.


BtoBastian:
Seems like a remnant from an older version.
I tried to use the heartbeat ack in the past to detect connection loses, but removed it because it produced lots of false positives.

PR #125, opened 2018-01-07
Improve Identify Ratelimit Handling

Vampire:
When hitting the once-every-five-seconds identify ratelimit, the API sends back an invalid session response like when trying to resume an invalid session.
Javacord did always log that resuming did fail and waited five seconds before retrying.
This is not too correct, because after the resuming failed you can immediately do the normal login instead and the log message is plainly wrong as a resume was not even tried.

Now it is remembered whether the last sent frame was an identify frame. If not and the session invalid response arrives immediately an identify is tried and the resume-failed message logged, if yes, the next identify call is delayed for 5.1 seconds and the log correctly complains about the hit identify ratelimit.

Furthermore it is now tried to actively prevent to hit that ratelimit by only starting a connection for the same token and thus account if the last identify frame was sent more than 5.1 seconds ago and otherwise this time is waited. This works as long as the identifies happen in the same JVM. If the identifies happen in different JVMs, e. g. for different shards, then of course it cannot be prevented and the handling mentioned above kicks in.

### DIFF
--- a/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
+++ b/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
@@ -352,7 +352,7 @@ public class DiscordWebSocketAdapter extends WebSocketAdapter {
                     sendIdentify(websocket);
                 } else {
                     // Invalid session :(
-                    int zeroToFourSeconds = new Random().nextInt(4000);
+                    int zeroToFourSeconds = (int) (Math.random() * 4000);
                     logger.info("Could not resume session. Reconnecting in {}.{} seconds...",
                                 1 + zeroToFourSeconds / 1000,
                                 1 + zeroToFourSeconds / 100 % 10);
--- a/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
+++ b/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
@@ -48,12 +48,16 @@ import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicMarkableReference;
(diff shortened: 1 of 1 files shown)


src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
```diff
@@ -311,10 +346,15 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
...
-                logger.info("Could not resume session. Reconnecting in 5 seconds...");
-                Thread.sleep(5000);
-                sendIdentify(websocket);
+                if (lastSentFrameWasIdentify.isMarked()) {
+                    logger.info("Hit identifying rate limit. Retrying in 5 seconds...");
+                    lastIdentificationPerAccount.put(api.getToken(), System.currentTimeMillis());
+                    sendIdentify(websocket);
+                } else {
+                    // Invalid session :(
+                    logger.info("Could not resume session. Reconnecting now...");
```
BtoBastian: Quote from the docs (https://discordapp.com/developers/docs/topics/gateway#resuming):

> If successful, the gateway will respond by replaying all missed events in order, finishing with a Resumed event to signal replay has finished, and all subsequent events are new. It's also possible that your client cannot reconnect in time to resume, in which case the client will receive a Opcode 9 Invalid Session **and is expected to wait a random amount of time—between 1 and 5 seconds—then send a fresh Opcode 2 Identify**.

You also must wait 1-5 seconds if resuming failed, regardless of the ratelimit.

Vampire: Ah, ok, wasn't aware of that. I just tried to resume with an invalid session and an immediate identify worked fine. I'll tweak the code a bit regarding this.

Vampire: added

BtoBastian:
Thank you, it should be fine now. :)

PR #126, opened 2018-01-07
Auto Sharding

Vampire:
With these commits the `DiscordApiBuilder` is now able to login multiple shards and even automatically request the amount of recommended shards from the Discord API.
With these changes you can do stuff like for example

```java
new DiscordApiBuilder()
        .setToken(botToken)
        .setTotalShards(6)
        .loginAllShards()
        // or
        //.loginShards(shard -> shard % 2 == 0)
        // or
        //.loginShards(0, 1)
        // or
        //.loginShards(IntStream.range(0, 4).toArray())
        .forEach(shard -> shard
                .thenAccept(this::handleReadyEvent)
                .exceptionally(Javacord::exceptionLogger)
        );
```

or

```java
new DiscordApiBuilder()
        .setToken(botToken)
        .setRecommendedTotalShards()
        .thenAccept(discordApiBuilder -> discordApiBuilder
                .loginAllShards()
                // or
                //.loginShards(shard -> shard % 2 == 0)
                // or
                //.loginShards(0, 1)
                // or
                //.loginShards(IntStream.range(0, 10).toArray())
                .forEach(shard -> shard
                        .thenAccept(this::handleReadyEvent)
                        .exceptionally(Javacord::exceptionLogger)
                ))
        .exceptionally(Javacord::exceptionLogger);
```

### DIFF
--- a/src/main/java/de/btobastian/javacord/DiscordApiBuilder.java
+++ b/src/main/java/de/btobastian/javacord/DiscordApiBuilder.java
@@ -1,8 +1,12 @@
 package de.btobastian.javacord;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.IntStream;
 
 import de.btobastian.javacord.utils.logging.LoggerUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


BtoBastian:
Is it really necessary for the `setRecommendedTotalShards()` to return a Future?
I feel like it would be easier to use, if this is done in the `loginAllShards()`

Vampire:
Hm, let me think about it

Vampire:
Well, if there would be a `getTotalShards()` method, it could be handy to have it set "immediately" and not delayed at login time.

If it is done at login time, then at least not only at `loginAllShards()`, but also at `loginShards(int...)` and `loginShards(IntPredicate)`, because you can e. g. start two JVMs, one to handle the even and one to handle the odd shards if it is too much for one process or whatever.

The main reason why I like it better as it is currently is probably the error handling.
As it is you can easily handle an exception that happens during the bot-gateway call.

The question is how it should be done if it is done in the login methods.
I could e. g. change `setRecommendedTotalShards()` to set totalShards to `-1` as marker for recommended account, then the lib-user can also call `setTotalShards()` afterwards without an HTTP call happened or happening which might be good and more flexible.
And in the login methods then only return one `CompletableFuture` in the result that failed exceptionally if the bot-gateway call failed. But then you loose the asynchronity a bit as you have to wait for the bot-gateway call to return to decide whether to return that one future or one future per shard.

Alternatively you could return one future per shard in any case but in case the bot-gateway call returned exceptionally complete all of them exceptionally.

What do you think?

Vampire:
Hm, maybe even the combination of the last two works

Vampire:
Hm, well, but then all the currentShard-totalShards-compatibility checks also have to be moved from the setters to the login call, as `totalShards` is not known upfront.

BtoBastian:
I think it would be fine, if all `CompletableFuture`s would complete exceptionally if the gateway call failed.
I also see no problem in the compatibility-checks causing the `CompletableFuture` to complete exceptionally instead of directly throwing the exception.

Vampire:
Hm, I have a hen-and-egg problem here.
For `login()` (one shard) it is easy to do the recommended shard count retrieving upfront by doing `getTotalShards().thenCompose(this::login)`. (`getTotalShards()` returns `CompletableFuture<Integer>`, `login(int)` returns `CompletableFuture<DiscordApi>`.
But for the multi-shard login methods I'm a bit stumped.
If I would know how many total shards there are it would be pretty easy, doing the same `thenCompose` for the single shards.
But without I don't see a way how to transform the `CompletableFuture<Integer>` into a collection of `CompletableFuture<DiscordApi>`.

Maybe I just have a knot in my head and you can push me out of that loop.
Otherwise what do you think about returning `CompletableFuture<Collection<CompletableFuture<DiscordApi>>>` instead from the multi-shard login methods. The client code would then look like e. g.

```java
new DiscordApiBuilder()
        .setToken(botToken)
        .useRecommendedTotalShards()
        .loginAllShards()
        .thenAccept(shards -> shards
                .forEach(shard -> shard
                        .thenAccept(this::handleReadyEvent)
                        .exceptionally(Javacord::exceptionLogger)
                ))
        .exceptionally(Javacord::exceptionLogger);
```

instead of previously

```java
new DiscordApiBuilder()
        .setToken(botToken)
        .setRecommendedTotalShards()
        .thenAccept(discordApiBuilder -> discordApiBuilder
                .loginAllShards()
                .forEach(shard -> shard
                        .thenAccept(this::handleReadyEvent)
                        .exceptionally(Javacord::exceptionLogger)
                ))
        .exceptionally(Javacord::exceptionLogger);
```

I think this variant would be doable.

BtoBastian:
You're right, that's something I didn't consider.
In that case i think your original solution would be the best.

Vampire:
ok, one moment, I'll anyway change some minor things I found while investigating

Vampire:
ok, updated

PR #130, opened 2018-01-08
Fix usage of it's where its would be correct

Vampire:


BtoBastian:
._.
I hate it. I know the difference but make this mistake nevertheless... ^^

Vampire:
Just always write "it is" instead of "it's", then it shouldn't happen. :-D
Is better language for official documents anyway - at least that is what university said for papers. :-)

PR #131, opened 2018-01-09
Improve the exception logger so that in more cases there is a path to the user-code in the stacktrace, even if it is only to the exceptionally call

Vampire:


src/main/java/de/btobastian/javacord/Javacord.java
```diff
@@ -42,22 +42,34 @@
...
+            }
+            Throwable rootCause = throwable;
+            while (rootCause.getCause() != null) {
+                rootCause = rootCause.getCause();
+            }
+            try {
+                rootCause.initCause(originCause);
+                logger.error("Caught unhandled exception!", throwable);
+            } catch (IllegalStateException ise) {
+                // cause was set to null explicitly, so we can only prepend it instead of appending
```
BtoBastian: I would always wrap it instead of having two different options.

src/main/java/de/btobastian/javacord/Javacord.java
```diff
@@ -42,22 +42,34 @@
...
-    public static <T> T exceptionLogger(Throwable throwable) {
-        Throwable cause = throwable.getCause();
-        while ((throwable instanceof CompletionException) && (cause != null)) {
-            throwable = cause;
-            cause = throwable.getCause();
-        }
-        logger.error("Caught unhandled exception!", throwable);
-        return null;
+    public static <T> Function<Throwable, T> exceptionLogger() {
+        Exception originCause = new Exception("origin of exceptionLogger call");
```
BtoBastian: Instead of using a "raw" Exception, maybe something like CompletionException would fit better?

src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
```diff
@@ -66,7 +66,7 @@ public RestRequest(DiscordApi api, RestMethod method, RestEndpoint endpoint) {
         this.method = method;
         this.endpoint = endpoint;
 
-        this.origin = new Exception();
+        this.origin = new Exception("origin of RestRequest call");
```
BtoBastian: With this change I would have removed the old way of creating an exception in the RestRequest class completely.
Otherwise we would end up with two different solutions for the some problem.

src/main/java/de/btobastian/javacord/Javacord.java
Vampire: I feared that you'd say that, but I thought I'd try to convince you anyway. :-D
This is where the code comes from, so I as user would expect it to be at the bottom of the stacktrace where always the bottom-most cause is found.
The wrapping is just a fallback if the root cause got set `null` as cause explicitly which didn't happen actually in any of my tests.

src/main/java/de/btobastian/javacord/Javacord.java
Vampire: _If_ we always wrap it, maybe, which I think would not be what a user excpects.

src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
Vampire: Well, not exactly. The "old" way shows a more correct location if the rest call was produced by user-code, that's why I kept it in. The "new" way can always only show the location where `exceptionLogger()` was called. And additionally the "old" way also works for custom exception handling that is not using the exception logger of Javacord.

src/main/java/de/btobastian/javacord/Javacord.java
BtoBastian: Well, but that's exactly how `.join()` on `CompletableFuture` does it as well.

src/main/java/de/btobastian/javacord/utils/rest/RestRequest.java
Vampire: Just hit a case where the exception logger code does not help at all, but the `RestRequest` code does:
```java
java.util.concurrent.CompletionException: Received a 400 response from Discord with body {"code": 10014, "message": "Unknown Emoji"}!
	at de.btobastian.javacord.Javacord.exceptionLogger(Javacord.java:53) ~[javacord-3.0.0.jar:?]
	at net.kautler.discord.bot.translator.TranslatorDiscordBot.handleMessageEditEvent(TranslatorDiscordBot.java:200) ~[main/:?]
	at de.btobastian.javacord.utils.handler.message.MessageUpdateHandler.lambda$dispatchEditEvent$7(MessageUpdateHandler.java:135) ~[javacord-3.0.0.jar:?]
	at de.btobastian.javacord.utils.DiscordWebSocketAdapter.lambda$null$10(DiscordWebSocketAdapter.java:638) ~[javacord-3.0.0.jar:?]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1374) ~[?:1.8.0_131]
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580) ~[?:1.8.0_131]
	at de.btobastian.javacord.utils.DiscordWebSocketAdapter.lambda$dispatchEvent$11(DiscordWebSocketAdapter.java:636) ~[javacord-3.0.0.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_131]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_131]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [?:1.8.0_131]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [?:1.8.0_131]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_131]
Caused by: de.btobastian.javacord.exceptions.DiscordException: Received a 400 response from Discord with body {"code": 10014, "message": "Unknown Emoji"}!
	at de.btobastian.javacord.utils.rest.RestRequest.executeBlocking(RestRequest.java:362) ~[javacord-3.0.0.jar:?]
	at de.btobastian.javacord.utils.ratelimits.RatelimitManager.lambda$null$6(RatelimitManager.java:137) ~[javacord-3.0.0.jar:?]
	... 5 more
Caused by: java.lang.Exception: origin of RestRequest call
	at de.btobastian.javacord.utils.rest.RestRequest.<init>(RestRequest.java:74) ~[javacord-3.0.0.jar:?]
	at de.btobastian.javacord.entities.message.Message.addReaction(Message.java:189) ~[javacord-3.0.0.jar:?]
	at de.btobastian.javacord.entities.message.Message.addReaction(Message.java:538) ~[javacord-3.0.0.jar:?]
	at jdk.nashorn.internal.scripts.Script$10$\^eval\_.:program(<eval>:1) ~[?:?]
	at jdk.nashorn.internal.runtime.ScriptFunctionData.invoke(ScriptFunctionData.java:637) ~[nashorn.jar:?]
	at jdk.nashorn.internal.runtime.ScriptFunction.invoke(ScriptFunction.java:494) ~[nashorn.jar:?]
	at jdk.nashorn.internal.runtime.ScriptRuntime.apply(ScriptRuntime.java:393) ~[nashorn.jar:?]
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:449) ~[nashorn.jar:?]
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:406) ~[nashorn.jar:?]
	at jdk.nashorn.api.scripting.NashornScriptEngine.evalImpl(NashornScriptEngine.java:402) ~[nashorn.jar:?]
	at jdk.nashorn.api.scripting.NashornScriptEngine.eval(NashornScriptEngine.java:155) ~[nashorn.jar:?]
	at javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:264) ~[?:1.8.0_131]
	at sun.reflect.GeneratedMethodAccessor2.invoke(Unknown Source) ~[?:?]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_131]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_131]
	at org.jboss.weld.bean.proxy.AbstractBeanInstance.invoke(AbstractBeanInstance.java:38) ~[weld-se-2.4.5.Final.jar:2.4.5.Final]
	at org.jboss.weld.bean.proxy.ProxyMethodHandler.invoke(ProxyMethodHandler.java:100) ~[weld-se-2.4.5.Final.jar:2.4.5.Final]
	at org.jboss.weldx.script.ScriptEngine$517641905$Proxy$_$$_WeldClientProxy.eval(Unknown Source) ~[weld-se-2.4.5.Final.jar:?]
	at net.kautler.discord.bot.translator.command.EvaluateCommand.execute(EvaluateCommand.java:65) ~[main/:?]
	at net.kautler.discord.bot.translator.command.EvaluateCommand$Proxy$_$$_WeldClientProxy.execute(Unknown Source) ~[main/:?]
	at net.kautler.discord.bot.translator.TranslatorDiscordBot.handleMessage(TranslatorDiscordBot.java:221) ~[main/:?]
	at java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:656) ~[?:1.8.0_131]
	at java.util.concurrent.CompletableFuture.uniAcceptStage(CompletableFuture.java:669) ~[?:1.8.0_131]
	at java.util.concurrent.CompletableFuture.thenAccept(CompletableFuture.java:1997) ~[?:1.8.0_131]
	... 11 more
```

Vampire:
Just added another commit that also provides an `UncaughtExceptionHandler` that behaves similar to the exception logger `Function<Throwable, T>`.

Vampire:
Now I also added the possibility to suppress certain `Throwable` sub-classes and / or only log throwables that pass a given predicate and moved the stuff into its own utility class, as it overtook the `Javacord` class too much in the meantime.

README.md
```diff
@@ -64,12 +64,14 @@ Logging in is very simple
 public class MyFirstBot {
 
     public static void main(String[] args) {
+        Thread.setDefaultUncaughtExceptionHandler(ExceptionLogger.getUncaughtExceptionHandler());
```
BtoBastian: Should this really be in the readme?
It's nice that the ExceptionLogger provides an UncaughtExceptionHandler, but the example in the readme should be kept as simple as possible (and only focus on the main Javacord concepts). That would probably better fit in an extra wiki article about exception handling.

Vampire: It is the first look at copy & paste example that people will start with, so I thought I put the uncaught exception handler in there too so that the average user has his exceptions logged instead of shown on stderr.
If you don't like it, I can easily remove it again of course.

BtoBastian: Yes, please. I really don't think we need this in the readme.

Vampire: changed

PR #133, opened 2018-01-09
Add TextChannel#canRemoveReactionsOfOthers(User)

Vampire:


BtoBastian:
I would rather call the method `canRemoveReactionsOfUser(User)` (which always returns `true`, if `user.isYourself()`).

Vampire:
This is "can the given user delete reactions of other users in this text channel", just like the other `canXXX` methods in `TextChannel` like e. g. `canAddNewReactions(User)`.
Or did you suggest adding a second parameter?

BtoBastian:
Ohh, my bad I misunderstood the method (which speaks for a better name).
Not sure if this method is even useful, as it's just an alias for another method.

Vampire:
Well yes, it is just an alias for `canManageMessages`.
My intention was that the user does not need to know which technical Discord permission you need to have to remove reactions of others.
Which better name would you suggest?
`canDeleteReactionsOfOtherUsers(User)`?
`canDeleteNonOwnReactions(User)`?
`canDeleteForeignReactions(User)`?
`canDeleteOthersReactions(User)`?
Something else? (except you dismiss the method as a whole of course)

BtoBastian:
`canRemoveReactionsOfOthers(User)` should be fine. Don't know why I misunderstood it in the first place...

Vampire:
Ok, fine. :-)

PR #135, opened 2018-01-09
Add Message#removeReactionByEmoji variants

Vampire:
- Add Message#removeReactionByEmoji(Emoji, User)
- Add Message#removeReactionByEmoji(String, User)
- Add Message#removeOwnReactionByEmoji(Emoji)
- Add Message#removeOwnReactionByEmoji(String)
- Add Message#removeReactionByEmoji(Emoji)
- Add Message#removeReactionByEmoji(String)

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/message/Message.java
+++ b/src/main/java/de/btobastian/javacord/entities/message/Message.java
@@ -580,6 +580,16 @@ public interface Message extends DiscordEntity, Comparable<Message> {
     }
 
     /**
+     * Removes all reactors of a given emoji reaction.
+     *
+     * @param emoji The emoji of the reaction.
+     * @return A future to tell us if the deletion was successful.
+     */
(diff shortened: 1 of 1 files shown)


BtoBastian:
What do you think about `removeOwnReactionByEmoji(...)` instead of `removeYourReactionByEmoji(...)`?

Vampire:
I just tried to stick to your naming conventions. If it is about the bot you always have "you" / "yourself". The only occurrence where you have "own" is in "canChangeOwnNickname" and there it does not refer to the bot but to the user that is given as argument. But if you prefer "own" over "your" here, just tell me and I'll change it, I have no strong feelings about this, especially as I would have used "me", "myself" and "my" probably. :-D

BtoBastian:
I think `own` fits fine into my naming conventions.

Vampire:
ok, then I'll happily change it later

Vampire:
done

PR #142, opened 2018-01-12
Fix ImplDiscordApi memory leak

Vampire:
If you have a shutdown hook that calls anything on an instance, it has a strong reference and thus the object and all dependencies cannot be garbage collected.
Because of that now `WeakReference`s are used, so that the instances can be garbage collected properly despite the shutdown hook.
The additional `finalize` that disconnects is necessary if the instance gets unreferenced and is garbage collected without `disconnect` being called before explicitly.

### DIFF
--- a/src/main/java/de/btobastian/javacord/ImplDiscordApi.java
+++ b/src/main/java/de/btobastian/javacord/ImplDiscordApi.java
@@ -65,6 +65,7 @@ import de.btobastian.javacord.utils.ratelimits.RatelimitManager;
 import okhttp3.OkHttpClient;
 import org.slf4j.Logger;
 
+import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -306,10 +307,21 @@ public class ImplDiscordApi implements DiscordApi {
                     , 30, 30, TimeUnit.SECONDS);
 
             // Add shutdown hook
-            ready.thenAccept(api -> Runtime.getRuntime().addShutdownHook(new Thread(api::disconnect)));
+            ready.thenAccept(api -> {
+                WeakReference<DiscordApi> discordApiReference = new WeakReference<>(api);
+                Runtime.getRuntime().addShutdownHook(new Thread(() -> Optional.ofNullable(discordApiReference.get()).ifPresent(DiscordApi::disconnect),
+                                                                String.format("%s Shutdown Disconnector (%s)", api.getClass().getSimpleName(), api)));
+            });
         } else {
-            Runtime.getRuntime().addShutdownHook(new Thread(this::disconnect));
+            WeakReference<DiscordApi> discordApiReference = new WeakReference<>(this);
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> Optional.ofNullable(discordApiReference.get()).ifPresent(DiscordApi::disconnect),
+                                                            String.format("%s Shutdown Disconnector (%s)", getClass().getSimpleName(), this)));
         }
+    }
+
+    @Override
+    protected void finalize() {
+        disconnect();
     }
 
     /**


BtoBastian:
I would just completely remove the shutdown hook then.
- The `WeakReference` solution won't work in most cases, as it's very unlikely for the GC to not run even once
- The `finalize` method is generally considered bad practice and got deprecated in Java 9 (see http://mail.openjdk.java.net/pipermail/core-libs-dev/2017-March/046650.html).

It should be the job of the user to properly close the connection anyway.

Vampire:
> The `WeakReference` solution won't work in most cases, as it's very unlikely for the GC to not run even once

I think you slightly misunderstand.

If you quit the JVM while the instance is strongly referenced, the finalizer will not run, but the shutdown hook with the `WeakReference` will work as intended. (Tried with `!eval System.exit(0)` and some `stdout` debugging)

If you quit the JVM while the instance is not strongly referenced and no garbage collection was done, obviously a garbage collection is done now, as the `finalize` kicked in.

If the instance is not strongly referenced and a garbage collection is done, the `finalize` kicks in.

So both mechanisms have their use-case and in a (imo) typical usage scenario (auto-sharding and then one instance per shard that is live until end of program), both will kick in at the appropriate places.

> The `finalize` method is generally considered bad practice and got deprecated in Java 9 (see http://mail.openjdk.java.net/pipermail/core-libs-dev/2017-March/046650.html).

Well, the `finalize` method is considered bad practice because of its quirks like
- not guaranteed to run
- can make the instance strongly referenced again
- because of that needs two GC cycles to be thrown away
- is ever only called once at most per instance, even if made strongly referenced again and then again eligible for collection
- needs extra time at object creation and destruction
- cannot be cancelled, e. g. after already doing the cleanup manually
- maybe others

But if used correctly, considering these quirks, it is ok to use it.

> - not guaranteed to run

It is just a safety precaution in case the user forgot to do it manually, or his code had some error that prevented it. If it is not done, the connection will time out, but if possible it would be nice to cleanly shut down, so not that tragic if it is not run.

> - can make the instance strongly referenced again

It doesn't do.

> - because of that needs two GC cycles to be thrown away

Well yes, that one cannot be prevented, but is not that problematic as there will not be that much instances of this class.

> - is ever only called once at most per instance, even if made strongly referenced again and then again eligible for collection

As it is not made strongly referenced again, not a problem.

> - cannot be cancelled, e. g. after already doing the cleanup manually

As I said, pretty few instances, so not a show-stopper.

> - needs extra time at object creation and destruction

As I said, pretty few instances, so not a show-stopper.

So yes, there are very few valid use-cases for `finalize()` methods and unexperienced developers might be tempted to use it in the wrong situations and because of that it probably is a good thing that it now got deprecated and a proper replacement provided with the `Cleaner` class that should care about all those points properly.

But such safety-clean-up-if-user-didn't-do are exact the use-cases `finalize` is intended to be used for. And maybe cleaning up native resources that the garbage collector is not able to handle of course. That's also the use-cases the JDK mostly uses `finalize`, like closing I/O-Streams if you didn't do it and so on.

> It should be the job of the user to properly close the connection anyway.

Yes, definitely, but it is always nice to have a backup solution if he didn't do it or his code is buggy or threw an exception and thus skipped it or whatever.

So actually I'd leave it like it is now as a safety net and if you upgrade the minimum requirement to Java 9, then use the `Cleaner` class instead. As `PhantomReferences` and `ReferenceQueues` are already present almost since the beginning, one could of course build an own clone of the `Cleaner` class to do the same the `Cleaner` class is doing, just in Java 8, but I don't think it is worth the effort with that less instances and only as a safety net.

Don't get me wrong, if you don't want to have it, well, let's throw it out.
I just wanted to share the thoughts I had with you, for you being able to make a qualified decision. :-)

BtoBastian:
Hmm, I remember playing with `WeakReference`s when designing the message cache and experienced some strange behavior, which I cannot reproduce now. I feel like I made a mistake back then which gave me a wrong understanding about `WeakReference`s. My bad.

However: I don't think there is a single use case in which the `finalize` method would be necessary: The instance will never be garbage collected anyway, if the user didn't call `disconnect()`, because there are still threads in the background to keep the connection alive (e.g. the heartbeat) which prevent it to be garbage collected.

Vampire:
WeakReference just means "do not prevent garbage collection", if it was garbage collected `get()` returns `null`, otherwise the instance. There are also `SoftReference`s and `PhantomReference`s.

Sure it is, given that disconnect is not called manually. E. g. auto-sharding creates a short-lived instance. Sure, currently it should cleanly disconnect, but maybe a bug hinders it somewhen in the future or similar. 

Or one implements a multi-bot where various instances as probably created and destroyed during the lifetime of the application.

Or some dynamic re-sharding, that automatically ends all current shards and starts the cluster freshly if enough servers are added or removed.

Vampire:
WeakReference just means "do not prevent garbage collection", if it was garbage collected `get()` returns `null`, otherwise the instance. There are also `SoftReference`s and `PhantomReference`s.

Sure it is, given that disconnect is not called manually. E. g. auto-sharding creates a short-lived instance. Sure, currently it should cleanly disconnect, but maybe a bug hinders it somewhen in the future or similar. 

Or one implements a multi-bot where various instances as probably created and destroyed during the lifetime of the application.

Or some dynamic re-sharding, that automatically ends all current shards and starts the cluster freshly if enough servers are added or removed.

Vampire:
WeakReference just means "do not prevent garbage collection", if it was garbage collected `get()` returns `null`, otherwise the instance. There are also `SoftReference`s and `PhantomReference`s.

Sure it is, given that disconnect is not called manually. E. g. auto-sharding creates a short-lived instance. Sure, currently it should cleanly disconnect, but maybe a bug hinders it somewhen in the future or similar. 

Or one implements a multi-bot where various instances as probably created and destroyed during the lifetime of the application.

Or some dynamic re-sharding, that automatically ends all current shards and starts the cluster freshly if enough servers are added or removed.

Vampire:
WeakReference just means "do not prevent garbage collection", if it was garbage collected `get()` returns `null`, otherwise the instance. There are also `SoftReference`s and `PhantomReference`s.

Sure it is, given that disconnect is not called manually. E. g. auto-sharding creates a short-lived instance. Sure, currently it should cleanly disconnect, but maybe a bug hinders it somewhen in the future or similar. 

Or one implements a multi-bot where various instances as probably created and destroyed during the lifetime of the application.

Or some dynamic re-sharding, that automatically ends all current shards and starts the cluster freshly if enough servers are added or removed.

PR #143, opened 2018-01-12
Add Message#removeReactionsByEmoji and Message#removeOwnReactionsByEmoji

Vampire:
- Message#removeReactionsByEmoji(User, Emoji...)
- Message#removeReactionsByEmoji(User, String...)
- Message#removeReactionsByEmoji(Emoji...)
- Message#removeReactionsByEmoji(String...)
- Message#removeOwnReactionsByEmoji(Emoji...)
- Message#removeOwnReactionsByEmoji(String...)

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/message/Message.java
+++ b/src/main/java/de/btobastian/javacord/entities/message/Message.java
@@ -31,6 +31,7 @@ import de.btobastian.javacord.utils.rest.RestMethod;
 import de.btobastian.javacord.utils.rest.RestRequest;
 
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


BtoBastian:
Hey,
can you please create the corresponding static methods (just take a look at `delete(...)`, etc.) and the methods in the `MessageEvent` class as well? Usually all non-static methods of in the Message class have a static counterpart and a method in `MessageEvent` as well, because you cannot expect all Messages to be cached.

Vampire:
I have a déjà-vu, we had this conversation before about #135 on the Discord server didn't we?
The methods all are just convenience methods that ultimately forward to the static `Reaction.removeUser`.
I thought this is sufficient.
At least it was for the singular forms of these methods in #135.

Vampire:
And about the `MessageEvent`, actually I thought about adding them there.
But then I have seen that only `removeAllReactionsFromMessage` is implemented in there and none of the other 6 ones.
But I now just realised that those 6 ones are also from me in #135 and there you obviously just forgot to whine about `MessageEvent`. :-D

Vampire:
Added them to `MessageEvent` now.

BtoBastian:
👍

PR #146, opened 2018-01-14
Use MDCCloseable with try-with-resources block instead of manual removal and prevent the MDC warning of slf4j if no proper logging binding is present

Vampire:


BtoBastian:
👍 
That's something that bothered me, too. I just wasn't sure what would be the right solution:
- Suppress the warning or
- Remove my custom warning and let just slf4j handle it

I'm fine with both solution, so thank you. :-)

Vampire:
Well, I think it is. Using the `MDC` only makes sense if you actually log it. If you use some binding that does not have an `MDC` implementation, it is ok to warn and try to convince the user to use a proper logging binding. 

But when the fallback logger is used, the pattern is not configurable, so the MDC is never logged anyway and thus it is ok to just ignore the internal `MDC` calls. User-code should not use the new method, but `MDC` itself and then will again get the warning, because if he uses `MDC`, it is likely he wants to use its.

PR #150, opened 2018-01-14
Added Message#getCustomEmojis

BoneyIsSpooky:


BtoBastian:
Hey, thank you for the PR.
However, there are some things which needs to be changed before I can merge it:
* Why `getLocalCustomEmojis()` instead of `getCustomEmojis()`?
* The method should not only return the emojis of the current server (is that what you meant with `local`?) but every custom emoji in the message. If the emoji is unknown, you'll have to create a new `CustomEmoji` for it (You can extract the `id`, `name` and `animated` properties from the string). You'll need to create a new constructor in `ImplCustomEmoji` for this, as it currently requires a `JsonNode`.
* You should use the new class https://github.com/BtoBastian/Javacord/blob/v_3/src/main/java/de/btobastian/javacord/utils/DiscordRegexPattern.java for the regex pattern. To be fair: This class was created after your PR 😉 
* Just a small thing: Take a look at the JavaDoc design of the other methods (or here: https://github.com/BtoBastian/Javacord3-Docs/blob/master/.github/CONTRIBUTING.md#code-conventions), as there's a missing linebreak. 

Please tell me if you plan to fix this yourself, or want me to fix it.

BoneyIsSpooky:
Thank you for the feedback, I will fix it.

BoneyIsSpooky:
I have applied the requested changes.

BtoBastian:
Would you mind cleaning up the the commits a little bit like suggested by Vampire on the Discord server?

BoneyIsSpooky:
Done

PR #156, opened 2018-01-15
Do not confuse the heartbeat timers of different web sockets

Vampire:
If one websocket gets discarded and a new one created, it happened to me that two websockets at the same time did try to access the heartbeat timer from different threads as each `WebSocket` instance has its own `ReadingThread` instance. The one then canceled the timer and `null`ed it, then the other got a `NullPointerException`.
Now the heartbeat timer is assigned to the web socket it heartbeats to to prevent this.

### DIFF
--- a/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
+++ b/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
@@ -67,6 +67,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -103,7 +104,7 @@ public class DiscordWebSocketAdapter extends WebSocketAdapter {
 
     private WebSocket websocket = null;
 
-    private Timer heartbeatTimer = null;
+    private final Map<WebSocket, Timer> heartbeatTimerByWebSocket = Collections.synchronizedMap(new WeakHashMap<>());
 
     private int lastSeq = -1;
     private String sessionId = null;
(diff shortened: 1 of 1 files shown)


BtoBastian:
I really don't get how the current implementation could result in a NPE.
There's only one place where `heartbeatTimer` is accessed and this is the `onDisconnect()` method and this method should never by called simultaneously by different threads.

Vampire:
Not easy to grasp like most multi-threading issues. :-)
There were multiple websocket instances.
Each websocket instance has a separate `ReadingThread`.
The `ReadingThread` calls `onDisconnected()` and thus the variable is set to `null`.
Another thread might then get NPE.
The heartbeat timer is intrinsically coupled with the websocket as it sends the heartbeat to it, so it should also tied to the websocket.

Other stuff like `lastSentFrameWasIdentify` and so on could of course also suffer from this behavior, but it is quite unlikely that there are multiple websockets that access these variables at the same time. For the heartbeat related fields (this and the added one in the other PR) though are more likely to hit the problem when a reconnect happens and actually I did hit the problem, hence I fixed it. :-)

BtoBastian:
Sure, if there are multiple threads calling `onDisconnect()` this might be a problem, but when does this ever happen? There should only be a single ReadingThread at once per `DiscordWebSocketAdapter` class.
The websockets reconnects **after** setting `heartbeatTimer` to `null`, so reconnects should not affect this at all. 

Maybe it was caused by one of your other changes to the WebSocketAdapter, because I really don't see how this can happen with the current code.

Vampire:
Hm, maybe I had the problem due to the reconnect opening multiple websockets, let me think about it.

Vampire:
I don't remember in which situation I got the error, but yes, with the current implementation and my fix, for the commanded reconnect, the variable should be thread confined to the the reading thread of the current websocket, so synchronization should not be necessary as long as it is only accessed from that one thread. Let's close this PR until it happens again, then we can revive it. I guess I got this due to the reconnect bug or something like that.

I'll also change the other PR with the ACKs to not have one per websocket, as it is unneccessary due to the same reason. It needs synchronization as different threads use it, but not a distinction per web socket.

PR #157, opened 2018-01-15
Close connection if a heartbeat package was not answered properly until the next hearbeat is due

Vampire:
Implements the API doc requirement `If a client does not receive a heartbeat ack between its attempts at sending heartbeats, it should immediately terminate the connection with a non-1000 close code, reconnect, and attempt to resume.`.

### DIFF
--- a/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
+++ b/src/main/java/de/btobastian/javacord/utils/DiscordWebSocketAdapter.java
@@ -74,6 +74,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


BtoBastian:
Maybe there should also be a log messages when we disconnect the websocket?
Something like `Didn't receive ack for last heartbeat. Reconnecting...`?

Vampire:
Just pushed a new version without conflicts and with per-websocket non-sense removed.

Regarding your comment, `onDisconnected` already logs the close reason, did you not think of that or do you not consider it enough?

```
2018-01-17 11:40:52,452 <INFO > <ReadingThread            > <[]> <{shard=0}> <.btobastian.javacord.utils.DiscordWebSocketAdapter> Websocket closed with reason Heartbeat was not answered properly and code 1001 by client!
```

BtoBastian:
My bad, I didn't think about that

PR #158, opened 2018-01-16
Custom Emoji Update And Delete Events

Vampire:
- Send `CustomEmojiUpdateEvent`s
- Send `CustomEmojiDeleteEvent`s
- fix JavaDoc of `CustomEmojiCreateListener`
- Show the `CustomEmoji.animated` property in the `toString` representation

### DIFF
--- a/src/main/java/de/btobastian/javacord/DiscordApi.java
+++ b/src/main/java/de/btobastian/javacord/DiscordApi.java
@@ -55,6 +55,7 @@ import de.btobastian.javacord.listeners.server.channel.ServerChannelDeleteListen
 import de.btobastian.javacord.listeners.server.channel.ServerTextChannelChangeTopicListener;
 import de.btobastian.javacord.listeners.server.emoji.CustomEmojiChangeNameListener;
 import de.btobastian.javacord.listeners.server.emoji.CustomEmojiCreateListener;
+import de.btobastian.javacord.listeners.server.emoji.CustomEmojiDeleteListener;
 import de.btobastian.javacord.listeners.server.member.ServerMemberBanListener;
 import de.btobastian.javacord.listeners.server.member.ServerMemberJoinListener;
 import de.btobastian.javacord.listeners.server.member.ServerMemberLeaveListener;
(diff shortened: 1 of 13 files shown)


BtoBastian:
Hey, thank you. :)
However there are some things which needs to be changed:
- There should also be methods to add listeners to CustomEmoji objects (e.g. `emoji.addCustomEmojiDeleteListener(event -> /* do something */);`)
- CustomEmojiUpdateEvent does not fit into Javacords listener design (there's also no `ServerUpdateEvent` for example): There should be an extra event for every changed property, e.g. `CustomEmojiChangeNameEvent` with a `getOldName()` and a `getNewName()` method. Afaik the only thing you can update is the name, so this should not be a big deal to change.

Vampire:
Yeah, right about the object listeners of course, I'm gonna add them.
Hm, there is `MessageEditEvent`, I wonder why I didn't do `CustomEmojiEditEvent`.
Why is there no `ServerUpdateEvent` and similar? Maybe it would sense to have it?

BtoBastian:
Well `MessageEditEvent` is also only for the content (you could call it `MessageChangeContentEvent` to be consisten, but I feel like "edit" is fine in case of messages).
What would be the use case of a `ServerUpdateEvent`? Usually you want to listen for a specific property which changed.

Vampire:
Ah, ok, hm.

Dunno, maybe if you have a front-end for your bot that displays the servers, you might be interested in any change, without the need to add new listeners when they arise and easily be able to forget that.

Vampire:
Ok, update pushed, I hope I didn't forget anything this time. :-)

BtoBastian:
What's the reason for the `DiscordApi#addCustomEmojiDeleteListenerToCustomEmoji(...)` and `DiscordApi#addCustomEmojiDeleteListenerToServer(...)` methods?
Such methods usually don't exist in Javacord, with one exception: Messages, as they are not guaranteed to be cached.
Known servers however are always cached, so there's no need to have extra methods for it.

Vampire:
I'm not sure right now add I'm on mobile, I guess I thought you're all up to convenience methods and there is no reason to get the custom emoji or server first if you already have its ID and it would also work if not all servers are available yet. I'll have a look later and then probably remove them again.

Vampire:
Ok, update pushed, I think now it fit into Javacord. :-)

BtoBastian:
A `CustomEmojiChangeNameEvent#getEmoji()` method might be useful as well. Listening for name changes is pretty useless, if you can't get the emojis which name was changed.

Vampire:
Oh, damn, of course, this PR is not my best. :-D
There is a refactored out `CustomEmojiEvent` now that is the parent for all three like for other events.

BtoBastian:
Looks fine now 👍

PR #160, opened 2018-01-17
Added Message#getUnicodeEmojis, and a unicode emoji pattern

BoneyIsSpooky:


src/main/java/de/btobastian/javacord/utils/DiscordRegexPattern.java
```diff
@@ -22,6 +22,11 @@
      */
     public static final Pattern CUSTOM_EMOJI = Pattern.compile("<a?:(?<name>[0-9a-zA-Z_]+):(?<id>[0-9]+)>");
 
+    /**
+     * A pattern which checks for unicode emojis (e.g. {@code 🤔}
+     */
+    public static final Pattern UNICODE_EMOJI = Pattern.compile("\\p{So}|[\\uD83C-\\uDBFF\\uDC00-\\uDFFF]+");
```
Vampire: Are you sure your regex is fine?
I see two problems with it.
1. "Symbol Other" character group contains many characters, even if some of them are emojis, for sure not all, see http://www.fileformat.info/info/unicode/category/So/list.htm
2. Even if the ranges in the second character class are correct - and I don't say they are, I didn't check - it would not be correct, as two or more consecutive emojis would be matched as one

BoneyIsSpooky: Your second statement is incorrect, I have tested this and each emote is separate: https://i.imgur.com/0encWs1.png

BtoBastian: I can confirm number 1:

BoneyIsSpooky:
Closing because I haven't been able to fix the regex.

PR #162, opened 2018-01-19
Improve message caching

Vampire:
- remove deleted messages from the cache immediately
- use a map with weak keys for the messages in DiscordApi
- only process needed messages for cleanup of the messages cache in DiscordApi using a ReferenceQueue instead of looping through all messages
- remove Message#valid, Message#keepCached and Message#deleted
- do not keep messages in the cache for at least two minutes by tight coupling but just keep it in the cache until the garbage collector gets rid of it
  - it is very unlikely that the GC runs between the creation of the result message in sendMessage and the arrival of the MESSAGE_CREATE event, so the same message object should be returned there usually
  - if a message gets deleted it is a matter of luck whether the message is still in the cache when the delete event will be fired, but this is the same as currently and for bulk deletes it was quite unlikely before already as it just gets a list of IDs of messages that don't have to be in the cache at all anyway; if you would like to make the users able to get the deleted message in the delete event, then you would need to preserve a hard reference for some time for a deleted message and additionally before deleting a message or bulk-deleting messages you would need to first request those messages from the API to get them into the cache; even then it would be uncertain, because if some other user deletes a message, you have no chance to get hold of it if it wasn't in the cache already

### DIFF
--- a/src/main/java/de/btobastian/javacord/ImplDiscordApi.java
+++ b/src/main/java/de/btobastian/javacord/ImplDiscordApi.java
@@ -75,6 +75,8 @@ import de.btobastian.javacord.utils.rest.RestRequest;
 import okhttp3.OkHttpClient;
 import org.slf4j.Logger;
 
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -82,7 +84,9 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
@@ -227,7 +231,19 @@ public class ImplDiscordApi implements DiscordApi {
     /**
      * A map with all cached messages.
      */
-    private final ConcurrentHashMap<Long, Message> messages = new ConcurrentHashMap<>();
+    private final Map<Long, WeakReference<Message>> messages = Collections.synchronizedMap(new ConcurrentHashMap<>());
+
+    /**
+     * A map to retrieve message IDs by the weak ref that point to the respective
+     * message or used to point to it for usage in the messages cleanup,
+     * as at cleanup time the weak ref is already emptied.
+     */
+    private final Map<Reference<? extends Message>, Long> messageIdByRef = Collections.synchronizedMap(new WeakHashMap<>());
+
+    /**
+     * The queue that is notified if a message became weakly-reachable.
+     */
+    private final ReferenceQueue<Message> messagesCleanupQueue = new ReferenceQueue<>();
 
     /**
      * A map which contains all listeners.
@@ -317,9 +333,17 @@ public class ImplDiscordApi implements DiscordApi {
                 }
             });
 
-            getThreadPool().getScheduler().scheduleAtFixedRate(
-                    () -> messages.entrySet().removeIf(entry -> !((ImplMessage) entry.getValue()).keepCached())
-                    , 30, 30, TimeUnit.SECONDS);
+            // After minimum JDK 9 is required this can be switched to use a Cleaner
+            getThreadPool().getScheduler().scheduleWithFixedDelay(() -> {
+                for (Reference<? extends Message> messageRef = messagesCleanupQueue.poll();
+                     messageRef != null;
+                     messageRef = messagesCleanupQueue.poll()) {
+                    Long messageId = messageIdByRef.remove(messageRef);
+                    if (messageId != null) {
+                        messages.remove(messageId, messageRef);
+                    }
+                }
+            }, 30, 30, TimeUnit.SECONDS);
 
             // Add shutdown hook
             ready.thenAccept(api -> {
@@ -350,6 +374,7 @@ public class ImplDiscordApi implements DiscordApi {
         unavailableServers.clear();
         customEmojis.clear();
         messages.clear();
+        messageIdByRef.clear();
         timeOffset = null;
     }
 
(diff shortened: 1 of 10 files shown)


src/main/java/de/btobastian/javacord/entities/message/Message.java
```diff
@@ -57,7 +57,7 @@
                 .setUrlParameters(String.valueOf(channelId), String.valueOf(messageId))
                 .setRatelimitRetries(250)
                 .execute(result -> {
-                    api.getCachedMessageById(messageId).ifPresent(msg -> ((ImplMessage) msg).setDeleted(true));
+                    api.getCachedMessageById(messageId).ifPresent(msg -> ((ImplMessageCache) msg.getChannel().getMessageCache()).removeMessage(msg));
```
BtoBastian: I know it was like this in the original code as well, but maybe that's not a good idea as it's inconsistent with the rest of the library where stuff like this (keeping the cache up to date) is exclusively handled using the websocket events (just take a look at `ServerChannel#delete()` for example).
In the event however the message should be completely removed from the cache, including the map inside `ImplDiscordApi` even if there's still a reference.

src/main/java/de/btobastian/javacord/ImplDiscordApi.java
```diff
@@ -461,8 +483,12 @@ public CustomEmoji getOrCreateCustomEmoji(Server server, JsonNode data) {
...
         long id = Long.parseLong(data.get("id").asText());
         // The constructor already adds the message to the cache.
-        // If we use #computeIfAbsent() here, it would cause a deadlock
-        return messages.getOrDefault(id, new ImplMessage(this, channel, data));
+        // If we use #compute() here, it would cause a deadlock
+        WeakReference<Message> resultRef = messages.get(id);
+        // If the map has no entry or the weak ref was still in the map but cleared, create a new message
+        return (resultRef == null) || (resultRef.get() == null)
+               ? new ImplMessage(this, channel, data)
+               : resultRef.get();
```
BtoBastian: Shouldn't this be put into a `synchronized(messages)` block?
Otherwise it would be able to get two different `Message` objects for the same message from this message if it's called simultaneously. Or am I missing something?

src/main/java/de/btobastian/javacord/ImplDiscordApi.java
```diff
@@ -213,7 +218,19 @@
...
      */
-    private final ConcurrentHashMap<Long, Message> messages = new ConcurrentHashMap<>();
+    private final ConcurrentMap<Long, WeakReference<Message>> messages = new ConcurrentHashMap<>();
+
+    /**
+     * A map to retrieve message IDs by the weak ref that point to the respective
+     * message or used to point to it for usage in the messages cleanup,
+     * as at cleanup time the weak ref is already emptied.
+     */
+    private final Map<Reference<? extends Message>, Long> messageIdByRef = Collections.synchronizedMap(new WeakHashMap<>());
```
BtoBastian: I don't quite get why there's a need for this map?
To remove the entries from `messages`, wouldn't be
```java
messages.entrySet().removeIf(entry -> entry.getValue().get() == null);
```
enough?

src/main/java/de/btobastian/javacord/entities/message/Message.java
Vampire: Agreed about the cache update in the delete method.
Actually I also wondered about it but thought you might have had a reason, so kept it there.

About deleting in the event handlers, ok, done, this also makes it consistent that you cannot get the deleted message in a deletion listener as it then is guaranteed to be deleted. Or at least you cannot get it from the cache, you could only have it stored somewhere yourself.

This also makes the `RequestableMessageEvent` interface you talked about in Discord unnecessary, as you then can neither request the message event from discord, nor from the cache anymore. It will simply be a `MessageEvent` now.

src/main/java/de/btobastian/javacord/ImplDiscordApi.java
Vampire: Ah, you are partly right I think. We also need a `Collections.synchronizedMap(new ConcurrentHashMap<>())` then though. Because `ConcurrentHashMap` does not specify that it locks on itself, so even if it is the case it would be an implementation detail your rely on that could change any time or in any different JVM. `Collections.synchronizedMap` specifies that you can lock on the returned instance for iteration and `ConcurrentHashMap` is still necessary for atomicity guarantees of the `compute` method which are not present on `Map` itself.

src/main/java/de/btobastian/javacord/ImplDiscordApi.java
Vampire: Yes, it would be. But then you have to synchronize over `messages` again to safely iterate **and**  more importantly, you then have to iterate over all cached messages again for cleanup which is not necessary. With the current implementation you only directly remove the entries that became weakly reachable and thus were added to the `ReferenceQueue`. This time the cleanup is done in a matter of milliseconds, even if you are in maaaaany servers with maaaany messages in the cache.

An alternate implementation could stick closer to what `Cleaner` does and create a subclass of `WeakReference` that additionally holds the id, then you would not need this map as you can get the id from the `ReferenceQueue` entry. If you prefer this implementation tell me and I'll change it.

Vampire:
Update pushed

src/main/java/de/btobastian/javacord/entities/message/Message.java
BtoBastian: I think you misunderstood me.
I still would like the `MessageDeleteEvent#getMessage()` method which gets the cached message if there was one. 
- Get the message from cache if it is cached
- Remove the message from the cache
- Dispatch the `MessageDeleteEvent` and let the `getMessage()` method return the message we received from the cache before removing it

Vampire: Oh, ok, but that would only work in a limited range of cases actually. If you want it that way, I'll do it of course, I'm just not sure it really makes sense, as then it is a matter of luck of whether you get hold of the deleted message or not. Wouldn't it maybe be better / more consistent to just never be able to get the deleted message?

BtoBastian: Well, that's the reason why `Message#setCachedForever(boolean)` exists to eliminate the "luck factor".

Vampire: Well, if you set that property, you can always request the message from the API yourself too. :-)

Vampire: Ok, pushed a new version, now there **is** a `RequestableMessageEvent` and `MessageDeleteEvent` is an `OptionalMessageEvent`. Everything else that was an `OptionalMessageEvent` is now a `RequestableMessageEvent`.

PR #180, opened 2018-01-27
Cleanupable To Prevent Memory Leaks

Vampire:
Introduced Cleanupable to prevent memory leaks with repetetive scheduled tasks in the central scheduler that should not outlive a reconnect.
As a side effect also improved the private and group channel handling.


BtoBastian:
good stuff

PR #183, opened 2018-01-28
Consider backslash-escapes when rendering readable content of a message to do it like the official clients do

Vampire:


BtoBastian:
I really don't get these regex patterns anymore, but I trust you. :)

Vampire:
Maybe I make another PR that changes them to commented versions to have them better understandable :-) Basically the mention regexes not make sure there is not an uneven amount of backslashes in front which means escaping for discord and the `ESCAPED_CHARACTER` regex is to remove the escaping character in the cases where discord client also removes them and keep them where it keeps it. I made a loop to test all unicode characters in the BMP for finding out which they escape. :-)

PR #187, opened 2018-01-29
Add Updatable and UpdatableFromCache

Vampire:


BtoBastian:
I don't think this method should be implemented by all discord entities, but only the ones where it makes sense (e.g. not webhooks, or message author).
Also we should try to find a better method name.

Vampire:
Yeah, thought about that, my intention was to "remind" for new entities to not forget to add it. :-)
But we can maybe also wait for some user to complain that it is missing for an entity.
So I guess a separate interface like `Mentionable`, `Categorizable`, etc.?
Actually I also was not all too happy with the name and hoped you complain and suggest a better name. :-D

Why don't you think it makes sense for `Webhook`? Imho it makes perfect sense for `Webhook`, except the method should exclusively deliver cached entities, then the `Message` implementation also needs to be changed, but I don't see why `Message`s and `Webhook`s should not also be updatable.

So some questions:
- How do you like the interface being called?
- How do you like the method being called?
- For `Message` and `Webhook` (if online-refresh stays) should there be a separate method that returns a `CompletableFuture`, or should it refresh the entity synchronously like it is implemented currently?
- If second method, then in a separate interface? Names?

BtoBastian:
Yes, it should definitely have it's own interface.
It makes sense in the webhooks class, but not with an optional return type. It's blocking, so it should return a CompletableFuture.

I would split it into two different methods:
- `Xyz#getCurrentCachedInstance()` (With a better name, but it should be clear that it doesn't request it from Discord)
- `Xyz#requestLatestVersion()`, also with a better name. This method should check the cache first and if it isn't cached it should request it from Discord. 

Channel would for example only implement the first method, Message would implement both methods and Webhook would only implement the second method.

Vampire:
What do you think about this version? :-)

BtoBastian:
Why the generic?

Vampire:
To have the `Updatable` and `UpdatableFromCache` method return the entity of concrete type

Vampire:
If you do `((Channel) channel).getCurrentCachedInstance()` you get an `Optional<Channel>`, if you do `((ServerTextChannel) channel).getCurrentCachedInstance()` you get an `Optional<ServerTextChannel>`.

BtoBastian:
You can do this without generics as well.

Vampire:
How?

BtoBastian:
Ok, I just noticed that generics are weird again...

Vampire:
Just a little bit. :-D
Damn backwards compatibility. *g*

Vampire:
Next try :-)

BtoBastian:
Why does `UpdatableFromCache` extend `Updatable`?
Stuff like `Channel` should not extend `Updatable`, as it doesn't make sense for them.
The only entity I can think of were it would make sense to implement both, would be `Message`.

Vampire:
Imho it does make sense.

`Updatable` provides `getLatestInstance()`.
If an entity is cached, getting the latest instance (e. g. after a reconnect) is the same as getting the `getCurrentCachedInstance`, which is why `UpdatableFromCache#getLatestInstance` provides a default implementation that forwards to `getCurrentCachedInstance()` if not overwritten e. g. by `Message`.

Even for `Message` if the instance is already in the cache, it is delivered from the cache and not rerequested from the API.

So objects that are always cached like channels, are just a special case of that where the latest instance always comes from the cache which is exactly the default implementation of `UpdatableFromCache#getLatestInstance`.

If you want the latest instance, no matter whether a request is necessary or it comes from the cache always, you can just use `Updatable#getLatestInstance()`, otherwise you have to check whether `Updatable` is implemented and then use `getLatestInstance()` and if not check whether `UpdatableFromCache` is implemented and then use `getCurrentCachedInstance()` which is simply not necessary with the current code.

Currently if you always want the latest variant, no matter whether a request is necessary, you can simply use `Updatable#getLatestInstance()` and if you only want the latest cached version immediately without any request being done, you can use `UpdatableFromCache#getCurrentCachedInstance()`.

Btw. as we talk about it, an overwrite for `User` is missing as you can request non-cached users now.

Vampire:
Forget the last line, I actually already did that after the non-cacheable user PR was merged :-)

PR #189, opened 2018-01-30
Fix message history retrieval

Vampire:
- history around with a multiple of 100 and at least 200 tried to do the first request with 0 which is invalid
- history around abort condition after first request was not working correctly for 100
- history around calculation of subsequent batch sizes were wrong and unnecessarily complex and unnecessarily a floating point calculation
- history around had unnecessary null and 0 checks that can never fail
- history around did one request more than necessary and was unnecessarily complex
- history (no before, after or around) did always fetch the first 100 messages in all subsequent batches
- history did unnecessary floating point calculation


Vampire:
Part or #191

PR #190, opened 2018-01-31
Improve Type Continuously

Vampire:
- Add TextChannel#typeContinuouslyAfter for delayed starting continuously typing
- Use an AutoCloseable sub-interface that does not throw any checked exceptions from its close method


BtoBastian:
Would you mind giving `NonThrowingAutoCloseable` a short javadoc comment, before I merge the PR? :)

Vampire:
whoops, my bad, fixed

PR #195, opened 2018-02-06
Listener Management Improvements

Vampire:
- Remove unnecessary class parameters for listener registration methods
- Do not register listeners multiple times to the same context and always use the same listener manager, which now also gets its remove handlers invoked whenever the listener is removed from its context, not only when removed through the listener manager
- Add DiscordApi#removeObjectListener and DiscordApi#removeListener

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/message/Message.java
+++ b/src/main/java/de/btobastian/javacord/entities/message/Message.java
@@ -20,6 +20,8 @@ import de.btobastian.javacord.entities.message.emoji.impl.ImplCustomEmoji;
 import de.btobastian.javacord.entities.message.emoji.impl.ImplUnicodeEmoji;
 import de.btobastian.javacord.entities.permissions.PermissionType;
 import de.btobastian.javacord.entities.permissions.Role;
+import de.btobastian.javacord.listeners.ObjectAttachableListener;
+import de.btobastian.javacord.listeners.message.MessageAttachableListener;
 import de.btobastian.javacord.listeners.message.MessageDeleteListener;
 import de.btobastian.javacord.listeners.message.MessageEditListener;
 import de.btobastian.javacord.listeners.message.reaction.ReactionAddListener;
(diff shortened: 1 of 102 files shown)


Vampire:
Ok, totally changed what I started to do, but now it is ready for review. I hope I did all correct, especially for the docs, as there were some renamings like `...AttachableListener` and so on were involved. Feel free to complain if something is inconsistent. :-)

src/main/java/de/btobastian/javacord/entities/message/Message.java
```diff
@@ -353,6 +360,612 @@
...
+     * @param messageId The id of the message which should be listened to.
+     * @param listener The listener to add.
+     * @return The manager of the listener.
+     */
+    static ListenerManager<MessageDeleteListener> addMessageDeleteListener(
+            DiscordApi api, String messageId, MessageDeleteListener listener) {
+        try {
+            return addMessageDeleteListener(api, Long.valueOf(messageId), listener);
+        } catch (NumberFormatException ignored) {
+            return null;
```
BtoBastian: I don't think returning null is the right approach here.
I remember running into the same issue when creating the `ListenerManager` class.
Iirc I just removed the methods with a `String` parameter in the places where it would cause these kind of issues.
Another approach would be to just throw an `IllegalArgumentException`.

Vampire: However you like it better.
I just tried to follow how it was done for others.
If you want an IAE, we can just remove the try/catch, as NFE is an IAE.
The question is, whether this is too inconsistent then, as some methods return a default value (empty list) and some throw an exception when the string is not a proper number.

BtoBastian: I think it would be best to just removed these methods.

Vampire: ok, will do

Vampire:
New push available:
- change `newName` and `newColor` to be fields of the respective role events
- fixed the formatting of the JavaDoc of `ServerMemberJoinListner`
- added the multi-listener magic for the other channels that don't have attachable events right now for easy addition in the future
- removed the null-returning `Message` methods

PR #196, opened 2018-02-08
[WIP] Add User(Join/Leave)VoiceChannelEvent & respective listeners

shawishi:
Also add ability to get List of User connected to VoiceChannel (User#getConnectedVoiceChannel()) & VoiceChannel User is connected to (VoiceChannel#getConnectedUsers()).

### DIFF
--- a/src/main/java/de/btobastian/javacord/DiscordApi.java
+++ b/src/main/java/de/btobastian/javacord/DiscordApi.java
@@ -1,7 +1,25 @@
 package de.btobastian.javacord;
 
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import javax.imageio.ImageIO;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+
 import de.btobastian.javacord.entities.Activity;
 import de.btobastian.javacord.entities.ActivityType;
 import de.btobastian.javacord.entities.ApplicationInfo;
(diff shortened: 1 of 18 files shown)


shawishi:
This isn't ready to be merged yet (some formatting issues) - but gives you an idea so you can comment on what you think should be changed in the PR to merge..

Vampire:
From what I have seen while fixing the formatting for you there are some flaws in your changes.
Here what I have seen from a quick look, this is not a full review:
- Users can be connected to multiple voice channels, your changes imply only one connection (normal users can be connected to one voice channel per client, but I can e. g. be in one voice channel with web ui and in another voice channel with Android app (not in the same channel though) and besides that, bot users can connect to one voice channel per guild anyway even with the same "device") This manifests e. g.  in the connected voice channel property of the user (only one channel) and in th user join voice channel event (there should be no "previous channel")
- JavaDocs in the Events are not correct
- As there is only one channel relevant in the events, they should only be called `channel`
- The listeners should not only be attachable globally, but also to a user, to a voice channel and to a server
- Things like `getConnectedVoiceChannel` should maybe called `getJoinedVoiceChannels` if the listener and event are called `...Join...`, or the event and listeners should be `...Connect...` and `...Disconnect...` instead. Actually I think `...Join...` and `...Leave...` are fine, fits better with `ServerMember...` stuff. Probably I'd rename the methods to just `User#getVoiceChannels` and  `VoiceChannel#getUsers`.
- To stay consistent with `ServerMember...` stuff, things should probably be renamed to `VoiceChannelUser...`
- For the VoiceChannel, the `getConnectedUsers` should probably be called somehow more voice specfiic. `PrivateChannel#getConnectedUsers` is not very meaningful on its own. Maybe it should simply be `getVoiceUsers`
- What's the sense of the sneaky new `ImplUser` constructor? I don't think that makes much sense and it is nowhere used and also contradicts with the sense of the cache, I think you need to remove that
- You replace a `getIdAsString()` with `getId()`, I guess this is neither intended nor correct

Vampire:
- `Server#getVoiceChannel(User)` and `User#getVoiceChannel(Server)` would also be handy as you can have at most one joined voice channel per account, no matter whether user or bot account and using one or multiple devices.

Vampire:
- `addConnectedUser` and `removeConnectedUser` should not be part of the `VoiceChannel` interface as it is nothing a user of the API should call.

Vampire:
- The helper methods in the handler should be private and have JavaDoc
- You should not use `DiscordApi#getVoiceChannelById` if you have a server id, because that has to search all private, group and server voice channels instead of only the voice channels of the respective server

shawishi:
Now part of #207

PR #203, opened 2018-02-16
Replace web socket onDisconnect listener with LostConnectionListener

Vampire:


BtoBastian:
The listener thread is also part of the threadpool. Are you sure this won't cause any issues?
I can imagine this causing issues if there are other `LostConnectionListener`s.

Vampire:
Well, if it makes problems, I think it made problems before already, or even worse problems.
But actually I don't think so, as long as you can trust on "listeners get called in order of addition".
Actually as now our own listener mechanism is used, this can even be guaranteed, opposed to before where the web socket library was in charge.

Before it was like this:
Add a web socket listener for `onDisconnect` that gets called after the `onDisconnect` of the `DiscordWebSocketAdapter`, so the `LostConnectionListener` invocations were already scheduled in the thread pool. Then the thread pool gets shutdown which is no problem, because it is a graceful `shutdown()`, not a `shutdownNow()`, which means all schedules actions are executed before actual shutdown, just no new ones are accepted.

Now it is basically the same, procedure with the difference that:
- it is shorter
- it is nicer
- we guarantee order of listener invocation in addition order
- the shutdown is actually done after all other previously registered lost connection listeners which means that the other lost connection listeners can also still schedule tasks to be done in the thread pool which they weren't able to before, or rather maybe were able to, heavily depending on timing. Now they always can.

Especially due to the last point I think that using the lost connection listener is the better way to do it.
Actually when doing the change I just thought it is less code and looks nicer. :-D

PR #207, opened 2018-02-19
Voice Channel Join And Leave Events

Vampire:
Follow-up to #196 which fixes formatting and does the changes I suggested in that PR.


BtoBastian:
Did you test, if this works with private channels and group channels?

Vampire:
I tested with private channels, not with group channels.
It works for the join, but not for leave, as you don't get the event for leaves.
For leaves it would be CALL_DELETE or so, but I didn't put too much effort in non-bot usage, as bots can only do calls in `ServerVoiceChannels` currently anyway.
Bot accounts cannot join DMs and you can only call in private channels if you are befriended and you cannot befriend a bot account.

Vampire:
changes pushed

PR #208, opened 2018-02-23
[WIP] voice support

Vampire:
I played a bit with voice support and thought I share what I have so far if anyone wants to pick up on that.
Basically the happy-path of initiating a voice connection is done.
There are of course some events missing and some asynchronity missing and the API is not usage-ready.
I just tried to get it somehow working at all first.

It is in a state where the udp channel is connected and should be able to send and receive voice data,
but actually no bytes are received over the channel even if someone is speaking, only the SPEAKING
packets are received on the voice web socket if someone starts speaking.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -238,7 +238,9 @@
         <module name="EmptyCatchBlock">
             <property name="exceptionVariableName" value="ignored"/>
         </module>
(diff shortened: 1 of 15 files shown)


Bastian:
I think we can close this as the current `audio` branch is on a more recent state.
Please correct me if I am wrong and I repoen the PR :)

PR #211, opened 2018-02-25
Add the possibility to remove the server splash

Vampire:


src/main/java/de/btobastian/javacord/entities/ServerUpdater.java
```diff
@@ -551,7 +576,10 @@ public ServerUpdater removeAllRolesFromUser(User user, Collection<Role> roles) {
         if (icon != null) {
             patchServer = true;
         }
-        if (splash != null) {
+        if (updateSplash) {
+            if (splash != null) {
+                body.putNull("splash");
```
BtoBastian: I'm pretty the if statement here is wrong

Vampire: argh, copy & paste
1 sec

Vampire: fixed
that's the problem if you cannot test :-D

BtoBastian: That's why reviewing exists :-)

PR #217, opened 2018-02-26
Bulk Delete Improvement

Vampire:
Make TextChannel#bulkDelete methods automatically handle size and age restricitions of Discord API

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/DiscordEntity.java
+++ b/src/main/java/de/btobastian/javacord/entities/DiscordEntity.java
@@ -10,6 +10,19 @@ import java.time.Instant;
 public interface DiscordEntity {
 
     /**
+     * Calculates the creation date from the given Discord entity id.
+     *
+     * @param entityId The entity it to calculate from.
+     * @return The creation date of the Discord entity.
+     * @see <a href="https://discordapp.com/developers/docs/reference#snowflake-ids">Discord docs</a>
+     */
+    static Instant getCreationTimestamp(long entityId) {
+        // The first 42 bits (of the total 64) are the timestamp
+        // Discord starts its counter at the first second of 2015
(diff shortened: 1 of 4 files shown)


BtoBastian:
I'm usually not a fan of hard-codeding **any** limts/etc. at all, but in that case we can probably make an exception.
However, instead of replacing the bulk delete method with this, I would rather have a `DicordApi#deleteAll(...)` method and a `MessageSet#deleteAll()` method. These methods should also allow messages from different channels.

Vampire:
How about `DicordApi#deleteAll(...)`, `MessageSet#deleteAll()` and `TextChannel#deleteAll(...)` and removing the `bulkDelete` messages completely?
You usually want the user to not have to care about Discord craziness, so forwarding the limitation of 2 weeks and 2-100 messages to the user is not too much in that sense I'd say, that's why I made this PR. :-)

BtoBastian:
I still think there are usecases where you don't want to use the normal delete endpoint but just use the bulk delete endpoint and let it fail in case some of the messages are invalid.
There's no harm in keeping the bulk delete methods besides maybe a few users who didn't notice the difference between `deleteAll` and `bulkDelete`.

Vampire:
Ok, if you think so, I'll change it to that, while I still think this more produces confusion about why there is `bulkDelete` and `deleteAll` than there is a real use-case for `bulkDelete`.

Btw. I don't think `DicordApi#deleteAll(...)` is the right place, but `Message.deleteAll(...)` is, what do you think?

BtoBastian:
Imo it would make sense in both places.

Vampire:
We (I did and you accpeted) moved all the message-id stuff to static message methods, so it would be the only one in DiscordApi, wouldn't it?

BtoBastian:
Valid point.

Vampire:
New version pushed

PR #219, opened 2018-02-26
Use SoftReferences in the MessageCache to release memory if needed and make it possible to disable automatic cache cleanup

Vampire:


Bastian:
As this is a pretty old PR (sorry for that), I'll try to recap what effect this PR has:
### By default
Everything is like before, but now a memory-shortage will no longer end in a `OutOfMemoryError`, but first clean soft references in the message cache and warn the user.

### With automatic message cache cleanup _disabled_
It will cache all messages as long as possible. It will only cleanup the cache (GC removes soft references) if the bot runs out of memory.

Did I miss something?

Vampire:
Boah, don't ask me things like that after half a year.
I think yes, you are right.
And cache-forever messages are hardly referenced and thus are not affected by the out-of-memory cleanup.
And the auto-cleanup can be configured globally or per message cache.

Bastian:
Looks fine to me 👍

PR #231, opened 2018-03-01
Throttle web socket frame sending to ensure not hitting the 120/60 ratelimit (#354)

### LINKED ISSUE #354: Prevent hitting the websocket ratelimits
Javacord should automatically throttle websocket packets to prevent hitting the 120/60 ratelimit.
At the moment it's nearly impossible to hit this ratelimit, but it might become a problem once Javacord has audio support.

Vampire:
Fixes #354

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioWebSocketAdapter.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioWebSocketAdapter.java
@@ -54,7 +54,11 @@ public class AudioWebSocketAdapter extends WebSocketAdapter {
     public AudioWebSocketAdapter(AudioConnectionImpl connection) {
         this.connection = connection;
         this.api = (DiscordApiImpl) connection.getChannel().getApi();
-        this.heart = new Heart(api, websocket, true);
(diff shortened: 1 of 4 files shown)


Vampire:
new version pushed

Vampire:
I didn't test it on the audio branch as I don't use the audio branch yet, but I did my best to port over my changes properly.

Bastian:
https://github.com/Vampire/Javacord/blob/web-socket-frame-sending-throttling/javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java#L777
The `sendVoiceStateUpdate(...)` method still directly accesses the websocket.

Besides that, the code seems to work. 

Just one small suggestion: It's not optimal for debugging that the actual sending of packets does not get logged (besides trace). This can be a little bit confusing. I think it would be best to also log the frame when actually sending it and not only when it gets queued.

Vampire:
Ah, good catch, I didn't look for new code sending frames.
Fixed that and added two debug log lines, one for waiting, one for actual sending.

Bastian:
Thanks! :)

PR #232, opened 2018-03-02
Overwritten Permissions Retrieval

Vampire:
- Add ServerChannel#getOverwrittenPermissionsForUsers
- Add ServerChannel#getOverwrittenPermissionsForRoles
- Add ServerChannel#getOverwrittenPermissions

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannel.java
@@ -133,6 +133,13 @@ public interface ServerChannel extends Channel {
     Permissions getOverwrittenPermissions(Role role);
 
     /**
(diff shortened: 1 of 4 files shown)


BtoBastian:
I think `getOverwrittenUsersPermissions` instead of `getOverwrittenPermissionsForUsers` would be more consistent with the current method and field naming.

Do you have any reasons for choosing `getOverwrittenPermissionsForUsers` over `getOverwrittenUsersPermissions`? Personally I prefer the other option. Would you mind changing it before I merge the PR?

Vampire:
Ah, and I maybe didn't name it like that because such methods alread exist in the Impl

Vampire:
I changed it to `getOverwrittenUserPermissions` and `getOverwrittenRolePermissions` now, renaming the internal methods.

PR #235, opened 2018-03-03
Let the embed builder actually build an object.

BoneyIsSpooky:


BtoBastian:
Hey,
thank you, but that's not how it was intended.
* The embed interfaces should be split into two different classes `RichEmbed` and `Embed`, `RichEmbedAuthor` and `EmbedAuthor`, etc. The builder should produce "normal" embeds and the `Message#getEmbeds()` methods should return the "extended" embed interface, which has additional methods like `getProxyUrl()` which the embed builder cannot produce.
* Just allowing the embed builder to produce these objects doesn't provide any benefit. The whole "idea" behind it, was to have 
```java
// New
Embed embed = new EmbedBuilder().setTitle("Example").build();
channel.sendMessage(embed);
``` 
instead of passing the builder itself as a method parameter:
```java
// Old
EmbedBuilder embedBuilder = new EmbedBuilder().setTitle("Example");
channel.sendMessage(embedBuilder );
```

PR #236, opened 2018-03-04
Added Message#extractChannels() to get the Mentioned ServerTextChannels from a Message

DerTobi01:
Because I really wanna contribute x_x
lolz

### DIFF
--- a/src/main/java/de/btobastian/javacord/entities/message/Message.java
+++ b/src/main/java/de/btobastian/javacord/entities/message/Message.java
@@ -7,6 +7,7 @@ import de.btobastian.javacord.ImplDiscordApi;
 import de.btobastian.javacord.entities.DiscordEntity;
 import de.btobastian.javacord.entities.Server;
 import de.btobastian.javacord.entities.User;
+import de.btobastian.javacord.entities.channels.Channel;
 import de.btobastian.javacord.entities.channels.GroupChannel;
 import de.btobastian.javacord.entities.channels.PrivateChannel;
 import de.btobastian.javacord.entities.channels.ServerChannel;
@@ -982,6 +983,27 @@ public interface Message extends DiscordEntity, Comparable<Message> {
             emojis.add(emoji);
         }
         return Collections.unmodifiableList(emojis);
+    }
+
+    /**
+     * Gets a collection of all <code>ServerTextChannels</code> mentioned in the message.
+     *
+     * @return The collection of
+     */
+    default List<ServerTextChannel> extractChannels() {
+        String content = getContent();
+        List<ServerTextChannel> channels = new ArrayList<>();
+        List<String> matches = new ArrayList<>();
+        Optional<Channel> chlOpt;
+        Matcher matcher = Pattern.compile("(<#)[0-9]+(>)")
+                .matcher(content);
+        while (matcher.find()) {
+            matches.add(matcher.group());
+        }
+        for (String x : matches) {
+            getApi().getChannelById(x.substring(x.indexOf("<#")+2, x.indexOf(">"))).ifPresent(a -> a.asServerTextChannel().ifPresent(channels::add));
+        }
+        return channels;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/de/btobastian/javacord/entities/message/Message.java
```diff
@@ -984,6 +985,27 @@ default String getReadableContent() {
...
+     * Gets a collection of all <code>ServerTextChannels</code> mentioned in the message.
+     *
+     * @return The collection of
+     */
+    default List<ServerTextChannel> extractChannels() {
+        String content = getContent();
+        List<ServerTextChannel> channels = new ArrayList<>();
+        List<String> matches = new ArrayList<>();
+        Optional<Channel> chlOpt;
+        Matcher matcher = Pattern.compile("(<#)[0-9]+(>)")
```
xaanit: Why not do `"<#([0-9]+)>"`

src/main/java/de/btobastian/javacord/entities/message/Message.java
```diff
@@ -984,6 +985,27 @@ default String getReadableContent() {
...
+                .matcher(content);
+        while (matcher.find()) {
+            matches.add(matcher.group());
+        }
+        for (String x : matches) {
+            getApi().getChannelById(x.substring(x.indexOf("<#")+2, x.indexOf(">"))).ifPresent(a -> a.asServerTextChannel().ifPresent(channels::add));
```
xaanit: Using the aforementioned pattern you could just use the matcher.group(), you could also just merge the while and for loop.

src/main/java/de/btobastian/javacord/entities/message/Message.java
```diff
@@ -984,6 +985,25 @@ default String getReadableContent() {
...
+    default List<ServerTextChannel> extractChannels() {
+        String content = getContent();
+        List<ServerTextChannel> channels = new ArrayList<>();
+        String x;
+        Optional<Channel> chlOpt;
+        Matcher matcher = Pattern.compile("<#([0-9]+)>")
+                .matcher(content);
+        while (matcher.find()) {
+            x = matcher.group();
+            getApi().getChannelById(x.substring(x.indexOf("<#")+2, x.indexOf(">"))).ifPresent(a -> a.asServerTextChannel().ifPresent(channels::add));
```
xaanit: You didn't change this at all lol, x will only be the numbers with my pattern.

DerTobi01: it works like this 🤔 
grafik

xaanit: Cause regex is weird.

Using just `group()` will grab the whole string (group(0)). Use `.group(1)` instead and you can eliminate `x` entirely by just calling the method in the `getChannelById`

BtoBastian:
Hey,
thank you, but please take a look at Javacord3's [contribution guidelines](https://github.com/BtoBastian/Javacord3-Docs/blob/master/.github/CONTRIBUTING.md).
There a serveral issues with this PR:
* Instead of compiling your own regex pattern every time, you should use the precompiled pattern from [DiscordRegexPattern.java](https://github.com/BtoBastian/Javacord/blob/v_3/src/main/java/org/javacord/util/DiscordRegexPattern.java)
* The code doesn't follow Javacord's code conventions (e.g. calling a variable just `x`)
* The JavaDoc comment is just a half sentence (`@return The collection of`)
* All returned collections in Javacord should be immutable
* The method name doesn't really fit to the other used methods names (e.g. `#getMentionedUsers()`, `#getMentionedRoles()`, `#getCustomEmojis()`, ...)
* You can mention voice channels, even though they are not displayed in the auto completion of the client (not sure about categories).

DerTobi01:
@BtoBastian when editing the stuff to what you needed, i realized this is already in there xd didnt notice.

/closed

Vampire:
And how is it different from `Message#getMentionedChannels`? (Didn't look too close at the changes)

BtoBastian:
I didn't even notice that this method already exists.
Just assumed it doesn't. ^^

FunkFrog:
_why._

PR #240, opened 2018-03-05
Split BotInviteBuilder into implementation and interface

Vampire:


src/main/java/org/javacord/ImplBotInviteBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public ImplBotInviteBuilder(String clientId) {
+        try {
+            this.clientId = Long.parseLong(clientId);
+        } catch (NumberFormatException e) {
+            this.clientId = 0;
+        }
+    }
+
+    @Override public BotInviteBuilder setPermissions(Permissions permissions) {
```
BtoBastian: https://github.com/BtoBastian/Javacord3-Docs/blob/master/.github/CONTRIBUTING.md#code-conventions

Vampire: Argh, forgot to format, sry. IJ does not do it according to formatting settings. :-(

Vampire: fixed

PR #245, opened 2018-03-08
Fix exception validation

Vampire:
The exception validation needs to stay in the `DiscordException` constructor, or you miss the "there is no actual result" case.
As the `FactoryBuilder` now not only delivers factories, maybe we should find a better name?
`ServiceLocator`?
`HelperLocator`?
`UtilitiesLocator`?
`UtilsLocator`?


Vampire:
Incorprated into #249

PR #250, opened 2018-03-17
Gradle and API-split

Vampire:
You can see the result at
https://circleci.com/gh/Vampire/Javacord/39#artifacts/containers/0
https://jitpack.io/com/github/Vampire/Javacord/gradle-SNAPSHOT/build.log
https://jitpack.io/com/github/Vampire/Javacord/javacord/gradle-SNAPSHOT/
https://jitpack.io/com/github/Vampire/Javacord/javacord-api/gradle-SNAPSHOT/
https://jitpack.io/com/github/Vampire/Javacord/javacord-core/gradle-SNAPSHOT/

In the javacord one you find the shaded JAR and a POM that has api as compile and core as runtime dependency.
That means you can simply depend on `com.github.Vampire.Javacord:javacord:gradle-SNAPSHOT` and your dependencies are like we wanted it, with API on the compile classpath and core and its dependencies only on the runtime classpath.
Actually it is like that if you use Maven or if you use Gradle with a special feature preview that will become the default with 5.0. :-)
Before and without the feature preview enabled Gradle had put them both on the compile classpath.


BtoBastian:
Thank you. :)
Just some minor changes:
- I think it would be better to have `org.javacord.api` for the api artifact instead of just `org.javacord`
- It's just a very small detail, but I changed the `simple to use` to `easy to use` in the readme and it's probably best to apply this change to the build.gradle, too

Otherwise everything looks fine.

Vampire:
- I changed the descriptions
- I fixed a merge error that was in the intermediary commit
- replaced the pom configuration with a plugin that makes the syntax much nicer (less indentation, no `delegate.` necessary)

And I added additionally:
- restriction to two Gradle workers on CircleCI as the CPU detection is not working properly on their infrastructure
- force JavaDoc to be built with at least Java 9 to get a nicer doclet with HTML 5 support and a nifty search box in the upper right (you can set a Gradle project property to allow building JavaDoc with Java 8 though)
- some JavaDoc options to have a split index and the usage pages, document `@author` and `@version` tags if they are present, generate HTML 5 if building with Java 9 and link JRE classes to online JRE JavaDoc
- fixed JavaDoc that produced invalid HTML
- moved the not-to-be-used classes in the API package to `internal` sub-packages to make it even clearer to the user that he shouldn't use them and to make it possible to encapsulate them away when used as Java 9 module and to have two separate Groups in JavaDoc, `Public API` and `Internal Helpers`
- made the produced JARs proper multi-release JARs that have a `module-info.class` for Java 9 and newer that make them proper Jigsaw modules that encapsulate packages properly if built with Java 9 or newer, otherwise just the `Automatic-Module-Name` is set in the manifest and a warning emitted.

Vampire:
I've seen in #rewrite-discussion that you indeed mean the package name.
Ok, I'll change that.
What do you think about the internal packages?
Better
```
org.javacord.api.entity.channel.internal.ServerVoiceChannelUpdaterDelegate
org.javacord.api.entity.emoji.internal.CustomEmojiBuilderDelegate
...
```
or better
```
org.javacord.api.internal.entity.channel.ServerVoiceChannelUpdaterDelegate
org.javacord.api.internal.entity.emoji.CustomEmojiBuilderDelegate
...
```
?

Vampire:
The package renaming is also done now

BtoBastian:
Moving the delegates into their own `internal` package was a good idea. 👍 
I think it's fine to have `org.javacord.api.entity.emoji.internal.CustomEmojiBuilderDelegate` like we had the `impl` before the split.
The PR looks fine for me now.

PR #253, opened 2018-03-28
Catch all throwables in repeating tasks or they stop repeating

Vampire:


javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -422,25 +422,33 @@ public DiscordApiImpl(
...
-                     userRef != null;
-                     userRef = usersCleanupQueue.poll()) {
-                    Long userId = userIdByRef.remove(userRef);
-                    if (userId != null) {
-                        users.remove(userId, userRef);
+                try {
+                    for (Reference<? extends User> userRef = usersCleanupQueue.poll();
+                         userRef != null;
+                         userRef = usersCleanupQueue.poll()) {
+                        Long userId = userIdByRef.remove(userRef);
```
BtoBastian: The indentation seems to be wrong here

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -422,25 +422,33 @@ public DiscordApiImpl(
...
-                     messageRef != null;
-                     messageRef = messagesCleanupQueue.poll()) {
-                    Long messageId = messageIdByRef.remove(messageRef);
-                    if (messageId != null) {
-                        messages.remove(messageId, messageRef);
+                try {
+                    for (Reference<? extends Message> messageRef = messagesCleanupQueue.poll();
+                         messageRef != null;
+                         messageRef = messagesCleanupQueue.poll()) {
+                        Long messageId = messageIdByRef.remove(messageRef);
```
BtoBastian: Here, too

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
Vampire: Define "wrong" please.
It is exactly 4 spaces more than before.
If is is wrong, it was wrong before already.
Or did you overlook that the for-loop is split into three lines?

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
Vampire: Here, too

PR #263, opened 2018-04-03
Some #equals Methods, that are safe to use.

Kaleidox00:
because i needed them, lol


BtoBastian:
Thank you, but all Discord entities already override the `#equals(Object)` method properly: https://github.com/BtoBastian/Javacord/commit/9edc6c9e42641e919e8486c57ba865479e67578d

PR #268, opened 2018-04-08
(Re-Open) Added Method Message#isPrivate

Kaleidox00:
Added Method Message#isPrivate which returns a boolean depending on if the Message was sent in a Private Channel (-> DM) or not.


BtoBastian:
Thank you :)

Kaleidox00:
i am finally a part of javacord lul :p

PR #269, opened 2018-04-09
Fix ban method ignoring the reason parameter

DoNotSpamPls:
If you don't know what the bug is, basically reasons didn't get attached to bans, either in the `Bans` list or in the audit log. I'm *not* sure if the setAuditLogReason line is required, or if addQueryParameter does it too, but I've tested this fix and it works just fine 😄


BtoBastian:
Hey,
thank you for the PR.
Would you mind testing, if `setAuditLogReason ` is required and if not, remove it?

DoNotSpamPls:
Done!

DoNotSpamPls:
Done!

BtoBastian:
Thx :)

PR #270, opened 2018-04-12
Add support for member count via invite.

Saladoc:
Quick stab at an item from the ToDo List.

Currently this change will always request counts when grabbing an invite.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/invite/Invite.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/invite/Invite.java
@@ -117,4 +117,18 @@ public interface Invite {
      */
     CompletableFuture<Void> delete(String reason);
 
+    /**
+     * Gets the approximate number of members for the target server.
+     *
+     * @return The count of members, if available.
+     */
+    Optional<Integer> getApproximateMemberCount();
+
+    /**
+     * Gets the approximate number of online members for the target server.
+     *
+     * @return The count of present members, if available.
+     */
+    Optional<Integer> getApproximatePresenceCount();
+
 }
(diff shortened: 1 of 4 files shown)


javacord-api/src/main/java/org/javacord/api/entity/server/invite/Invite.java
```diff
@@ -117,4 +118,18 @@ default URL getUrl() {
      */
     CompletableFuture<Void> delete(String reason);
 
+    /**
+     * Gets the approximate number of members for the target server.
+     *
+     * @return The count of members, if available
+     */
+    OptionalInt getApproximateMemberCount();
```
Kaleidox00: I think the "OptionalInt" makes a big inconsistency here.

BtoBastian:
Hey,

thank you for the PR!

I asked in the DAPI server about the impact of always requesting the counts:


It would be better to make the requesting optional. There's just the question on how to implement it:
- A boolean parameter? -> I usually try to avoid unintuitive parameters 
- An extra methods?  -> Like the `getXyzByName(...)` and `getXyzByNameIgnoreCase(...)` methods.
- Some kind of builder patterm?

Saladoc:
I've went with the extra method for consistency.

javacord-api/src/main/java/org/javacord/api/entity/server/invite/Invite.java
```diff
@@ -117,4 +117,18 @@ default URL getUrl() {
...
+     * Gets the approximate number of members for the target server.
+     *
+     * @return The count of members, if available
+     */
+    Optional<Integer> getApproximateMemberCount();
+
+    /**
+     * Gets the approximate number of online members for the target server.
+     *
+     * @return The count of present members, if available
```
BtoBastian: This will probably fail to build after your PR #276 because of a missing dot.
Could you please update your fork and verify that all checks pass?

BtoBastian:
You forgot to remove the `with_counts` parameter in the "normal" method.

BtoBastian:
Thx for the quick fix :)

PR #275, opened 2018-04-19
Add some Message#has[...] Methods

Kaleidox00:
Felt like those simplifying methods would come in handy.

The way I implemented them is questionable tho, correct me please.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -1030,6 +1030,60 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     }
 
     /**
+     * Returns whether or not the Message has any attachment
+     *
+     * @return True, when the message has any attachment
+     */
+    default boolean hasAttachments() {
+        return getAttachments().isEmpty();
+    }
+
+    /**
+     * Returns whether or not the Message has any reactions
+     *
+     * @return True, when the message has any reactions
+     */
+    default boolean hasReactions() {
+        return getReactions().isEmpty();
+    }
+
+    /**
+     * Returns whether or not the Message has any embeds
+     *
+     * @return True, when the message has any embeds
+     */
+    default boolean hasEmbeds() {
+        return getEmbeds().isEmpty();
+    }
+
+    /**
+     * Returns whether or not the Message has any users mentioned
+     *
+     * @return True, when the message has any users mentioned
+     */
+    default boolean hasUserMentions() {
+        return getMentionedUsers().isEmpty();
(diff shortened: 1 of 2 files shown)


BtoBastian:
Hey,
thank you for the PR. :)

Is there really a need for such methods? I'm not sure if there are a lot of benefits of `message.hasAttachments()` vs `message.getAttachments().isEmpty()`. -> I'll ask for other opinions in #rewrite_discussion

About the implementation: All the `getAttachments()` etc. methods create new immutable collections. It would be better to use the internal object for the check to prevent this overhead.

Kaleidox00:
Well, the internal Objects are what I dont know of where it is. This is how far I could do it and why I asked for correction. (beginner tho)

Vampire:
To use the internal object you have to be internal of course.
You would not write default implementations, but only specify the method in the interface and implement it in its `MessageImpl` implementation where you have access to the internal objects.

Kaleidox00:
Changed 👍 

(removed `hasChannelMentions` tho because i dont know where channelMentions are coming from?)

BtoBastian:
After thinking about it for a while and asking in our Discord server, I don't think we need these methods.
They might be convenient in some cases, but I'm unsure if these rare cases justify the extra bloat.

I won't merge this PR for these reasons, but thank you nevertheless. Contributions are always welcome!

Some quick notes for future PR though:
- Please try to keep the style of your JavaDoc comments consistent with the other JavaDoc comments of the project. So for example, instead of 
  ```java
  /**
   * Returns whether or not the Message has any reactions
   *
   * @return True, when the message has any reactions
   */
  boolean hasReactions();
  ```
  it should be
  ```java
  /**
   * Checks if the message has any reactions.
   *
   * @return Whether the message has any reactions or not.
   */
  boolean hasReactions();
  ```
  Just search for similar methods and copy their style. :)
- Before creating a PR for a new feature, I usually recommend to ask in the Discord server first. In case the feature isn't wanted, you didn't waste your time implementing it. It's not mandatory, but especially for larger PRs I highly recommend it.

Kaleidox00:
aighto, figured it. 

afterall, by implementing useless things you might still learn new things, and i learned a whole new stuff like this 🤔

BtoBastian:
It's just a recommendation. I don't mind reviewing PRs. In the end it doesn't matter for me if we discuss a new feature before or after implementing it. I just thought it would save you some time, but I guess for learning it's fine. 😃

WalkerKnapp:
Bringing up new things is always good. :) For instance, PR #277 has about a 0% chance of getting merged, but I still did it for the sake of getting people talking about it.

Kaleidox00:
that sneaky self advertisement /lenny

PR #276, opened 2018-04-20
Add checkstyle config and refactor existing code to conform to it

Saladoc:
I took the [Google checkstyle config](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml) and applied the changes noted at the [Contribution Guidelines](https://github.com/BtoBastian/Javacord3-Docs/blob/master/.github/CONTRIBUTING.md).

The initial run yielded just about 780 violations and I've since identified the most common ones. Most have been addressed by reformatting / rearranging the existing code as was discussed on Discord. I've tried to group the changes (which I did file by file) into one commit per issue to allow for easier review, they can however be squashed before merging.

To do: 
 - [x] Fix the failing `VariableDeclarationUsageDistanceCheck`s or drop the check from the Styleguide
 - [x] Create Javadoc comments for the [RateLimitBucket Constructors](https://github.com/BtoBastian/Javacord/blob/v_3/javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java#L25)
 - [ ] Maybe add some more integration with TeamCity so we have neat reports / a nice indicator on whether the code passes all inspections? I had hoped Vampire could take care of that as the resident gradle wizard

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ def rootBuildDir = buildDir
 allprojects {
     apply plugin: 'java-library'
     apply plugin: 'nebula.java-cross-compile'
+    apply plugin: 'checkstyle'
 
     group = 'org.javacord'
     description = 'A multithreaded but easy to use Discord Java API'
@@ -25,6 +26,11 @@ allprojects {
     tasks.withType(JavaCompile) {
         options.encoding 'UTF-8'
         options.incremental true
+    }
+
+    checkstyle {
+        toolVersion '8.9'
+        maxWarnings 0
     }
 }
 
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -1,0 +1,248 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+        "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
+
+<!--
+    Checkstyle configuration that checks the Javadoc coding conventions.
+    This is based on the Google checkstyle config by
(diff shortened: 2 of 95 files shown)


Saladoc:
I've removed the `VariableDeclarationUsageDistance` check since there was no basis for it to be found in https://google.github.io/styleguide/javaguide.html

Vampire:
I don't know how Bastian thinks about it, but I personally hate non-building commits.
If you check one out, you cannot build.
With tools like `git bisect` this becomes even worse, it is a great tool to find the commit that broke something, but it cannot work properly if the commits are not building properly.

Would you mind doing some slight changes to your commits, a) to make them buildable and b) to fit better to the other build script?
If you don't know how to post-process Git commits, here is a nice cheat-sheet: http://sethrobertson.github.com/GitPostProduction :-)

- Your first commit does not build because of the aged checkstyle version, please move the `apply plugin: 'checkstyle'` to the second commit, or the `toolVersion = "8.9"` to the first commit
- remove `configDir = rootProject.file('config/checkstyle')` from the second commit, it is not necessary anymore as we are on Gradle 4.8 now and this value is the default since 4.7
- Please use single quotes for the `toolVersion`
- Please remove the equals sign from the `toolVersion` line
- Please remove the empty line after the `checkstyle` block
- As noone looks at warnings if they don't make a build fail, but Bastian obviously does care about them, or he wouldn't have the rule and asked you to fix them, please add `maxWarnings 0` in an additional final commit (not in the starting commits, because then all other commits are not buildable anymore)

The `checkstyle` block should now look like
```groovy
checkstyle {
    toolVersion '8.9'
    maxWarnings 0
}
```

BtoBastian:
Thank you for the PR and applying the requested changed! 👍

PR #277, opened 2018-04-21
[WIP] Replaced usage of nv-websocket-client with OkHttp's websocket

WalkerKnapp:
One dependency down, 4 to go! /s

Most of the changes in this PR are self-explanatory, just moving over to the new system, although a notable change is how `INVALID_SESSION`s are delt with. The previous logic would not have worked with OkHttp's websocket (It doesn't give a method to know when frames are sent), so it would have to change. This version is not as accurate as the old logic due to the queuing of frames potentially throwing off the timing, thus this PR being WIP. It is definitely in a functional state, but this shouldn't be merged until a better solution is found for this.


Kaleidox00:
Wasnt Javacord at okHTTP before and switched for some reason?

WalkerKnapp:
I might be wrong, but I think it has been like this at least since v2. Idk though

Kaleidox00:
grafik
my bad, was the other way round.

WalkerKnapp:
Added proxy support. This can easily be bundled with this PR, so there is no downside to doing it here.

Saladoc:
What became of this? It was a pretty good idea, but how finished is it?

Vampire:
iirc there is an open issue about not determining / distinguishing invalid session vs. identify rate limit exeeded accurately or something like that.

WalkerKnapp:
All of the features should be working, there are just slight timing differences. To be honest, this isn't going to work unless okhttp updates their API. Closing this, potentially to be opened at a future date when that happens.

PR #280, opened 2018-05-03
Added MessageCreateEvent#getMessageAuthor()

Kaleidox00:
Felt like this would come in handy dandy, when I caught myself not needing a Message but the MessageAuthor of the event. /shrug

Did I Implement this correctly?


Kaleidox00:
let me work on this later when i can focus.

PR #283, opened 2018-05-06
Added Server#getHighestRole(User) and #getColor(User)

Kaleidox00:
Used different Method Names from those 
grafik because those felt Awkward

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -88,7 +88,7 @@ import org.javacord.api.listener.user.UserChangeStatusListener;
 import org.javacord.api.listener.user.UserStartTypingListener;
 import org.javacord.api.util.event.ListenerManager;
 
-import java.awt.*;
+import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.InputStream;
@@ -98,6 +98,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -451,26 +452,20 @@ public interface Server extends DiscordEntity, UpdatableFromCache<Server> {
     default Optional<Role> getHighestRole(User user) {
         List<Role> collect = new ArrayList<>(this.getRolesOf(user));
 
-        return Optional.ofNullable(collect.get(collect.size()-1));
+        return Optional.ofNullable(collect.get(collect.size() - 1));
     }
 
     /**
-     * Gets the displayed Color of an User.
+     * Gets the displayed color of an user.
      *
-     * @param user The user whose Color is to be found
-     * @return The colour of the User in this guild.
+     * @param user The user whose color is to be found
+     * @return The color of the user in this guild.
      */
-    default Color getColor(User user) {
-        List<Color> collect = this
-                .getRolesOf(user)
-                .stream()
-                .map(Role::getColor)
-                .filter(Optional::isPresent)
-                .map(Optional::get)
-                .peek(System.out::println)
-                .collect(Collectors.toList());
-
-        return (collect.size() > 0 ? collect.get(collect.size()-1) : Color.WHITE);
+    default Optional<Color> getRoleColorOf(User user) {
+        return getRolesOf(user).stream()
+                .filter(role -> role.getColor().isPresent())
+                .max(Comparator.comparingInt(Role::getPosition))
+                .flatMap(Role::getColor);
     }
 
     /**
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -88,11 +88,13 @@ import org.javacord.api.listener.user.UserChangeStatusListener;
 import org.javacord.api.listener.user.UserStartTypingListener;
 import org.javacord.api.util.event.ListenerManager;
 
+import java.awt.*;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.InputStream;
 import java.net.URL;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


Kaleidox00:
Fix coming, Color cannot be cast to Comparable

Kaleidox00:
Fixed 👍

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -88,11 +88,13 @@
 import org.javacord.api.listener.user.UserStartTypingListener;
 import org.javacord.api.util.event.ListenerManager;
 
+import java.awt.*;
```
BtoBastian: Can you please update your fork to the latest version of Javacord?
This should make the build fail in later version because of a checkstyle violation.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -440,6 +442,37 @@ default String getDisplayName(User user) {
...
+        List<Color> collect = this
+                .getRolesOf(user)
+                .stream()
+                .map(Role::getColor)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .peek(System.out::println)
+                .collect(Collectors.toList());
+
+        return (collect.size() > 0 ? collect.get(collect.size()-1) : Color.WHITE);
```
BtoBastian: Why is this in brackets?

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -440,6 +442,37 @@ default String getDisplayName(User user) {
...
+     * @return The colour of the User in this guild.
+     */
+    default Color getColor(User user) {
+        List<Color> collect = this
+                .getRolesOf(user)
+                .stream()
+                .map(Role::getColor)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .peek(System.out::println)
```
BtoBastian: Please remove this

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -339,6 +341,37 @@ default boolean isYourself() {
...
+    /**
+     * Gets the highest Role of the User in a server.
+     *
+     * @param server The Server on which the highest role should be found.
+     * @return An Optional that may contain the highest role of the user in the given server.
+     */
+    default Optional<Role> getHighestRole(Server server) {
+        List<Role> collect = new ArrayList<>(server.getRolesOf(this));
+
+        return Optional.ofNullable(collect.get(collect.size()-1));
```
BtoBastian: This will cause an exception for users which are not in the server, as the list would be empty.

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -339,6 +341,37 @@ default boolean isYourself() {
...
     }
 
+    /**
+     * Gets the highest Role of the User in a server.
+     *
+     * @param server The Server on which the highest role should be found.
+     * @return An Optional that may contain the highest role of the user in the given server.
+     */
+    default Optional<Role> getHighestRole(Server server) {
+        List<Role> collect = new ArrayList<>(server.getRolesOf(this));
```
BtoBastian: Why do you create a new list here?

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -440,6 +442,37 @@ default String getDisplayName(User user) {
                         .collect(Collectors.toList()));
     }
 
+    /**
+     * Gets the highest Role of a given User.
```
BtoBastian: Please check whether a word should be lower or upper case. This should be `Gets the highest role of the given user.`.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -440,6 +442,37 @@ default String getDisplayName(User user) {
...
+     *
+     * @param user The user whose Role is to be found
+     * @return An Optional that may contain the highest Role of the given User in this guild.
```
BtoBastian: You can usually just ignore that it returns an Optional in JavaDoc comments (unless it's not clear in which cases it returns an Optional). If you look at other methods which return an Optional, you will see the difference. This should just be `The highest role of the given user in this server.`. Also in Javacord it's called `server` not `guild`.

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -339,6 +341,37 @@ default boolean isYourself() {
...
+
+        return Optional.ofNullable(collect.get(collect.size()-1));
+    }
+
+    /**
+     * Gets the displayed Color of the User in a server.
+     *
+     * @param server The server on which the displayed Color should be found.
+     * @return The colour of the user in the given server.
```
BtoBastian: Please use American English, not British English.

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -339,6 +341,37 @@ default boolean isYourself() {
...
+        List<Color> collect = server
+                .getRolesOf(this)
+                .stream()
+                .map(Role::getColor)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .peek(System.out::println)
+                .collect(Collectors.toList());
+        
+        return (collect.size() > 0 ? collect.get(collect.size()-1) : Color.WHITE);
```
BtoBastian: Instead of having a default color, the method should return an Optional. There's no default color in Discord.
Also I don't really like the implementation. I would suggest the following implementation:
```java
/**
 * Gets the color of the given user in this server.
 * The Optional is empty, if the user is not a member of the server or of the user has no role with a custom color.
 *
 * @param user The user.
 * @return The color of the user.
 */
default Optional<Color> getColorOf(User user) {
    List<Role> roles = getRolesOf(user);
    // Iterate in reverse order
    for (int i = roles.size() - 1; i <= 0; i--) {
        Role role = roles.get(i);
        Optional<Color> roleColor = role.getColor();
        if (roleColor.isPresent()) {
            return roleColor;
        }
    }
    return Optional.empty();
}
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -440,6 +442,37 @@ default String getDisplayName(User user) {
...
+     */
+    default Optional<Role> getHighestRole(User user) {
```
BtoBastian: I just looked at the code and saw, that there's already a `Server#getHighestRoleOf(User)` method. My bad for adding it on the TODO list.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
Kaleidox00: "bad habit" of myself, will fix

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
Kaleidox00: big whoopsie

Kaleidox00:
Updated things

PR #290, opened 2018-05-14
Tests / getAllServers / getAllServersById

Vampire:
- Enable tests with Spock and JaCoCo coverage
- Fix DiscordApiImpl#getAllServers only returning non-ready servers
- Test that DiscordApiImpl methods that return a collection return unmodifiable collections
- Rename DiscordApiImpl#getAllServerById to DiscordApiImpl#getPossiblyUnreadyServerById

Before we forget it, I thought I make a PR for it as reminder,
we can discuss about the new name of the renamed method of course,
or rather you can come up with a better name. :-)

I also took the erroneous `getAllServers` method as chance to start writing tests for Javacord.
I really fell in love with Spock when I started to use it and got used to its syntax.
You can read more about it at http://spockframework.org/spock/docs/1.1/
I also measure code coverage with JaCoCo which you can produce a report of with a Gradle task and also nicely displayed in TeamCity.


BtoBastian:
Any specific reasons why you prefer Spock? How well known is it?
I feel like it's better for open source project with multiple developers to stick with more common frameworks, in that case JUnit. This lowers the barrier for contributing to Javacord as people are more likely to already have experience with JUnit than with Spock + if they don't have experience with both, it's probably easier to get started with JUnit as you write the tests in Java and not in Groovy.

Vampire:
> Any specific reasons why you prefer Spock?

Because it is really awesome. :-D
You can write much more concise and expressive what you actually want to test with much less boiler-plate.
It also has very nicely integrated mocking, stubbing, behaviour driven testing ("if I call this, then that method has to be called twice and this one once"), amazing data-driven tests (especially compared to JUnit 4, not sure about 5, this is "do this test with this dataset and this dataset and this dataset" and similar).
For example look at the test specification included in this PR.
This is actually 44 individual isolated tests.

Also the assertions in case of a test failure are amazingly helpful compared to JUnit.
You can improve there with JUnit by using hamcrest matchers, but in Spock you just write simple boolean statements and you get as a result e. g. 

```
Condition not satisfied:

api.allServers.size() == 2
|   |          |      |
|   |          1      false
|   [Mock for type 'Server' named 'readyServer']
org.javacord.core.DiscordApiImpl@21337f7b

Expected :2

Actual   :1
```

or

```
Condition not satisfied:

api.allServers.containsAll readyServer, nonReadyServer
|   |          |           |            |
|   |          false       |            Mock for type 'Server' named 'nonReadyServer'
|   |                      Mock for type 'Server' named 'readyServer'
|   [Mock for type 'Server' named 'readyServer']
org.javacord.core.DiscordApiImpl@60df60da
```

Actually before I first used it I prevented to send PRs to Gradle that needed tests, as I was pretty sceptical about Spock.
But then I had one PR where I had to add a test and read the userguide which isn't too long.
After that I fell in love with Spock and would prefer always being able to use Spock for testing. :-)

> How well known is it?

I'm not fully sure about its well-known-ness.
It is developed by a bunch of the core Gradle guys and the whole Gradle project is tested with Spock.

> I feel like it's better for open source project with multiple developers to stick with more common frameworks, in that case JUnit. This lowers the barrier for contributing to Javacord as people are more likely to already have experience with JUnit than with Spock

As Spock tests are running with a JUnit runner, you can easily have both, native JUnit tests and Spock tests and either keep with that or somewhen in the future transform the JUnit tests to Spock tests too. While I would prefer only accpeting Spock tests. :-D But both will be run in the same test run and are siblings as if only one kind of tests is used.

>if they don't have experience with both, it's probably easier to get started with JUnit as you write the tests in Java and not in Groovy.

I don't thinks so.
Spock is much more easily usable, you don't need to know how to use assertions, tests are written more naturally readable and you simply write some expressions and some boolean statements for simple tests and for more complex tests you can do complex Java or you can learn some Groovy and write it a bit nicer.

Also, one of the main goals of Groovy was to be **the** scripting language for Java and thus almost all Java 7 code is also valid Groovy code (they work on Java 8 syntax support afair).
So actually writing Groovy can be done more or less like writing Java, but if you know Groovy, you can use some syntactic sugar to write the code more concise and expressive. But this is mostly optional.

In my spec included in this PR I of course use quite some of that syntactic sugar, as I know Groovy pretty well, but instead of

```groovy
def 'getAllServers returns all servers'() {
    given:
    def readyServer = Mock(Server)
    def nonReadyServer = Mock(Server)
    api.@servers << [0: readyServer]
    api.@nonReadyServers << [1: nonReadyServer]

    expect:
    with (api.allServers) {
        it.size() == 2
        it.containsAll readyServer, nonReadyServer
    }
}
```
you could e. g. also have written
```java
void getAllServers_returns_all_servers() {
    given:
    Server readyServer = Mock();
    Server nonReadyServer = Mock();
    api.@servers.put(0L, readyServer);
    api.@nonReadyServers.put(1L, nonReadyServer);

    expect:
    api.getAllServers().size() == 2;
    api.getAllServers().containsAll(Arrays.asList(readyServer, nonReadyServer));
}
```
which is almost normal Java code, just some labels and and some easy statements.
The only Groovy-specific thing is the `.@` which allows to directly access the field and bypass the likewise named getter that Groovy otherwise would use.

PR #291, opened 2018-05-14
Listener Manager Generation

Vampire:
- Add a generic generate task to the root project that depends on all generation tasks in all projects
- Generate module-info.java for javacord-api with Javaparser instead of text manipulation
- Rename CustomEmoji events and listeners to KnownCustomEmoji events and listeners
- Make CachedMessageUnpinListener implement MessageAttachableListener
- Generate listener manager methods into own interfaces
- Generate listener manager method implementations into own interfaces
- Generate listener retrieval methods into utility class
- move channel, text channel and voice channel attachable listener into channel sub-package

With this change I remove all the boiler-plate you have to write for registering and retrieving listeners.
This also minimizes the risk of doing it wrongly or forgetting something like marking a listener as XXXAttachableListener
or forgetting to retrieve the listener from an entity where you can attach it.

As all the listener manager methods and implementations get generated into own interfaces,
this not only makes the classes much leaner and concise, but also the JavaDocs much better readable as all the listener-specific
stuff is not shown in the normal entity-methods anymore but in the implemented interfaces.

With this change adding a new listener or making a listener attachable to an additional entity is just a matter of letting the
listener extend the according attachable interface and generating the code, everything else is generated automatically.

There is a task called `generate` now too, that can be called manually to do all code generation tasks,
or to depend on from within IntelliJ to be executed when refreshing the project or when rebuilding the project.
This can be done from the Gradle tool window in IntelliJ, then the generation is automatically done on certain
situations you define, e. g. before "Make" is done or before "Rebuild Project" is done.

I also move the three channel attachable listeners into the `channel` package.
It was most probably a mistake that those were outside.


Vampire:
I now also added generation for the listener manager methods and implementations for `UncachedMessageUtil` and `UncachedMessageUtilImpl`.
Thereby I also removed the methods in `UncachedMessageUtil` and the static methods that received `Message` as parameter.
If you have a `Message` instance anyway, there is no sense or benefit in using a static method or method meant for _uncached_ messages, but you can use the instance methods directly.

matozoid:
Awesome!

gradle/listener-manager-generation.gradle
```diff
@@ -0,0 +1,1154 @@
...
+    internalAttachableListenerManagerFile.addImport listener.qualifiedName
+
+    if (specialMessageAttachable) {
+        addMessageSpecificConcreteListenerMethods typeSolver, attachableInterface, listener,
+                attachableListenerManagerFile, attachableListenerManagerInterface
+
+        uncachedMessageAttachableListenerManagerInterface
+                .addMethod("add$listener.name")
+                .setJavadocComment("""
+                    Adds a {@link $listener.name}.
```
Bastian: There shouldn't be a link.

Vampire:
I replaced almost all generated `@link` by `@code` now, only 4 left where I think they should stay.
- `This class can be used to add and retrieve {@link ${attachableInterface.name}}s.`
- `This class can be used to add and retrieve {@link MessageAttachableListener}s for potentially uncached messages.`
- `The id of the {@link Message}.`
- `The id of the {@link User}.`

PR #299, opened 2018-05-20
Added method Messageable#sendMessage(Object... items)

Kaleidox00:
Because I needed a way of posting both an EmbedBuilder and a String without having a strict type (did that make sense)


BoneyIsSpooky:
If your use case is sending Strings, Embeds, and Files together, why not just use a MessageBuilder? 
> first of all i felt like a Object-sending method was missing

Why? As Panda said, this is generally bad practice. I don't think this is necessary at all.

> secondly,i think it could be handy to have a "type safe" method, is this understandable

No, that doesn't make sense. Accepting Object and downcasting is about as far away from type safety you can get in Java. Not to mention that your implementation will silently ignore any Object that is not of the three accepted types, which reminds me of JS.

Kaleidox00:
yea agreed, figured that by now

PR #301, opened 2018-05-22
Make Javacord artifacts OSGi ready

Vampire:


cap5lut:
Checked some basic stuff like echoing messages and sharding. Because the nature of osgi itself, at that point there should not be any osgi related issues.

PR #302, opened 2018-05-23
Simplify asXyz methods

cap5lut:
This will remove the maintenance overhead for the `Optional<Xyz> asXyz()` methods in `Channel` and `Emoji`.

It also introduces an internal `ChannelBase` interface, to make the API of the `Channel` interface more consistent.

For even more consitency I think it would be better to rename `Emoji.asUnicodeEmoji()` simply to `Emoji.asUnicode()` or `Emoji.asString()` (I would prefer the latter) (this is not part of the PR)

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/util/Specifiable.java
+++ b/javacord-api/src/main/java/org/javacord/api/util/Specifiable.java
@@ -7,7 +7,7 @@ import java.util.Optional;
  * This interface is used to provide a more stream lined way to interact with the instance as sub types.
  * @param <S> Base instance type.
  */
-public interface Specializable<S> {
+public interface Specifiable<S> {
 
     /**
      * Get an {@link Optional} instance of an instance as specified type.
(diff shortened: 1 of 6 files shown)


javacord-api/src/main/java/org/javacord/api/util/Specifiable.java
```diff
@@ -0,0 +1,45 @@
...
+     * @param type Specified type, this instance will be tried to casted to.
+     * @param instance Instance which will be tried to get casted.
+     * @param <T> Specified type.
+     * @return Returns an {@link Optional} of the specified type if the instance is castable, if not it returns an empty
+     *         {@link Optional}.
+     * @throws NullPointerException Throws a {@link NullPointerException} if either {@code type} or {@code instance} is
+     *                              {@code null}.
+     * @see java.util.Optional
+     */
+    static <T> Optional<T> tryToSpecify(Class<T> type, Object instance) {
```
Saladoc: `T` should also extend `S`
```suggestion
    static <T extends S> Optional<T> tryToSpecify(Class<T> type, Object instance) {
```

Saladoc:
I think this should receive a Bump in Priority as we will probably make heavy use of this with audio sources.

PR #303, opened 2018-05-24
Replace slf4j by log4j2

Vampire:
I played around with replacing slf4j by log4j2.
Here is the result (properly mergeable).
As I'd have expected, most logging statements in Javacord are not affected, as they do not do calculation of log messages anyway.

But while I did this I found other things that are nicer with log4j2 than with slf4j, for example:

- The log4j2-api is capable of managing MDC (even if called differently) and its `SimpleLogger` can display it, that means you can have the shard of a log message
- No `JavacordLogger` is necessary anymore, we can simply use the `SimpleLogger` included in log4j2-api
- It is better OSGi compatible than slf4j
- The privacy protection logger is much easier and leaner to implement


Bastian:
> The log4j2-api is capable of managing MDC (even if called differently) and its SimpleLogger can display it, that means you can have the shard of a log message

So it's possible to display the shard in the log message?

This doesn't seem to be part of the PR at the moment:

Would you mind adding this, as it seems like a very useful feature.

Vampire:
Hm, I thought I did. I'll have another look later. It seems I actually also forgot to update the readme, which probably shouldn't mention `slf4j` anymore if you merge this.

Vampire:
Ah, by default the thread context map (MDC in log4j2 terminology) is implemented with `ThreadLocal`s.
I had a `log4j2.component.properties` file in my bot that set the property `log4j2.isThreadContextMapInheritable` to `true`. (This actually was to get the shards properly logged)
This causes the thread context map to be intherited by child threads.
That`s why I didn't notice it does not work as expected, as for me it did.
I added now a lowest priority property source for log4j-api that sets the property to `true` by default, but can easily set overridden to `false` by any consumer.
Now you should get the shard logged with the fallback logger.

I also updated the readme now with the correct information.

And while I was at that, I found that I used the deprecated `runtime` there, instead of `runtimeOnly`, so I also replaced that.
Even if you don't merge this one, this change should be done in the readme and also in the wiki articles. (sorry)

PR #304, opened 2018-05-25
Add methods to remove fields and set field properties in the EmbedBuilder

Vampire:


Vampire:
Updated the implementation to use predicates and consumers.
Much leaner and nicer now.

PR #309, opened 2018-06-06
Gradle Changes

Vampire:
- Do not publish the shaded JAR
- Switch to TeamCity as official CI in generated POMs
- Activate signing of release tags
- Update shadow plugin to 2.0.4
- Create draft GitHub release with all artifacts during release pipeline
- Upgrade to Gradle 4.8


gradle/pom.gradle
```diff
@@ -1,51 +1,55 @@
...
-                email 'Bjoern@Kautler.net'
-                url 'https://github.com/Vampire'
-                timezone 'Europe/Berlin'
+    publishing {
+        publications {
+            javacord {
+                pom {
+                    name = rootProject.name.capitalize() + (project.parent ? " ($shortName)" : '')
+                    description = project.description
+                    url = 'https://github.com/BtoBastian/Javacord/wiki'
```
BtoBastian: Should this really link to the wiki?
I guess it would be better to just link to the "normal" GitHub page (or better https://javacord.org/) as we won't use the GitHub wiki in the future, but move it to our homepage.

Vampire: Yeah, you're probably right, fixed

BtoBastian: Is there a reason, that you link to `http` instead of `https`?

Vampire: Well, I usually prefer to only use HTTPS if necessary.
Imho it is unneeded overhead to do the encryption if unencrypted transfer is also possible and no sensitive data is involved.
Actually right now as the URL forwards to an HTTPS page, I'd also use the HTTPS URL there, as some systems don't like HTTP-to-HTTPS redirects due to security.
But if I got you right, the URL will in the near future point to the documentation instead that is currently ported over from the Wiki and iirc those pages will not be HTTPS, but plain HTTP.
But if you prefer HTTPS, tell me, then I'll change it.

BtoBastian: The wiki will also use https. The performance impact of using https is negligible and besides that there are no downsides in using https. 

So I would be glad, if you could change the url to https. Besides that, the PR looks fine to me.

PR #311, opened 2018-06-25
Nameable Interface for Entities with Names

depressingillusion:


BtoBastian:
Why did you close this issue.
A `Nameable` interface is a good idea. :-)

Just two minor details:
- If an interface extends the new Nameable interface, you can also remove the `getName()` method from it
- As you probably noticed, some of your commits failed to build. That's no big deal, as you fixed it with your last commit. Just rebase it into a single commit or create a new commit if you are not familiar with git rebases.

Vampire:
He failed in doing the rebase and somehow crumbled his Git repo.
He didn't want to spend the time to fix it but redo it completely.
So we can probably expect another PR with the changes.

PR #312, opened 2018-07-07
Replace BtoBastian by Javacord in various places and update the readme a bit

Vampire:


Vampire:
Ok, I incorprated the feedback from different suggestions and discussions in the channel.
I hope I covered everything, was much chatter. :-D
How about current version (for immediate application, not for release, for that then the snapshot things get replaced by the final stuff).
What I basically changed from previous state:

- Only mention snapshot usage and JavaDoc for now
- swapped blocking an non-blocking login example
- removed the inner `exceptionally` calls
- Incorporated most of Aze changes
- I think the logging configuration should stay, e. g. for SLF4J warning in JDA even with the explicit link with the concrete information in the warning, the users keep asking about it instead of just doing what the warning tells them, so I think this bit is important to reduce questions eventually

What is missing:
- A current invite for #javacord@DAPI, the one in the readme is dead

Vampire:
Now also exchanged the shields by the own URL and removed the invite for the DAPI server

README.md
```diff
@@ -1,69 +1,93 @@
...
+If you never used Maven before you should take a look at the setup tutorial:
+* **[IntelliJ & Maven Setup](https://github.com/Javacord/Javacord/wiki/How-to-setup-(IntelliJ-and-Maven))**
+* **[Eclipse & Maven Setup](https://github.com/Javacord/Javacord/wiki/How-to-setup-(Eclipse-and-Maven))**
+
 ## Support
 
-* [Javacord server](https://discord.gg/0qJ2jjyneLEgG7y3) (recommended)
-* [DiscordAPI #java_javacord channel](https://discord.gg/0SBTUU1wZTVXVKEo)
+Javacord's Discord community is an excellent resource if you have questions about the library.  
+There is also the #java_javacord channel on the official DiscordAPI server.
```
Bastian: This line is obsolete now.

README.md
```diff
@@ -1,69 +1,93 @@
...
-A multithreaded but easy to use library to create a Discord bot in Java.
+# Javacord <a href="#"><img src="https://shields.javacord.org/badge/Version-3.0.0-brightgreen.svg?&style=flat-square" alt="Latest version"></a> <a href="https://ci.javacord.org/javadoc/"><img src="https://shields.javacord.org/badge/JavaDoc-latest-yellow.svg?style=flat-square" alt="Latest JavaDocs"></a> <a href="https://github.com/Javacord/Javacord/wiki"><img src="https://shields.javacord.org/badge/Wiki-Home-red.svg?style=flat-square" alt="Javacord Wiki"></a> <a href="https://discord.gg/0qJ2jjyneLEgG7y3"><img src="https://shields.javacord.org/discord/151037561152733184.svg?colorB=%237289DA&label=Discord&style=flat-square" alt="Discord Server"></a>
+An easy to use multithreaded library for creating Discord bots in Java.
```
Bastian: The new "slogan" should probably also be updated in the `build.gradle` to be consistent.

README.md
Vampire: I guess you just mean the second one? Removed.

README.md
Vampire: Totally correct of course, fixed.

Bastian:
Thanks :)

PR #316, opened 2018-07-08
Updated those URLs that I could find.

Kaleidox00:
Also updated the Example Maven Dependency in  README.md


Bastian:
Thanks, but Vampire was faster. ;)
https://github.com/Javacord/Javacord/pull/312

PR #321, opened 2018-07-10
Add Server#getRoleColor(User) and User#getRoleColor(Server)

BoneyIsSpooky:
From todo_general

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -93,6 +93,7 @@ import org.javacord.api.listener.user.UserChangeStatusListener;
 import org.javacord.api.listener.user.UserStartTypingListener;
 import org.javacord.api.util.event.ListenerManager;
 
+import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.InputStream;
@@ -101,6 +102,7 @@ import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -630,6 +632,19 @@ public interface Server extends DiscordEntity, UpdatableFromCache<Server> {
                 getRoles().stream()
                         .filter(role -> role.getUsers().contains(user))
                         .collect(Collectors.toList()));
+    }
+
+    /**
+     * Gets the displayed color of a user based on his roles on this server.
+     *
+     * @param user The user.
+     * @return The color.
+     */
+    default Optional<Color> getRoleColor(User user) {
+        return user.getRoles(this).stream()
+                   .filter(role -> role.getColor().isPresent())
+                   .max(Comparator.comparingInt(Role::getPosition))
+                   .flatMap(Role::getColor);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -520,6 +522,19 @@ default Role getEveryoneRole() {
...
+     *
+     * @param user The user.
+     * @return The color.
+     */
+    default Optional<Color> getRoleColorOf(User user) {
+        return user.getRoles(this).stream()
+                   .filter(role -> role.getColor().isPresent())
+                   .max(Comparator.comparingInt(Role::getPosition))
+                   .flatMap(Role::getColor);
+    }
```
Vampire: If you get a `List` you already have a defined order, in this case by position, so sorting is unnecessary.
This should be sufficient I think:
```java
        return user.getRoles(this).stream()
                   .filter(role -> role.getColor().isPresent())
                   .findFirst();
```

Vampire: It is also 4 lines, but it is faster, as it does not have to sort and thus does not have to go through all roles unnecessarily.

Vampire: Ah, forget what I said, the roles are sorted from smallest to biggest, so you would have to reverse the list first and for that you would have to create a new list first.

Vampire:
Iirc we usually also then add `User#getRoleColorOf(Server server)` that basically just calls the other for convenience.

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -244,6 +245,17 @@ default String getDiscriminatedName() {
...
     }
 
+    /**
+     * Gets the displayed color of the user based on his roles in a server.
+     *
+     * @param server The server.
+     * @return The color.
+     * @see Server#getRoleColorOf(User)
+     */
+    default Optional<Color> getRoleColor(Server server) {
```
Vampire: Besides that there is a PR from me that removes all "Of"s, is there a reason you name the method inconsistently?

BoneyIsSpooky: The first method was named as it was due to it being listed that way in the todo. I can change one of them, if that is needed.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -520,6 +522,19 @@ default Role getEveryoneRole() {
                         .collect(Collectors.toList()));
     }
 
+    /**
+     * Gets the displayed color of a user based on his roles in the server.
```
Vampire: to stay consistent with other docs I'd probably say `on the server`

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -244,6 +245,17 @@ default String getDiscriminatedName() {
         return server.getRolesOf(this);
     }
 
+    /**
+     * Gets the displayed color of the user based on his roles in a server.
```
Vampire: to stay consistent with other docs I'd probably say `on the given server`

BoneyIsSpooky: `in the given server` is most consistent with the methods around it.

Vampire: Hm, right, we need to improve consistency in the docs :-D

PR #322, opened 2018-07-10
Make EventDispatcher thread-safe and rework some of its synchronization code

Vampire:


Vampire:
Hm, something breaks currently :-/

Vampire:
Ok, fixed :-)

Bastian:
Is there any reason you removed/changed the padding from some JavaDocs?
The new padding is inconsistent with our code conventions (see [Contribution Guidelines](https://github.com/Javacord/Javacord/blob/v_3/.github/CONTRIBUTING.md#code-conventions)).

Vampire:
Just still not the correct formatting settings in IJ I'd say.
But actually I don't see which rule you see violated.
I only find https://google.github.io/styleguide/javaguide.html#s4.6.3-horizontal-alignment which basically says "If you want to do horizontal alignment, do it, but actually it would be better you don't do it."

How are your formatting settings for JavaDoc?
I can only get it to do 

```java
     * @param object    The object which is used to determine in which thread the consumer should be executed.
     *                  Usually the object is a server object (for server-dependent events), the discord api instance
     *                  (for server-independent events) or <code>null</code>. Providing a <code>null</code> parameter
     *                  means, that all previously consumers are going to run and then the new consumers. When
     *                  all null-parameter consumers were executed it begins to execute the other consumers like normal
     *                  again.
     * @param listeners A list with listeners which get consumed by the given consumer.
     * @param consumer  A consumer which consumes all listeners from the given list and is meant to call their
     *                  <code>onXyz(Event)</code> method.
     * @param <T>       The type of the listener.
```
or
```java
     * @param object The object which is used to determine in which thread the consumer should be executed.
     * Usually the object is a server object (for server-dependent events), the discord api instance
     * (for server-independent events) or <code>null</code>. Providing a <code>null</code> parameter
     * means, that all previously consumers are going to run and then the new consumers. When
     * all null-parameter consumers were executed it begins to execute the other consumers like normal
     * again.
     * @param listeners A list with listeners which get consumed by the given consumer.
     * @param consumer A consumer which consumes all listeners from the given list and is meant to call their
     * <code>onXyz(Event)</code> method.
     * @param <T> The type of the listener.
```
or
```
     * @param object The object which is used to determine in which thread the consumer should be executed.
     *         Usually the object is a server object (for server-dependent events), the discord api instance
     *         (for server-independent events) or <code>null</code>. Providing a <code>null</code> parameter
     *         means, that all previously consumers are going to run and then the new consumers. When
     *         all null-parameter consumers were executed it begins to execute the other consumers like normal
     *         again.
     * @param listeners A list with listeners which get consumed by the given consumer.
     * @param consumer A consumer which consumes all listeners from the given list and is meant to call their
     *         <code>onXyz(Event)</code> method.
     * @param <T> The type of the listener.
``` 
but not to only align within one parameter description.

Bastian:
The first one is the way to go.
I was talking about this JavaDoc example from the contribution guidelines:

I'm working on some clearer JavaDoc guidelines, but it's not ready yet.

The old event dispatcher breaks some of the guidlines anyway, like maximum line length.

Vampire:
Ah, I didn't really read that part as it is headlined with "this is in the Google styleguide anyway, I just mention it here becaue people ignore it".
At that line I switched to the Google styleguide instead.
So this should be moved to the "we do it differently than Google" part. :-)

Vampire:
Fixed

PR #324, opened 2018-07-12
[WIP] EmbedBuilder building an actual object

Kaleidox00:
dont judge yet, this is just to have a WIP PR open


Kaleidox00:
I think i cant do a proper way in the time i have left. Someone else should do this

PR #327, opened 2018-07-13
Read version from generated properties and add the commit id and build timestamp for a SNAPSHOT build

Vampire:


javacord-api/src/main/java/org/javacord/api/Javacord.java
```diff
@@ -1,14 +1,89 @@
...
-    public static final String VERSION = "3.0.0";
+    public static final String DISPLAY_VERSION;
+
+    static {
+        Properties versionProperties = new Properties();
+        try (InputStream versionPropertiesStream = Javacord.class.getResourceAsStream("/version.properties")) {
+            versionProperties.load(versionPropertiesStream);
+        } catch (IOException ignored) { }
+
+        String version = versionProperties.getProperty("version", "$version");
```
Bastian: Why so complicated instead of just doing 
```java
VERSION = versionProperties.getProperty("version", "<unknown>");
```
?

Bastian: Oh, I see why. Never mind

Vampire:
I added that the commit id and whether the worktree is clean are determined through Git if possible if the commit id is not supplied by CI

PR #329, opened 2018-07-14
Added method Message#toMessageBuilder

Kaleidox00:
From TODO

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -51,6 +51,17 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
             Pattern.compile("\\\\(?<char>[^a-zA-Z0-9\\p{javaWhitespace}\\xa0\\u2007\\u202E\\u202F])");
 
     /**
+     * Returns a {@code MessageBuilder} according to this {@code Message}.
+     *
+     * @see MessageBuilder#fromMessage(Message)
+     *
+     * @return The {@code MessageBuilder}.
+     */
+    default MessageBuilder toMessageBuilder() {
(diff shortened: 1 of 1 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -50,6 +50,15 @@
     Pattern ESCAPED_CHARACTER =
             Pattern.compile("\\\\(?<char>[^a-zA-Z0-9\\p{javaWhitespace}\\xa0\\u2007\\u202E\\u202F])");
 
+    /**
+     * Returns a {@link MessageBuilder} according to this {@link Message}.
+     *
+     * @return The {@link MessageBuilder}.
```
Vampire: I'd add an `@see` tag pointing to the `fromMessage` method

Kaleidox00: very agreed. maybe we should have this as a bigger PR; because im not sure if other methods have that?

Vampire: Our JavaDocs are generally not too consistent.
There are e. g. many "in server" and many "on server" and so on.
And the JavaDocs are partly very basic anyway.
Make a PR improving them if you feel like it :-)

Kaleidox00: are you challenging me to look through all the docs? lewl

Vampire: Not challenging, just mentioning it could be done. :-)

Kaleidox00: sounds like a challenge to me. consider it done in about a week (since i have a week off and nothing better to do lol)

Bastian:
👍

PR #330, opened 2018-07-14
Added missing documentation for role replacements for Message#getReadableContent()

Kaleidox00:
From TODO.


Bastian:
Thank you, but that's not 100% correct.
Role mentions start with `@` not `#`.

Kaleidox00:
Fixed

PR #331, opened 2018-07-14
Introduced the MessageActivity object.

Kaleidox00:
From TODO

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivity.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivity.java
@@ -22,7 +22,7 @@ public interface MessageActivity {
     Optional<String> getPartyId();
 
     /**
-     * Gets the {@link Message} of the activity.
+     * Gets the message of the activity.
      *
      * @return The message of the activity.
      */
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivity.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivity.java
@@ -1,0 +1,30 @@
+package org.javacord.api.entity.message;
+
+import java.util.Optional;
+
+/**
+ * This class represents a MessageActivity like Spotify group listening.
+ */
+public interface MessageActivity {
+
+    /**
+     * Gets the type of the activity.
+     *
+     * @return The type of the activity.
+     */
+    MessageActivityType getType();
+
+    /**
+     * Gets the party id of the activity.
+     *
+     * @return The party id of the activity.
+     */
+    Optional<String> getPartyId();
+
+    /**
+     * Gets the {@link Message} of the activity.
+     *
+     * @return The message of the activity.
+     */
+    Message getMessage();
+}
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivityType.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivityType.java
@@ -1,19 +1,22 @@
 package org.javacord.api.entity.message;
 
+/**
+ * Represents a message activity type.
+ */
 public enum MessageActivityType {
 
     /**
-     * A join invite activity.
+     * A join activity.
      */
     JOIN(1),
 
     /**
-     * A spectate invite activity.
+     * A spectate activity.
      */
     SPECTATE(2),
 
     /**
-     * A listen invite activity.
+     * A listen activity.
      */
     LISTEN(3),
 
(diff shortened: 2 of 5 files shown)


javacord-core/src/main/java/org/javacord/core/entity/message/MessageActivityImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import org.javacord.api.entity.message.Message;
+import org.javacord.api.entity.message.MessageActivity;
+import org.javacord.api.entity.message.MessageActivityType;
+
+import java.util.Optional;
+
+public class MessageActivityImpl implements MessageActivity {
+    private MessageActivityType type;
+    private String partyId = null;
+    private Message message;
```
Bastian: You never set the message field, it's always `null`

Kaleidox00: Fixed, my bad 👍

javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivityType.java
```diff
@@ -0,0 +1,69 @@
...
+            case 5:
+                return JOIN_REQUEST;
+            default:
+                return UNKNOWN;
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "MessageActivityType (" + type + ")";
```
Bastian: Why do overwrite the toString() method of an enum?
`MessageActivityType (3)` isn't really meaningful.

We also don't do this anywhere else in Javacord.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivityType.java
```diff
@@ -0,0 +1,69 @@
...
+        this.type = type;
+    }
+
+    /**
+     * Gets the type by its int representation.
+     *
+     * @param type The int representation.
+     * @return The message type.
+     */
+    public static MessageActivityType byType(int type) {
```
Bastian: We usually would call this method `getMessageActivityById(...)`.
Take a look at [Activity.java](https://github.com/Javacord/Javacord/blob/v_3/javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityType.java) as reference how an enum like this one should look like (this also includes stuff like renaming `type` to `id` etc, not only this single method name).

javacord-core/src/main/java/org/javacord/core/entity/message/MessageActivityImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public Message getMessage() {
+        return message;
+    }
+
+    @Override
+    public String toString() {
+        return "MessageActivityImpl (In message " + message.getId() + ")";
```
Bastian: That's not how Javacord's `toString()` methods are formatted.
Look at [Message.java](https://github.com/Javacord/Javacord/blob/v_3/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java#L449) as a reference.

It should look like this:
```java
return String.format("MessageActivity (type: %s, partyId: %s, message: %s)", getType(), getPartyId().orElese("none"), getMessage());
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivity.java
```diff
@@ -0,0 +1,29 @@
+package org.javacord.api.entity.message;
+
+import java.util.Optional;
+
+/**
+ * This class represents a MessageActivity like Spotify group listening.
+ */
+public interface MessageActivity {
+    /**
```
Bastian: This might be cherry picking, but I would prefer to have an empty line between the class/interface/enum definition and the first method/field/etc.

Of course not only here, but also in the other classes you added,

javacord-api/src/main/java/org/javacord/api/entity/message/MessageActivityType.java
```diff
@@ -0,0 +1,69 @@
...
+     * A join request activity.
+     */
+    JOIN_REQUEST(5),
+
+    /**
+     * An unknown activity.
+     */
+    UNKNOWN(-1);
+
+
```
Bastian: Cherry picking again, but why two empty lines?

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -260,6 +271,11 @@ public void removeAllReactionsFromCache() {
         reactions.clear();
     }
 
+    @Override
+    public Optional<MessageActivity> getActivity() {
+        return Optional.ofNullable(activity);
+    }
```
Bastian: The method should be at the same position like in the interface.
In this case between `getChannel()` and `isPinned()`

javacord-core/src/main/java/org/javacord/core/entity/message/MessageActivityImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import org.javacord.api.entity.message.Message;
+import org.javacord.api.entity.message.MessageActivity;
+import org.javacord.api.entity.message.MessageActivityType;
+
+import java.util.Optional;
+
+public class MessageActivityImpl implements MessageActivity {
+    private MessageActivityType type;
+    private String partyId = null;
+    private Message message;
```
Bastian: Whenever it makes sense, fields should be final.
As non of these fields every change, they should all be final.

Kaleidox00:
fixed

Bastian:
Awesome, thx. 👍 

I missed two point in my last review:
- Don't use `{@link Xyz}` where it's no necessary. E.g. "The {@link MessageActivityType} of the activity." should just be "The type of the activity." (also the case for `getMessage()`)
- Instead of "Returns abc", getter JavaDocs should say "Gets abc".

Otherwise everything looks good now. :)

Kaleidox00:
fixed again

Bastian:
Did you forget to push your fixes?

Kaleidox00:
oh no, i screwed rebasing up

PR #332, opened 2018-07-14
[WIP/Untestable] Introduced the MessageApplication object.

Kaleidox00:
This is just a commit to have something that can access those variables. So; maybe WIP?


Bastian:
I would like to wait until we actually know what this is and can properly test it.

Kaleidox00:
gave it a [WIP/Untestable] flag

PR #333, opened 2018-07-14
Added Methods: getBotUsers and getHumanUsers

Kaleidox00:
Thought I'd add these methods for the sake of coverage in terms of being able to get pre-filtered Collections of just Bot-Accounts or other way around.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -561,6 +561,20 @@ public interface DiscordApi {
     Collection<User> getCachedUsers();
 
     /**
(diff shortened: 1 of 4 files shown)


Bastian:
I don't think we need this. A user can easily just filter the member list using streams.
```java
server.getMembers().stream()
    .filter(User::isBot)
    .forEach(bot -> {
        // Do something
    });
```
If we add this, you could also argue for other "bloat" like `getMembersWithName(...)`, `getOnlineMembers()`, `getRolelessMembers()` and 10'000 others of the same kind.
There was the idea of having a Predicate as a possible argument to allow stuff like 
```java
server.getMemersWith(User::isBot);
```
but I've even decided against this more generic option.
Id and name are the only filters I consider important enough to have them as their own methods.

PR #334, opened 2018-07-15
Add the japan region

Tea-Ayataka:


Bastian:
Hey, thank you :)

Did you check if there's also a VIP region for Japan?
You can check it, by using the `GET /voice/regions` endpoint: https://discordapp.com/developers/docs/resources/voice#list-voice-regions

Tea-Ayataka:
I couldn't find any VIP server in that endpoint.

Bastian:
Ok, thank you :)

PR #336, opened 2018-07-18
Permissions on channel creation

Saladoc:
From the todo channel.

Major changes are

- Introduced the `Permissionable` interface for entities that can be granted permissions (currently `User` and `Role`)
- Reworked the server channel builders in order to reduce duplication of both code and documentation
- Included the possibility to set permission overrides on channel creation.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelBuilder.java
@@ -1,5 +1,6 @@
 package org.javacord.api.entity.channel;
 
+import org.javacord.api.entity.DiscordEntity;
 import org.javacord.api.entity.Permissionable;
 import org.javacord.api.entity.channel.internal.ServerChannelBuilderDelegate;
 import org.javacord.api.entity.permission.Permissions;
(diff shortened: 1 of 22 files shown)


javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * The server channel delegate used by this instance.
+     */
+    private final ServerChannelBuilderDelegate delegate;
+
+    /**
+     * Creates a new server channel builder without delegate.
+     */
+    protected ServerChannelBuilder() {
+        delegate = null;
```
Bastian: Wouldn't it make more sense to have the delegate as a parameter of the constructor instead of setting it "manually" in the child classes?
Just a minor detail.

Saladoc: I'm not sure about this one to be honest. I mimicked the `ServerChannelUpdater` implementation with this. Since the subclass will overwrite the variable with their delegate, I'm not entirely sure if it would work.

Saladoc:
I made a point to leave the methods for effective permissions with a `User` and have missed `#getOverwrittenPermissions`. Changed it now.

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelBuilder.java
Bastian: Let's just keep it like it is. :)

PR #337, opened 2018-07-18
Nameable interface

Saladoc:
Based on #311 by @depressingIllusion, original PR has been rebased and squashed into a single commit

Redundant declarations of `public String getName()` have been left where they updated the methods javadoc.


Saladoc:
Keeping the methods definitly was a conscious decision that I now reverted without ever disclosing the totally legitimate reasons behind it.

Bastian:
👍

PR #338, opened 2018-07-19
Make Mentionables implement Formattable.

Saladoc:
Default is just rendering as the mention tag.
However, on those entities where the mention can be certain to have a fixed length padding is supported.


Bastian:
As the discussion on the Discord server didn't have a final decision here my suggestion:
* Let every `DiscordEntity` implement `Formattable`
* Use the entity name (or in case of messages, the content) as default value
* Use the result of `toString()` for the alternate flag
* Make sure, that this PR doesn't break any `toString()` methods

This is just what I feel is the most reasonable solution. Alternative suggestion are welcome.

javacord-api/src/main/java/org/javacord/api/entity/Nameable.java
```diff
@@ -1,14 +1,38 @@
...
     String getName();
+
+    @Override
+    default void formatTo(Formatter formatter, int flags, int width, int precision) {
+        boolean alternate = (flags & FormattableFlags.ALTERNATE) != 0;
+        String s = alternate ? this.toString() : this.getName();
+        try {
+            FormatUtil.format(s, formatter, flags, width);
+        } catch (IOException e) {
+            ExceptionLogger.get().apply(e);
```
Bastian: That's not how the ExceptionLogger is meant to be used.
I can see why it's tempting to just create a default method in the api module, but to access the proper log4j2 logger, the implementation must be moved into the core module.

javacord-api/src/main/java/org/javacord/api/util/FormatUtil.java
```diff
@@ -0,0 +1,42 @@
+package org.javacord.api.util;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.FormattableFlags;
+import java.util.Formatter;
+
+public class FormatUtil {
```
Bastian: This class should be in the core module, too.

Saladoc: Problem: Both `Region` and `AuditLogChangeType` are in the api module, so either I leave the util in the api or I'll have to reimplement the method twice.

Vampire: I agree that the `FormatUtil` should not be in the API module and if, then at most in an `.internal.` package, as those are not meant to be used by the users of Javacord and are also not covered by backward compatibility guarantees.
That those enums are nameable is "bad" of course for this situation.
But as it is only about one method and one place that calls it, you could maybe just throw away the util and put the implementation right into the interface?

Saladoc: That would work too. However we still would have the logging problem as the log4j api is only present in the core package.
Of course I could just rethrow the exception wrapped in a RuntimeException and leave the problem to the user...

javacord-api/src/main/java/org/javacord/api/entity/Nameable.java
Vampire: As there are `Nameable` enums in the API part, it is hard to move the implementation into core as mentioned in the comment thread below.
So it might actually be ok to use the exception logger here.
I'd just use instead of `ExceptionLogger.get().apply(e);`
```suggestion
            ExceptionLogger.getConsumer().accept(e);
```
as it has the better semantic here to use the `Consumer` variant, not the `Function` variant as you don't use the return value.

Actually I also used this construct in `org.javacord.api.entity.channel.TextChannel#typeContinuouslyAfter(long, java.util.concurrent.TimeUnit, java.util.function.Consumer<java.lang.Throwable>)` and there you accepted it. I don't remember though why I didn't move the implementation to the core package. Maybe there was no `InternalTextChannel` interface yet and I didn't want to duplicate the code. Or I wanted the unwrap and enrich logic we provide in the exception logger.

Anyway, if we keep the code in API because of the two enums, I'd say it is ok to here use the exception logger, especially as it should most probably not be triggered anywy, but in the variant I mentioned above.

Saladoc:
- Changed `ExceptionLogger` usage according to Vampires suggestion
 - Inlined the method call, making the `FormatUtil` class obsolete. 

We can always re-extract it if needed.

PR #339, opened 2018-07-20
Fix Role#isHigherThan, Server#canKickUser and Server#canBanUser

Vampire:
There were some errors in role comparision, user-kickability and user-banability calculations.

# Role#isHigherThan

Before:
- Roles on different servers => true
- Roles have same position => true

After
- Roles on different servers => IllegalArgumentException
- Roles have same position => false

# Server#canKickUser and Server#canBanUser

Before:
- User to do the kick has generally right to kick / ban => false
- User to do the kick is server owner => still role order was considered

After:
- User to do the kick does not have generally right to kick / ban => false
- User to do the kick is server owner => true

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/permission/Role.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/permission/Role.java
@@ -32,7 +32,8 @@ import java.util.concurrent.CompletableFuture;
 /**
  * This class represents a Discord role, e.g. "moderator".
  */
-public interface Role extends DiscordEntity, Mentionable, Nameable, Permissionable, UpdatableFromCache<Role> {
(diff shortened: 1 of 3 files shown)


Bastian:
It would probably make sense to implement the Comparable interface.

Saladoc:
I'm not entirely sure. A `Comparator` yes, but `Comparable` implies that there is a natural ordering.

Vampire:
What would you define as natural order of two roles in Discord? I think it is a good idea, as the contract also allows exceptions to be thrown. 
Well, to be correct, I think it is a good idea if really the id defines order if position is equal which I still have to try out.

Bastian:
At least for channels it's the case that ids are used if the positions are the same.
I guess it's the same with roles.

Vampire:
Yeah, probably, just want to test our first. You know, Discord and assumptions that a sane mind would assume make sense are not always in accordance. :smile:

PR #340, opened 2018-07-21
Add Message#mentionsEveryone

Tea-Ayataka:
I really needed this

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
@@ -120,8 +120,10 @@ public class MessageUpdateHandler extends PacketHandler {
 
             MessageEditEvent editEvent = null;
             if (packet.has("edited_timestamp") && !packet.get("edited_timestamp").isNull()) {
-                message.ifPresent(msg ->
-                        msg.setLastEditTime(OffsetDateTime.parse(packet.get("edited_timestamp").asText()).toInstant()));
+                message.ifPresent(msg -> {
+                    msg.setLastEditTime(OffsetDateTime.parse(packet.get("edited_timestamp").asText()).toInstant());
+                    msg.setMentionsEveryone(packet.get("mention_everyone").asBoolean());
+                });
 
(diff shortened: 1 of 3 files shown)


Kaleidox00:
Proposal:
Not use this method but instead make the result of Message#getUserMentions contain some static "@everyone" Role, which has to be static, so it can be checked easily with #contains. I might make a PR example of what I mean.

Tea-Ayataka:
There is no way to check if the user had 'Mentions Everyone' permission or not. So i think this PR is not useless.

Kaleidox00:
I don't understand?

Tea-Ayataka:
Nevermind. I'll edit this PR because this method is still useful.

Vampire:
Please do not merge `v_3` into your PR branch, but rebase and force-push your PR branch instead.

Tea-Ayataka:
Is this all right?

Vampire:
Regarding merge commit, yes, thanks. I didn't review the content thoroughly though.

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
```diff
@@ -120,8 +120,10 @@ public void handle(JsonNode packet) {
 
             MessageEditEvent editEvent = null;
             if (packet.has("edited_timestamp") && !packet.get("edited_timestamp").isNull()) {
-                message.ifPresent(msg ->
-                        msg.setLastEditTime(OffsetDateTime.parse(packet.get("edited_timestamp").asText()).toInstant()));
+                message.ifPresent(msg -> {
+                    msg.setLastEditTime(OffsetDateTime.parse(packet.get("edited_timestamp").asText()).toInstant());
+                    msg.setMentionsEveryone(packet.get("mention_everyone").asBoolean());
```
Bastian: As far as I know this value never updates.

Bastian: Ok, I just tested it and it indeed updates.
My bad.

Bastian:
Thank you :)

PR #341, opened 2018-07-24
Added exceptions to Embed creation.

Kaleidox00:
Added a bunch of IllegalStateExceptions, which get thrown if specific boundaries for Embeds are broken.
Also, added those Boundaries as final static variables for the EmbedBuilderDelegateImpl.java


Bastian:
This is against Javacord's philosophy to not validate any user input but let Discord do the work.

Kaleidox00:
Well, I had in mind that the currently thrown exceptions do not allow easy tracking down of the cause for an invalid embed.

Bastian:
That's Discord's fault and will hopefully be improved in the next gateway version (v7).

PR #348, opened 2018-08-02
Prepare readme for final release

Bastian:
This prepares the README for the final release and contains some other small changes.


.github/README_template.md
```diff
@@ -21,40 +21,29 @@ New features introduced by Discord are typically added in less than one week, de
...
+    <artifactId>javacord</artifactId>
+    <version>$version</version>
+    <type>pom</type>
+</dependency>
 ```
 
 ### Optional Logger Dependency
 
-Any Log4j 2 API or SLF4j compatible logging framework can be used to provide a more sophisticated logging experience
+Any Log4j 2 API compatible logging framework can be used to provide a more sophisticated logging experience
```
Vampire: That was intentionally there. If you add `log4j-to-slf4j` artifact, you can use any SLF4J compatible logging framework like for example Logback, just the `j` should be uppercase.

.github/README_template.md
```diff
@@ -63,19 +52,11 @@ For example Log4j Core in Gradle
 dependencies { runtimeOnly 'org.apache.logging.log4j:log4j-core:2.11.0' }
 ```
 
-or Logback via SLF4J in Gradle
```
Vampire: That was intentionally there to show how to easily use any SLF4J compatible logger like Logback, especially as it seems to be pretty widespread even if imho log4j2 is better.

.github/README_template.md
```diff
@@ -63,19 +52,11 @@ For example Log4j Core in Gradle
...
-    runtimeOnly 'ch.qos.logback:logback-classic:1.2.3'
-}
-```
-
 ## IDE Setup
 
-If you never used Maven before you should take a look at the setup tutorial:
-* **[IntelliJ & Maven Setup](https://github.com/Javacord/Javacord/wiki/How-to-setup-(IntelliJ-and-Maven))**
-* **[Eclipse & Maven Setup](https://github.com/Javacord/Javacord/wiki/How-to-setup-(Eclipse-and-Maven))**
+If you never used Maven or Gradle before you should take a look at one of the setup tutorials:
```
Vampire: I intentionally didn't mention Gradle here.
This would imply either that the tutorials show how to use Gradle, or that if you neither used Maven nor Gradle before you should use Maven, which is a bad advise imho.
The only real reason to use Maven nowaday is because one is too used to it to try something new, but for new guys I think it is a bad recommendation.

.github/README_template.md
```diff
@@ -84,12 +65,11 @@ Javacord's Discord community is an excellent resource if you have questions abou
...
 ## Wiki
 
-For additional information, take a look at the [Javacord wiki](https://github.com/Javacord/Javacord/wiki).  
-The wiki for Javacord 3 is a work in progress and not complete!
+For additional information, take a look at the [Javacord wiki](https://javacord.org/wiki/getting-started/welcome/).  
 
 ## JavaDoc
 
-The JavaDoc for the latest snapshot build can be found on [TeamCity](https://javadoc.javacord.org).  
+The JavaDoc for the latest snapshot build can be found on [TeamCity](https://docs.javacord.org/api/v/latest/).
```
Vampire: `s/snapshot build/release/`
Do we want to want to mention TeamCity here? It is not visibly, it is practically only the filehoster.

.github/README_template.md
Bastian: My goal is to make the README as simple as possible.
I don't think we need examples for multiple logging frameworks/apis there.
This is something I would prefer to just have in the wiki and link to the corresponding article in the readme for everyone who is interested.

.github/README_template.md
Bastian: I would like to have setup-guides for Gradle, too. Maybe it would have been better to wait until these are available before mentioning Gradle.
The recommended setup if someone has zero experience will be IntelliJ + Gradle.

.github/README_template.md
Bastian: Maybe just write `The JavaDoc for the latest version can be found here` and make the `here` a link`?

Vampire: Would be one possibility, but actually I personally always find a linked `here` strange.
It says "here" but links to "there". (No, a linked "there" would not be better.)
I usually try to find a word or phrase that fits or if none can be found, have the actual URL also textual.

Bastian: Maybe `Full documentation of every class and method can be found in the [JavaDoc](link).` or similar?

Vampire: Maybe just `Developer documentation can be found in the [JavaDoc]`?
Or `API documentation can be found in the [JavaDoc]`?
Or something similar.
I think the "full" and the "every class and method" is superfluous for everyone that did Java before.
And if not, then a discord bot is probably not a good beginners project. :-D

Bastian: I like your first suggestion `Developer documentation can be found in the [JavaDoc]`

Vampire: Actually `... or in the [wiki]` :-D

Bastian: Done

Bastian:
Some new changes:
- Changed label of GitHub-release shield from `release` to `Version`
- Added link to the "Logger Configuration" wiki article
- Changed `or download it directly from GitHub` to `or download Javacord directly from GitHub`
- Some fixes, e.g. `The version number has the a 3-digit format` -> `The version number has a 3-digit format`

Vampire:
LGTM now except for 2 things.
- Still Maven is the suggested setup if user has no build tool knowledge :-D
- Section wiki, one sentence that links to wiki, section JavaDoc, one sentence that links to JavaDoc and wiki. Maybe combine it to one section "Developer documentation" and there have one sentence or two bullet points that send to JavaDocs and wiki respectively.

Bastian:
* We don't have setup guide for Gradle yet.
* Updated, both are now under `Documentation`

PR #366, opened 2018-08-06
Reworked ratelimit handling

### LINKED ISSUE #361: The 50/1 global REST ratelimit is not handled properly
At the moment, hitting the global 50/1 ratelimit (e.g. by trying to send 100 DMs to 100 different users at the same time) is not handled properly by Javacord.

Bastian:
This is a complete rework on how ratelimits are handled.
It's way easier to read and fixed #361 as good as possible without sacrificing parallel execution or hard-coding the global ratelimit.
As the global ratelimit information is stored in a `static` map with the token as key, it works across multiple shards, as long as they are within the same runtime.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java
@@ -3,20 +3,26 @@ package org.javacord.core.util.ratelimit;
 import org.javacord.api.DiscordApi;
 import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.util.rest.RestEndpoint;
+import org.javacord.core.util.rest.RestRequest;
 
-import java.util.Optional;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 public class RatelimitBucket {
 
+    // The key is the token, as global ratelimits are shared across the same account.
+    private static final Map<String, Long> globalRatelimitResetTimestamp = new ConcurrentHashMap<>();
+
     private final DiscordApiImpl api;
+
+    private final ConcurrentLinkedQueue<RestRequest<?>> requestQueue = new ConcurrentLinkedQueue<>();
 
     private final RestEndpoint endpoint;
     private final String majorUrlParameter;
 
-    private volatile long rateLimitResetTimestamp = 0;
-    private volatile int rateLimitRemaining = 1;
-
-    private volatile boolean hasActiveScheduler = false;
+    private volatile long ratelimitResetTimestamp = 0;
+    private volatile int ratelimitRemaining = 1;
 
     /**
      * Creates a RatelimitBucket for the given endpoint / parameter combination.
@@ -45,58 +51,58 @@ public class RatelimitBucket {
     }
 
     /**
-     * Gets the rest endpoint of the bucket.
+     * Sets a global ratelimit.
      *
-     * @return The endpoint of the bucket. If it's a global limit, the endpoint will be not be present.
+     * @param api A discord api instance.
+     * @param resetTimestamp The reset timestamp of the global ratelimit.
      */
-    public Optional<RestEndpoint> getEndpoint() {
-        return Optional.ofNullable(endpoint);
+    public static void setGlobalRatelimitResetTimestamp(DiscordApi api, long resetTimestamp) {
+        globalRatelimitResetTimestamp.put(api.getToken(), resetTimestamp);
     }
 
     /**
-     * Checks if this bucket has an active scheduler.
+     * Adds the given request to the bucket's queue.
      *
-     * @return Whether this bucket has an active scheduler or not.
+     * @param request The request to add.
      */
-    public synchronized boolean hasActiveScheduler() {
-        return hasActiveScheduler;
+    public void addRequestToQueue(RestRequest<?> request) {
+        requestQueue.add(request);
     }
 
     /**
-     * Sets if this bucket has an active scheduler.
+     * Polls a request from the bucket's queue.
      *
-     * @param hasActiveScheduler Whether this bucket has an active scheduler or not.
+     * @return The polled request.
      */
-    public synchronized void setHasActiveScheduler(boolean hasActiveScheduler) {
-        this.hasActiveScheduler = hasActiveScheduler;
+    public RestRequest<?> pollRequestFromQueue() {
+        return requestQueue.poll();
     }
 
     /**
-     * Checks if there is still "space" in this bucket, which means that you can still send requests without being
-     * ratelimited.
+     * Peeks a request from the bucket's queue.
      *
-     * @return Whether you can send requests without being ratelimited or not.
+     * @return The peeked request.
      */
-    public boolean hasSpace() {
-        return rateLimitRemaining > 0 || getTimeTillSpaceGetsAvailable() <= 0;
+    public RestRequest<?> peekRequestFromQueue() {
+        return requestQueue.peek();
     }
 
     /**
      * Sets the remaining requests till ratelimit.
      *
-     * @param rateLimitRemaining The remaining requests till ratelimit.
+     * @param ratelimitRemaining The remaining requests till ratelimit.
      */
-    public void setRateLimitRemaining(int rateLimitRemaining) {
-        this.rateLimitRemaining = rateLimitRemaining;
+    public void setRatelimitRemaining(int ratelimitRemaining) {
+        this.ratelimitRemaining = ratelimitRemaining;
     }
 
     /**
      * Sets the ratelimit reset timestamp.
      *
-     * @param rateLimitResetTimestamp The rateLimit reset timestamp.
+     * @param ratelimitResetTimestamp The ratelimit reset timestamp.
      */
-    public void setRateLimitResetTimestamp(long rateLimitResetTimestamp) {
-        this.rateLimitResetTimestamp = rateLimitResetTimestamp;
+    public void setRatelimitResetTimestamp(long ratelimitResetTimestamp) {
+        this.ratelimitResetTimestamp = ratelimitResetTimestamp;
     }
 
     /**
@@ -105,11 +111,13 @@ public class RatelimitBucket {
      * @return The time in seconds how long you have to wait till there's space in the bucket again.
      */
     public int getTimeTillSpaceGetsAvailable() {
-        if (rateLimitRemaining > 0) {
+        long globalRatelimitResetTimestamp =
+                RatelimitBucket.globalRatelimitResetTimestamp.getOrDefault(api.getToken(), 0L);
+        long timestamp = System.currentTimeMillis() + (api.getTimeOffset() == null ? 0 : api.getTimeOffset());
+        if (ratelimitRemaining > 0 && (globalRatelimitResetTimestamp - timestamp) <= 0) {
             return 0;
         }
-        long timestamp = System.currentTimeMillis() + (api.getTimeOffset() == null ? 0 : api.getTimeOffset());
-        return (int) (rateLimitResetTimestamp - timestamp);
+        return (int) (Math.max(ratelimitResetTimestamp, globalRatelimitResetTimestamp) - timestamp);
     }
 
     /**
--- a/javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
@@ -2,12 +2,8 @@ package org.javacord.core.util.ratelimit;
 
 import okhttp3.Response;
 import org.apache.logging.log4j.Logger;
-import org.javacord.api.DiscordApi;
 import org.javacord.api.exception.DiscordException;
-import org.javacord.api.exception.RatelimitException;
 import org.javacord.core.DiscordApiImpl;
-import org.javacord.core.util.Cleanupable;
-import org.javacord.core.util.concurrent.ThreadFactory;
 import org.javacord.core.util.logging.LoggerUtil;
 import org.javacord.core.util.rest.RestRequest;
 import org.javacord.core.util.rest.RestRequestResponseInformationImpl;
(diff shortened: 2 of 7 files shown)


Bastian:
This does not properly handle ratelimits in case of exceptions (e.g. sending 6 messages with more than 2000 characters in a row will cause a 429).

I'll fix this issue tomorrow.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
+                    Response response = result.getResponse();
+                    boolean global = response.header("X-RateLimit-Global", "false").equalsIgnoreCase("true");
+                    int remaining = Integer.valueOf(response.header("X-RateLimit-Remaining", "1"));
+                    long reset = request
+                            .getEndpoint()
+                            .getHardcodedRatelimit()
+                            .map(ratelimit -> currentTime + api.getTimeOffset() + ratelimit)
+                            .orElseGet(() -> Long.parseLong(response.header("X-RateLimit-Reset", "0")) * 1000);
+
+                    // Check if the received a 429 response
```
Vampire: Make proper English?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
+                    if (result.getResponse().code() == 429) {
+                        int retryAfter =
+                                result.getJsonBody().isNull() ? 0 : result.getJsonBody().get("retry_after").asInt();
+                        bucket.setRateLimitRemaining(0);
+                        bucket.setRateLimitResetTimestamp(currentTime + retryAfter);
```
Vampire: Move this into the `else` (`!global`)?
Doesn't make sense to set both for a global limit, does it?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
+
+                        // Update bucket information
+                        bucket.setRateLimitRemaining(remaining);
+                        bucket.setRateLimitResetTimestamp(reset);
+                    }
+
+                } catch (Throwable t) {
+                    if (request.getResult().isDone()) {
+                        logger.warn("Received exception for a request who is already done. "
+                                + "This should not be able to happen!", t);
```
Vampire: I think to make it proper English
`s/who/which/`
`s/be able to//`
@WalkerKnapp?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java
```diff
@@ -45,40 +54,49 @@ public RatelimitBucket(DiscordApi api, RestEndpoint endpoint, String majorUrlPar
     }
```
Vampire: `hasActiveScheduler` is still present but not used anymore

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
+                    // Execute the request
+                    result = currentRequest.executeBlocking();
+
+                    // Calculate the time offset, if it wasn't done before
+                    long currentTime = System.currentTimeMillis();
+                    calculateOffset(currentTime, result);
+
+                    // The headers of the response
+                    Response response = result.getResponse();
+                    boolean global = response.header("X-RateLimit-Global", "false").equalsIgnoreCase("true");
```
Vampire: Should the ratelimit stuff maybe all be done in the `finally` block, either with the result set a few lines above, or the result that was in the exception if any?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
+                    }
+
+                    // Poll a new quest
+                    synchronized (bucket) {
+                        bucket.pollRequestFromQueue();
+                        currentRequest = bucket.peekRequestFromQueue();
+                        if (currentRequest == null) {
+                            // Remove the bucket and leave the loop if there are no more requests in the bucket's queue
+                            buckets.remove(bucket);
+                            break;
```
Vampire: Maybe better remove this line and have `while (currentRequest != null)` instead of `while (true)`.
You have a clear abortion criterion, so use it in the loop, instead of a `break`.

While I think about it, you also introduce a race conditions here.
You only start the queue processing if there is no request in the bucket.
You abort the queue processing if there is no more request in the bucket.
This can happen e. g. with thread A (enqueuer) and B (queue worker):
- A gets `true` for `alreadyInQueue = bucket.peekRequestFromQueue() != null;`
- B finishes the request and does `bucket.pollRequestFromQueue();`
- B gets `null` for `currentRequest = bucket.peekRequestFromQueue();` and quits the processing
- A calls `bucket.addRequestToQueue(request);`
- Now the bucket is dead forever until bot restart

Too in the middle of the night to properly think about synchronization issues, but maybe putting
```java
alreadyInQueue = bucket.peekRequestFromQueue() != null;
buckets.add(bucket);
bucket.addRequestToQueue(request);
```
into a `synchronized (bucket) { }` could solve the issue properly.

Bastian: My bad.
The 
```java
// Poll a new quest
synchronized (bucket) {
```
should have been
```java
// Poll a new quest
synchronized (buckets) {
```

Vampire: Ah, ok, that could help too. :-)

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
 
-                        if (api.getTimeOffset() == null) {
-                            calculateOffset(currentTime, result);
-                        }
+                // Sleep until space is available
+                while (sleepTime > 0) {
+                    try {
+                        Thread.sleep(sleepTime);
+                    } catch (InterruptedException e) {
+                        logger.warn("We got interrupted while waiting for a rate limit!", e);
```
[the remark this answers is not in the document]
Bastian: It was a warn before. I don't know any reason why this thread should get interrupted?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -226,19 +166,19 @@ public void queueRequest(RestRequest<?> request) {
...
     private void calculateOffset(long currentTime, RestRequestResult result) {
-        // Discord sends the date in their header in the format RFC_1123_DATE_TIME
-        // We use this header to calculate a possible offset between our local time and the discord time
-        String date = result.getResponse().header("Date");
-        if (date != null) {
-            long discordTimestamp = OffsetDateTime.parse(date, DateTimeFormatter.RFC_1123_DATE_TIME)
-                    .toInstant().toEpochMilli();
-            api.setTimeOffset((discordTimestamp - currentTime));
-            logger.debug("Calculated an offset of {} to the Discord time.", api::getTimeOffset);
+        synchronized (api) {
```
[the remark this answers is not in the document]
Bastian: Well, this synchronizing was mainly there to prevent duplicate log messages. This can also get called when we hit a 429 and recalculate the offset.
Double-checking sounds like a good idea.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,161 @@
...
-            1, new ThreadFactory("Javacord - RatelimitBucket Delay Scheduler - %d", true));
-
-    private final Set<RatelimitBucket> buckets = ConcurrentHashMap.newKeySet();
-    private final HashMap<RatelimitBucket, ConcurrentLinkedQueue<RestRequest<?>>> queues = new HashMap<>();
-
+    // The discord api instance for this ratelimit manager
     private final DiscordApiImpl api;
 
+    // A set with all buckets which are currently in the queue
+    private final Set<RatelimitBucket> buckets = new HashSet<>();
```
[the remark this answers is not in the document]
Bastian: See my answer to your comment about `buckets.remove(bucket);`.
It was just a typo and `synchronized(bucket)` should have been `synchronized(buckets)`

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Bastian: About `while (currentRequest != null)`: I had it like this before, and changed it for a reason I don't remember. 🙈

Vampire: Hm, remember it and document it, or let me show you a better fix. :-D

Vampire: You could change it back until it fails, maybe the original reason is already fixed by some other change *g*

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Bastian: At the moment I'm trying to find the reason why the old code worked...
`executeBlocking()` throws an exception if the request failed (e.g. a too long message) and doesn't give access to the response headers then. 
I tested the old code and it ratelimits exceptions fine and I have no clue why.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Vampire: Maybe that the `currentRequest` setting and evaluating in this case is also guarded by the synchronized block?
Wouldn't make sense though, as it is a local variable anyway.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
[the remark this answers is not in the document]
Vampire: Yeah, probably not, but then maybe `throw new AssertionError("We got interrupted while waiting for a rate limit!", e);` would be better to express that we expect this to never happen and immediately show if our assumption is wrong.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Vampire: doesn't give access to the response headers?
well, the exception has the information, doesn't it?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java
```diff
@@ -45,40 +54,49 @@ public RatelimitBucket(DiscordApi api, RestEndpoint endpoint, String majorUrlPar
...
-     * Gets the rest endpoint of the bucket.
+     * Sets a global ratelimit.
      *
-     * @return The endpoint of the bucket. If it's a global limit, the endpoint will be not be present.
+     * @param api A discord api instance.
+     * @param resetTimestamp The reset timestamp of the global ratelimit.
      */
-    public Optional<RestEndpoint> getEndpoint() {
-        return Optional.ofNullable(endpoint);
+    public static void setGlobalRatelimit(DiscordApi api, long resetTimestamp) {
```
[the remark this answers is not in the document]
Bastian: No. The result is the same, so I just had to choose one and decided to go with `api` because it makes the call shorter.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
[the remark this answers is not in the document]
Vampire: But be carefull to do it right, the double-checked locking idiom can too easily be done wrongly.
Iirc https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java has a good description of how to do it wrongly and why and how to do it correctly. :-)

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Bastian: https://github.com/Javacord/Javacord/blob/master/javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java#L174
Thanks ^^
I completely missed this.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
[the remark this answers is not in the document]
Vampire: Yeah, forgot the recalculation on 429 when I wrote this.
I've seen it later and wanted to amend my comment here, but forgot it. :-(

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Vampire: Ah, yeah, I didn't actually look at the old code much, but that's what I meant. :-D

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,205 @@
...
-                        RestRequestResult result = restRequest.executeBlocking();
-                        rateLimitHeadersSource = result;
 
-                        long currentTime = System.currentTimeMillis();
+                    // Execute the request
+                    result = currentRequest.executeBlocking();
+
+                    // Calculate the time offset, if it wasn't done before
+                    responseTimestamp = System.currentTimeMillis();
+                    calculateOffset(responseTimestamp, result);
```
Vampire: Should the offset be recalculated in the `DiscordException` case too?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,205 @@
...
+
+                // Update the bucket information
+                bucket.setRateLimitRemaining(0);
+                bucket.setRateLimitResetTimestamp(responseTimestamp + retryAfter);
+            }
+        } else {
+            // Check if we didn't already complete it exceptionally.
+            CompletableFuture<RestRequestResult> requestResult = request.getResult();
+            if (!requestResult.isDone()) {
+                requestResult.complete(result);
```
Vampire: Just a detail, but isn't it a bit strange that the `updateRatelimitInformation` method actually completes the future now? Maybe instead return a `boolean` that signals whether we got ratelimited and complete the request outside if not?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java
```diff
@@ -45,40 +52,59 @@ public RatelimitBucket(DiscordApi api, RestEndpoint endpoint, String majorUrlPar
...
-     * Sets if this bucket has an active scheduler.
+     * Gets the bucket's current queue.
+     * This queue is the actual internal used reference and thus can be modified.
      *
-     * @param hasActiveScheduler Whether this bucket has an active scheduler or not.
+     * @return The buckets queue.
      */
-    public synchronized void setHasActiveScheduler(boolean hasActiveScheduler) {
-        this.hasActiveScheduler = hasActiveScheduler;
+    public ConcurrentLinkedQueue<RestRequest<?>> getRequestQueue() {
```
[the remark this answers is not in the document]
Bastian: Yes

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitBucket.java
```diff
@@ -3,21 +3,28 @@
...
 
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 public class RatelimitBucket {
 
+    // The key is the token, as global ratelimits are shared across the same account.
+    private static final Map<String, Long> globalRatelimit = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
Bastian: Agreed

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,205 @@
...
-                            bucket.setRateLimitRemaining(0);
-                            bucket.setRateLimitResetTimestamp(currentTime + retryAfter);
-                        } else {
-                            restRequestResult.complete(result);
+                    // Poll a new quest
+                    synchronized (buckets) {
+                        bucket.pollRequestFromQueue();
+                        currentRequest = bucket.peekRequestFromQueue();
+                        if (currentRequest == null) {
+                            // Remove the bucket and leave the loop if there are no more requests in the bucket's queue
```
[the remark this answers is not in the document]
Bastian: I think I now remember the reason for `while (true) {`
Without the `break;` inside the synchronization, the following can happen:
- A calls `buckets.remove(bucket);` at the end of the loop
- A leaves the synchronization block
- B checks if there are buckets in the queue (`alreadyInQueue = bucket.peekRequestFromQueue() != null;`) -> False
- A jumps to the `while (currentRequest != null)` check -> `true`
- B also starts a second thread

=> We have two different Threads handling the same bucket.

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Bastian: Yes

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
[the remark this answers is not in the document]
Vampire: We wont, will we?
`currentRequest?` is a local variable.
So if we are at `buckets.remove(bucket);`, `currentRequest` is `null` and will stay so.
Then the `while` condition is checked and the thread ends.
If between these another thread starts processing the next bucket, this would be ok.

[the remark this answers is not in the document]
Bastian: Hmm, you're right.
Dunno

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
```diff
@@ -2,221 +2,205 @@
...
+                logger.warn("Hit a global ratelimit! This means you were sending a very large "
+                        + "amount within a very short time frame.");
+                RatelimitBucket.setGlobalRatelimit(api, responseTimestamp + retryAfter);
+            } else {
+                logger.debug("Received a 429 response from Discord! Recalculating time offset...");
+                // Setting the offset to null causes a recalculate for the next request
+                api.setTimeOffset(null);
+
+                // Update the bucket information
+                bucket.setRateLimitRemaining(0);
```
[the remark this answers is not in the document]
Bastian: The issue is not, that it does not get set to `0`, but in case of a global ratelimit, the remaining ratelimits must not be considered. That's currently wrong in `getTimeTillSpaceGetsAvailable()`.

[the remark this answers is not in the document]
Vampire: Is it?
It currently waits for what is farther in the future, either the individual or the global, so it should be safe, shouldn't it?

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
Bastian: Yes, it's a little bit weird,
I just renamed the method to `handleResponse()` though

Bastian:
Done

javacord-core/src/main/java/org/javacord/core/util/ratelimit/RatelimitManager.java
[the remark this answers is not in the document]
Vampire: Oh, with the latest changes I see it, you're right of course. :-)

PR #367, opened 2018-08-07
Some smaller tweaks to the readme

Bastian:
* Fixed some spelling mistakes (most are taken from https://github.com/Javacord/javacord.github.io/pull/2)
* The "Version" badge now links to the latest GitHub release
* Replaced the deprecated `compile` with `implementation`
* Updated `log4j-core` dependency
* Updated link and name to the example bot
* To be consistent, most links are now in upper case. Before it was inconsistent (e.g. `Logger Configuration` but `Javacord wiki`)


.github/README_template.md
```diff
@@ -43,30 +43,30 @@ dependencies { compile 'org.javacord:javacord:$version' }
 
 ### Optional Logger Dependency
 
-Any Log4j 2 API compatible logging framework can be used to provide a more sophisticated logging experience
+Any Log4j-2-compatible logging framework can be used to provide a more sophisticated logging experience
 with being able to configure log format, log targets (console, file, database, Discord direct message, ...),
 log levels per class, and much more.
 
-For example Log4j Core in Gradle
+For example, Log4j Core in Gradle
```
Vampire: Are you sure about this comma?
I always do commas wrongly, even in German, but this looks really strange to me.

.github/README_template.md
```diff
@@ -43,30 +43,30 @@ dependencies { compile 'org.javacord:javacord:$version' }
...
 ```groovy
-dependencies { runtimeOnly 'org.apache.logging.log4j:log4j-core:2.11.0' }
+dependencies { runtimeOnly 'org.apache.logging.log4j:log4j-core:2.11.1' }
```
Vampire: I'm not sure whether we should use the one we package the API for or the latest here.
Actually we could also generate this from the template to be the version we also use as dependency, what do you think?

.github/README_template.md
Bastian: See https://github.com/Javacord/javacord.github.io/pull/2/files#diff-538eb5d9a549263cf8095185bdfa11a9R44

.github/README_template.md
Bastian: Sounds like a good idea 👍

.github/README_template.md
Vampire: Hm, ok, if the commi says it, it is probably fine :-)

PR #369, opened 2018-08-09
Code Generation Improvements

Vampire:
- Make code-generation tasks cacheable, even across worktrees
- Support objects to attach listeners to everyhwere in code generation instead only entities


Vampire:
As MysteryPotatoGuy asked what the thread selector is, I added a commit that
- renames `threadSelector` to `queueSelector`, as the thread is arbitrary and the queue is what is selected
- updates the JavaDoc of the `queueSelector` in `EventDispatcherBase`
- adds the JavaDoc to the `queueSelector` in `EventDispatcher`
- adds a `DispatchQueueSelector` interface with which valid queue selectors are now explicitly marked as such to prevent being given something else

Issue #364, opened 2018-08-06
Get the creator of a custom emoji


Bastian:
Add a method to receive the creator of a custom emoji.
As this information is only available using the `GET/guilds/{guild.id}/emojis/{emoji.id}` endpoint, we will need a method like this in `KnownCustomEmoji`:
```java
public CompletableFuture<User> requestCreator();
```
As an emoji's creator cannot change (but the user can leave the server!), we can cache the user id and use this id for subsequent requests as long as we have the user with this id in the cache.

PR #371, opened 2018-08-12
Added Method KnownCustomEmoji#requestCreator

ghost:
Tested.

This closes Issue https://github.com/Javacord/Javacord/issues/364.

-> Had to make the CompletableFuture contain an Optional of a user, because the [Emoji Documentation](https://discordapp.com/developers/docs/resources/emoji) does not ensure the user Object within the emoji object.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/emoji/KnownCustomEmoji.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/emoji/KnownCustomEmoji.java
@@ -75,7 +75,7 @@ public interface KnownCustomEmoji extends CustomEmoji, UpdatableFromCache<KnownC
      *
      * @return A completable Future to contain the creator of the KnownCustomEmoji.
      */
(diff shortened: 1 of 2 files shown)


Kaleidox00:
Changed the method so the method completes exceptionally when there was no creator found on the request.

WalkerKnapp:
I don't agree with this change to be honest. It's not that intuitive of behavior for it to throw an exception when there is no creator. I like how it was originally.

Bastian:
Thank you for the PR.

I would like to first solve #377 before merging.
I also have to agree with @WalkerKnapp: It should be `CompletableFuture<Optional<User>>` not `CompletableFuture<User>` with an exception.

burdoto:
As stated in #347 I cannot continue working on this flawlessly, so I will eventually open a new PR with the changes made.

burdoto:
[Closed from my End; won't fix]

Issue #370, opened 2018-08-11
The EmbedBuilder is missing a method to set a video


Bastian:
It's not possible to set the video with the embed builder.

Bastian:
I got told, that it's not possible to manually set embed videos.
So what's happening here: You upload the video and the field in the embed just gets ignored. So it's basically just an uploaded video and an empty embed. Thank you for testing it!

Issue #358, opened 2018-08-06
Let the EmbedBuilder actually build an embed


Bastian:
It's pretty awkward that you use the `EmbedBuilder` class itself to send embeds.The `EmbedBuilder` should get a `build()` method that builds an embed. For this change we need a new embed class, as such an embed cannot have methods like `EmbedImage#getProxyUrl()` or `EmbedImage#getWidth()`.
I'm thinking about a hierarchy like this:
![](https://i.imgur.com/z3zq9t5.png)
* The `BaseEmbed` contains the basic embed information
* The `SentEmbed` contains methods which can only exist for sent embeds like to get the proxy url
* The `EmbedDraft` contains methods which are only available to an unset embed, e.g. methods to receive files which should be attached to the message. It can also contain methods to directly edit the embed. I'm unsure about this classes' name. Maybe `UnsetEmbed` would be better?

Of course we need this hierarchy for all "children" as well, e.g. `SentEmbedImage` etc.

The `MessageBuilder#setEmbed​(EmbedBuilder)` and related methods should become deprecated and replaced by `MessageBuilder#setEmbed​(BaseEmbed)`.

Bastian:
Renaming the `Embed` class to `SentEmbed` would be a breaking change.

Bastian:
It's an interesting idea, but I don't know if we should implement it.
The issue I have with it are multiple edits.
E.g. if you want to edit the title and the description, it would first edit the title, send a request to Discord, then edit the description and send a second request to Discord.

RealYusufIsmail:
Any chance I can work this?

KILLEliteMaste:
I think we postpone this until there's an actual need for this. There was a small discussion in our Discord where someone mentioned that it might be appropriate that the EmbedBuilder does not build and embed. For context see [here](https://discord.com/channels/151037561152733184/354648262453624833/839193619813892106)

RealYusufIsmail:
ok

PR #374, opened 2018-08-13
[WIP] Reworked Embed structure

Kaleidox00:
This PR works on two issues:
1) https://github.com/Javacord/Javacord/issues/358
2) https://github.com/Javacord/Javacord/issues/370


Kaleidox00:
https://github.com/Javacord/Javacord/issues/370#issuecomment-412652892

Video field suppoort has been marked deprecated.

Vampire:
Mergeable?
You add new stuff that is right-away deprecated.
I don't think or hope this will ever get merged in that shape.
I'm also don't think the existing video fields should be deprecated.
Just that you don't yet can send a video does not mean you cannot receive a video.

Kaleidox00:
What would you do? Remove them? Keep them without real functionality?

Vampire:
What? The messages you newly add as deprecated?
Of course remove them.
Why should you introduce something new that is deprecated already.
Deprecation is to keep backwards compatibility for some time before you delete it.
And the video fields for received embed are filled e. g. if you get an embed with a youtube video.
It does not make any sense to deprecated those.

Kaleidox00:
There we go, getting Videos only, but not setting Videos for the reason that Bastian stated in #370.

Bastian:
Thank you for the PR, but there is a reason I self-asigned #358 to me. I already started working on it.

The main reason I moved all issues to GitHub is, that we don't have multiple people working on the same issue as you can easily assign an issue to you (or comment, if you don't have permissions to self-asign).

Kaleidox00:
well its not like this isnt finished and working already ¯\_(ツ)_/¯

burdoto:
Due to the fact of me having locked myself out of my account Kaleidox00, I can't continue working on any of my PRs, or I would have to reopen all PRs.

Vampire:
If you cannot solve it by standard password reset per mail procedure, drop a mail to the GitHub support, it is usually quite responsive and helpful in my experience.

burdoto:
I know. Already did that. Thing is I lost both the linked Authenticator and the Recovery codes, so there is nothing they can do.

burdoto:
[Closed from my End]

PR #381, opened 2018-09-09
Add missing overrides in ChannelCategoryBuilder

Saladoc:
Needed permissions-related methods for proper chaining.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelCategoryBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelCategoryBuilder.java
@@ -1,6 +1,9 @@
 package org.javacord.api.entity.channel;
 
+import org.javacord.api.entity.DiscordEntity;
+import org.javacord.api.entity.Permissionable;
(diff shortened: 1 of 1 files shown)


javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelCategoryBuilder.java
```diff
@@ -37,6 +40,19 @@ public ChannelCategoryBuilder setName(String name) {
         return this;
     }
 
+    @Override
+    public <T extends Permissionable & DiscordEntity> ServerChannelBuilder addPermissionOverwrite(
```
Vampire: Shouldn't this return a `ChannelCategoryBuilder`?

Saladoc: That would require me to amend my commit *before* pushing. Imagine if I was paying attention...

Vampire: I don't understand what you mean

PR #388, opened 2018-10-17
Update jacoco tool to jdk 11 compatible version.

Saladoc:


gradle/tests.gradle
```diff
@@ -10,6 +10,10 @@ subprojects {
         testRuntimeOnly 'org.objenesis:objenesis'
     }
 
+    jacoco {
+        toolVersion = "0.8.2"
```
Vampire:
```suggestion
        toolVersion = '0.8.2'
```

PR #389, opened 2018-10-17
Allow requesting an emojis creator.

Saladoc:
As seen on ~~TV~~ #364 and #371

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/emoji/KnownCustomEmoji.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/emoji/KnownCustomEmoji.java
@@ -3,6 +3,7 @@ package org.javacord.api.entity.emoji;
 import org.javacord.api.entity.UpdatableFromCache;
 import org.javacord.api.entity.permission.Role;
 import org.javacord.api.entity.server.Server;
+import org.javacord.api.entity.user.User;
 import org.javacord.api.listener.server.emoji.KnownCustomEmojiAttachableListenerManager;
 
 import java.util.Collection;
@@ -68,6 +69,13 @@ public interface KnownCustomEmoji extends CustomEmoji, UpdatableFromCache<KnownC
      * @return Whether this emoji is managed or not.
      */
     boolean isManaged();
+
+    /**
+     * Gets the creator of the emoji.
+     *
+     * @return The user who created the emoji.
+     */
+    CompletableFuture<Optional<User>> getCreator();
 
     /**
      * Updates the name of the emoji.
--- a/javacord-core/src/main/java/org/javacord/core/entity/emoji/KnownCustomEmojiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/emoji/KnownCustomEmojiImpl.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import org.javacord.api.entity.emoji.KnownCustomEmoji;
 import org.javacord.api.entity.permission.Role;
 import org.javacord.api.entity.server.Server;
+import org.javacord.api.entity.user.User;
 import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.listener.server.emoji.InternalKnownCustomEmojiAttachableListenerManager;
 import org.javacord.core.util.rest.RestEndpoint;
@@ -44,6 +45,11 @@ public class KnownCustomEmojiImpl extends CustomEmojiImpl
     private final boolean managed;
 
     /**
+     * Id of the user who created the emoji. 0 if no user was given.
+     */
+    private volatile long creatorId;
+
+    /**
      * Creates a new known custom emoji.
      *
      * @param api The discord api instance.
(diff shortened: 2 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/entity/emoji/KnownCustomEmoji.java
```diff
@@ -69,6 +70,13 @@ default CustomEmojiUpdater createUpdater() {
      */
     boolean isManaged();
 
+    /**
+     * Gets the creator of the emoji.
+     *
+     * @return The user who created the emoji.
+     */
+    CompletableFuture<Optional<User>> getCreator();
```
Vampire: Can there really be a custom emoji without a creator?

Bastian: I guess the field might not be present when you don't have enough permissions.
At least in the UI you can only view the creator of an emoji in the server settings:

javacord-core/src/main/java/org/javacord/core/entity/emoji/KnownCustomEmojiImpl.java
```diff
@@ -111,6 +118,38 @@ public boolean isManaged() {
...
     }
 
+    @Override
+    public CompletableFuture<Optional<User>> getCreator() {
+        if (creatorId == 0L) {
+            CompletableFuture<Optional<User>> future = new CompletableFuture<>();
+            new RestRequest<Long>(getApi(), RestMethod.GET, RestEndpoint.CUSTOM_EMOJI)
+                    .setUrlParameters(server.getIdAsString(), this.getIdAsString())
+                    .execute(result -> result.getJsonBody().has("user")
+                            ? result.getJsonBody().get("user").asLong(0)
```
Bastian: The endpoint returns a user entity, not a user id.
A response might look like this:
```json
{
   "managed":false,
   "name":"JC3",
   "roles":[

   ],
   "user":{
      "username":"Bastian",
      "discriminator":"0001",
      "id":"157862224206102529",
      "avatar":"a_796bef60840438a232f53f24ba29165c"
   },
   "require_colons":true,
   "animated":false,
   "id":"429927176268152842"
}
```

javacord-api/src/main/java/org/javacord/api/entity/emoji/KnownCustomEmoji.java
Bastian: {"managed": false, "name": "JC3", "roles": [], "require_colons": true, "animated": false, "id": "429927176268152842"}

Bastian: Just tested it.
With permissions:
```json
{"managed": false, "name": "JC3", "roles": [], "user": {"username": "Bastian", "discriminator": "0001", "id": "157862224206102529", "avatar": "a_796bef60840438a232f53f24ba29165c"}, "require_colons": true, "animated": false, "id": "429927176268152842"}
```
Without permissions:
```json
{"managed": false, "name": "JC3", "roles": [], "require_colons": true, "animated": false, "id": "429927176268152842"}
```

javacord-core/src/main/java/org/javacord/core/entity/emoji/KnownCustomEmojiImpl.java
```diff
@@ -111,6 +119,38 @@ public boolean isManaged() {
...
+                        }
+                        return null;
+                    })
+                    .exceptionally(t -> {
+                        future.completeExceptionally(t);
+                        return null;
+                    });
+            return future;
+        } else {
+            return getApi().getUserById(creatorId).thenApply(Optional::of);
```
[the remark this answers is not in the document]
Saladoc: It's `getUserById`, not `getCachedUserById`. The creator is always requested if an id is stored and may be served from cache if it is present there. The user will have to store the creator as long as they deem it relevant. I see no point in having the creators of every emoji that were requested once clutter the cache if they are not relevant for other reasons.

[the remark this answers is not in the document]
Bastian: My bad

javacord-core/src/main/java/org/javacord/core/entity/emoji/KnownCustomEmojiImpl.java
```diff
@@ -60,6 +66,7 @@ public KnownCustomEmojiImpl(DiscordApiImpl api, Server server, JsonNode data) {
             }
         }
         requiresColons = !data.hasNonNull("require_colons") || data.get("require_colons").asBoolean();
+        creatorId = data.has("creator") ? data.get("creator").asLong() : 0L;
```
Bastian: Why `creator` and not `user`?

Bastian:
There's still one open question I would like to answer: What happens when the creator of the emoji deletes his account? I can imagine that `getUserById(...)` will fail with a 404 Not Found, but the `user` field may still contain the old name and id of the user.

javacord-core/src/main/java/org/javacord/core/entity/emoji/KnownCustomEmojiImpl.java
Saladoc: missed that one when I corrected the rest.

Bastian:
As I have no clue how long it may take for account to be deleted, I'm going to merge the current state of the PR.
In case account deletion is causing issues, we can still fix it.

PR #393, opened 2018-10-20
Generalize asFoo methods

Saladoc:
Supersedes #302 

Basically just a cleanup of the work done by @cap5lut 

I have dropped the rename from `Specializable` to `Specifiable` and rewritten some javadocs, also applied many of Vampires suggestions.

re: `Similarly you can make a PotentialMessageAuthor and a PotentialAuditLogEntryTarget` 
not as per the definition of `Specializable` which intends on the *same* instance to be returned after a cast. Both `MessageAuthor` and `AuditLogEntryTarget` return a different value, with the `asUser` method on the latter being different in its return type.


Bastian:
Maybe I miss something, but what's the reason for the `ChannelSpecialization` interface?
Why not use `Specializable<Channel>`?

Vampire:
grafik
:-(

Bastian:
Didn't notice it :(

Saladoc:
Well it is just the checkstyle, so if you want to go down force push avenue again it will be an easy fix :^)

Vampire:
No, not if it is already in the official `development` branch, just a reminder to keep an eye open.

PR #394, opened 2018-10-20
Remove messages from cache on lost visibility.

### LINKED ISSUE #260: Message cache can be outdated
In some cases there might be messages in the cache which are no longer up to date:
- Message gets send in a channel with read-access
- Bot loses read-access for this channel
- Message gets edited
- Bot gains read-access again

=> The message content is out of date, but still in the cache

This can cause some issues with methods like `MessageEditEvent#getOldContent()`.

**Possible solution:**
- Remove a message from the cache when the bot loses read-access

Saladoc:
Fixes #260

Did I miss anything?


Bastian:
Looks good to me. Did you test the PR?

Also, what's about leaving and rejoining a server / a server becoming unavailable?

Saladoc:
I honestly had no opportunity to. Leaving a server / a server becoming unavailable are cases I did not think about when I made this, but I will add them later.

Saladoc:
Added removal on:
`RESUME` packet
leaving a server
server becoming unavailable

Bastian:
Why on resume?

Saladoc:
Bot connection dies -> no more updates are received
Change happens in the meantime
Bot resumes -> Cached message contents are out of date

Bastian:
https://discordapp.com/developers/docs/topics/gateway#resuming
> If successful, the gateway will respond by replaying all missed events in order, finishing with a Resumed event to signal replay has finished, and all subsequent events are new. 

If the bot resumes, it will not miss any events. And in case resuming is not possible, the cache get's cleared anyway.

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -106,6 +107,9 @@ private void handleServerChannel(JsonNode jsonChannel) {
                         (DispatchQueueSelector) channel.getServer(), channel.getServer(), channel, event);
             }
 
+            // Hacky way to be able to change the variable inside lambdas
+            final boolean[] isBotAffected = {false};
```
Vampire: A little less hacky would be to use an `AtomicBoolean`

Vampire:
Is this needed at all?
I think it shouldn't be necessary.
The only hard references (besides users or events hard-referencing them) are in the `MessageCache` that is in the text channel instance.
The `messages` in `DiscordApiImpl` are `Map<Long, WeakReference<Message>>`, so the message reference is only available as long as something else references the message.
And the `messagesCleanupQueue` given to the weak references makes sure that the map entries are thrown out of the map pretty soon after the message was garbage collected.
And as long as the message is referenced somewhere (be it an event or some user code) it is ok and intended to also have the message in this map.

Did you experience any problem / leak or did you just thought this would make sense?

Bastian:
> Is this needed at all?

- Message gets sent in a channel with read-access
- Bot leaves server
- Message gets updated
- Bot joins server
- Message gets updated again

=> `event.getOldContent()` is outdated if the GC did not ran.

> And as long as the message is referenced somewhere (be it an event or some user code) it is ok and intended to also have the message in this map.

Are you sure?

> Did you experience any problem / leak or did you just thought this would make sense?

I came up with it, but I think it's a valid issue

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelDeleteHandler.java
```diff
@@ -84,6 +85,10 @@ private void handleServerTextChannel(JsonNode channelJson) {
                 .ifPresent(server -> server.getTextChannelById(channelId).ifPresent(channel -> {
                     dispatchServerChannelDeleteEvent(channel);
                     ((ServerImpl) server).removeChannelFromCache(channel.getId());
+                    api.getCachedMessages().stream()
+                            .filter(msg -> msg.getChannel().getId() == channelId)
+                            .map(Message::getId)
+                            .forEach(api::removeMessageFromCache);
```
Vampire: I think this is an unnecessary performance hog, especially if you have a big bot in many big servers.
`api.getCachedMessages()` transforms the `Map` to a `MessageSet`, then you iterate all messages and compare the channel id just to remove the messages that should be the set of messages in the `MessageCache` of the channel in question.
Even if some messages that somehow are in the cache in api impl but not in the message cache of the channel I think it would be better to add a `getMessages` or `getMessageIds` method to `MessageCacheImpl` and then simply remove those in the handlers.

And the same should most probably be done for private and group channels, not only for server text channels.
I know that bot cannot join group DMs, but where we support self-bot things already we can also maintain them properly right away. :-)

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
Vampire: And a different name.
Javacord is not only able to handle bot accounts but also user accounts.
And the naming for such things in Javacord unfortunately is `you`, `yourself` and so on.
So this should most probably be `AtomicBoolean areYouAffected = new AtomicBoolean()` to stay consistent. :-)

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -196,8 +205,16 @@ private void handleServerChannel(JsonNode jsonChannel) {
...
                             channel, PermissionsImpl.EMPTY_PERMISSIONS, oldPermissions, role);
+                    isBotAffected[0] |= role.getUsers().stream().anyMatch(User::isYourself);
                 });
             }
+
+            if (isBotAffected[0] && !channel.canYouSee()) {
+                api.getCachedMessages().stream()
+                        .filter(msg -> msg.getChannel().getId() == channelId)
+                        .map(Message::getId)
+                        .forEach(api::removeMessageFromCache);
```
Vampire: same as above

javacord-core/src/main/java/org/javacord/core/util/handler/guild/GuildDeleteHandler.java
```diff
@@ -33,6 +34,10 @@ public void handle(JsonNode packet) {
 
                 api.getEventDispatcher().dispatchServerBecomesUnavailableEvent(
                         (DispatchQueueSelector) server, server, event);
+                api.getCachedMessages().stream()
+                        .filter(msg -> msg.getServer().map(s -> s.getId() == serverId).orElse(false))
+                        .map(Message::getId)
+                        .forEach(api::removeMessageFromCache);
```
Vampire: same as above

javacord-core/src/main/java/org/javacord/core/util/handler/guild/GuildMemberRemoveHandler.java
```diff
@@ -34,7 +35,13 @@ public void handle(JsonNode packet) {
                     ServerMemberLeaveEvent event = new ServerMemberLeaveEventImpl(server, user);
 
                     api.getEventDispatcher().dispatchServerMemberLeaveEvent(server, server, user, event);
+                    if (user.isYourself()) {
+                        api.getCachedMessages().stream()
+                                .filter(msg -> msg.getServer().map(server::equals).orElse(false))
+                                .map(Message::getId)
+                                .forEach(api::removeMessageFromCache);
```
Vampire: Isn't this better suited in the `GuildDeleteHandler` in the branch where the `ServerLeaveEvent` is dispatched?
Is the `GUILD_MEMBER_REMOVE` event even triggered when the leaving user is you?
Even if it is so, I think it belongs to where the according event is dispatched.

and same as above

javacord-core/src/main/java/org/javacord/core/util/handler/guild/GuildMemberUpdateHandler.java
```diff
@@ -95,6 +100,18 @@ public void handle(JsonNode packet) {
...
+
+                    if (user.isYourself()) {
+                        Set<Long> unreadableChannels = server.getTextChannels().stream()
+                                .filter(((Predicate<ServerTextChannel>)ServerTextChannel::canYouSee).negate())
+                                .map(ServerTextChannel::getId)
+                                .collect(Collectors.toSet());
+                        api.getCachedMessages().stream()
+                                .filter(msg -> unreadableChannels.contains(msg.getChannel().getId()))
+                                .map(Message::getId)
+                                .forEach(api::removeMessageFromCache);
```
Vampire: same as above

javacord-core/src/main/java/org/javacord/core/util/handler/guild/role/GuildRoleUpdateHandler.java
```diff
@@ -97,6 +103,17 @@ public void handle(JsonNode packet) {
...
+                // If bot is affected remove messages from cache that are no longer visible
+                if (role.getUsers().stream().anyMatch(User::isYourself)) {
+                    Set<Long> unreadableChannels = role.getServer().getTextChannels().stream()
+                            .filter(((Predicate<ServerTextChannel>)ServerTextChannel::canYouSee).negate())
+                            .map(ServerTextChannel::getId)
+                            .collect(Collectors.toSet());
+                    api.getCachedMessages().stream()
+                            .filter(msg -> unreadableChannels.contains(msg.getChannel().getId()))
+                            .map(Message::getId)
+                            .forEach(api::removeMessageFromCache);
```
Vampire: same as above

Bastian:
Any updates here?

javacord-core/src/main/java/org/javacord/core/util/handler/guild/GuildMemberRemoveHandler.java
Saladoc: it does seem to be only triggered when other users leave the server.

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
Bastian: Still unresolved

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -163,6 +167,9 @@ private void handleServerChannel(JsonNode jsonChannel) {
                         overwrittenPermissions.put(entity.getId(), newOverwrittenPermissions);
                         dispatchServerChannelChangeOverwrittenPermissionsEvent(
                                 channel, newOverwrittenPermissions, oldOverwrittenPermissions, entity);
+                        areYouAffected.compareAndSet(false,entity instanceof User && ((User) entity).isYourself());
+                        areYouAffected.compareAndSet(false,entity instanceof Role
```
Bastian: I don't want to be nitpicking, but there's missing a space (also in the line above). 😉 
```suggestion
                        areYouAffected.compareAndSet(false, entity instanceof Role
```

Saladoc:
Should be good to merge now.

Bastian:
Looks good to me, thx. 👍 
I'll merge it after the conflicts get resolved.

Issue #399, opened 2018-11-05
Utility Methods for Role Management


Saladoc:
Add the following methods in `Server`:
 - `canManageRoles(User)` and `canYouManageRoles()` to check whether a `User` has the *general ability* to manage roles on the server
 - `canManageRolesOf(User, User)` and `canYouManageRolesOf(User)` to check whether a user has the ability to manage roles of *a specific other user* 

Possibly also a utility method to check whether a user can manage a specific role?

PR #400, opened 2018-11-08
Implemented Role utility methods

burdoto:
Closes https://github.com/Javacord/Javacord/issues/399

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -2233,6 +2233,30 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
     }
 
     /**
+     * Checks if the user can manage the roles of the target user.
+     *
+     * @param user The user that tries to manage roles.
+     * @param target The user whose roles are to be managed.
+     * @return Whether the user can manage the target's roles.
+     */
+    default boolean canManageRolesOf(User user, User target) {
+        return canManageRole(user, getHighestRole(target).orElseGet(this::getEveryoneRole));
+    }
+
+    /**
+     * Checks if the user can manage the target role.
+     *
+     * @param user The user that tries to manage the role.
+     * @param target The role that is to be managed.
+     * @return Whether the user can manage the role.
+     */
+    default boolean canManageRole(User user, Role target) {
+        return canManageRoles(user)
+                && getRoles().contains(target)
+                && getHighestRole(user).orElseGet(this::getEveryoneRole).compareTo(target) >= 0;
+    }
+
+    /**
      * Checks if the given user can manage the server.
      *
      * @param user The user to check.
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
@@ -79,6 +79,16 @@ public interface User extends DiscordEntity, Messageable, Nameable, Mentionable,
     Optional<Activity> getActivity();
 
     /**
+     * Checks if the user can manage the target role.
+     *
+     * @param role The role that is to be managed.
+     * @return Whether the user can manage the role.
+     */
+    default boolean canManageRole(Role role) {
+        return role.getServer().canManageRole(this, role);
+    }
+
+    /**
      * Gets the server voice channels the user is connected to.
      *
      * @return The server voice channels the user is connected to.


javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2224,6 +2224,29 @@ default boolean canYouManageRoles() {
...
+    /**
+     * Checks if the user can manage the target role.
+     *
+     * @param user The user that tries to manage the role.
+     * @param target The role that is to be managed.
+     * @return Whether the user can manage the role.
+     */
+    default boolean canManageRole(User user, Role target) {
+        return canManageRoles(user)
+                && getHighestRole(user).orElseGet(this::getEveryoneRole).getPosition() <= target.getPosition();
```
Bastian: Shouldn't this be `>=`?
I'm also not sure if you can manage roles that have the same position like your own role. Did you check this?

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2224,6 +2224,29 @@ default boolean canYouManageRoles() {
...
 
+    /**
+     * Checks if the user can manage the roles of the target user.
+     *
+     * @param user The user that tries to manage roles.
+     * @param target The user whose roles are to be managed.
+     * @return Whether the user can manage the target's roles.
+     */
+    default boolean canManageRolesOf(User user, User target) {
+        return canManageRole(user, getHighestRole(target).orElseGet(this::getEveryoneRole));
```
Bastian: If one of the two users is not in the server, it should also always return `false`:
```java
return getMembers().contains(user) 
        && getMembers().contains(target) 
        && canManageRole(user, getHighestRole(target).orElseThrow(AssertionError::new));
```
`getHighestRole(target)` only returns an empty Optional, if the user is not in the server.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2224,6 +2224,29 @@ default boolean canYouManageRoles() {
...
+    }
+
+    /**
+     * Checks if the user can manage the target role.
+     *
+     * @param user The user that tries to manage the role.
+     * @param target The role that is to be managed.
+     * @return Whether the user can manage the role.
+     */
+    default boolean canManageRole(User user, Role target) {
```
Bastian: Can you also add `User#canManageRole(Role)` please :)

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
Vampire: Besides that also here the `orElse` if probably wrong, assuming this follows the same semantics as usual, then if the position is equal, the id is taken into account. Don't compare the positions actually, `Role` implements `Comparable` which properly follows the algorithm.

burdoto: About 95% sure same roles can modify each other,didn't check

Not sure whether @everyone is on the 0 end of the list, if so, the `<=`comparison is wrong and needs to be turned into `>[=]`; after I checked the first thing.

The comparable algorithm doesn't matter here as it only matters whether the role is EQUAL position (= equal role) or HIGHER position; no need to check anything else than the position, yet a `#compare` call might look a slightly bit less funky

WalkerKnapp: You can not manage roles the same position as your own.

Vampire: > You can not manage roles the same position as your own.

Sure you can, if your role is higher.
If the position is the same, the id is compared which cannot be the same, except the role is exactly the same.
That's why one should not compare the position but use the `Role#compareTo` method which does exactly that, it compares the position and if equal, the id.

> About 95% sure same roles can modify each other,didn't check

Then please check, I think they cannot.
And I do not mean different roles with same position, as this is not significant enough, as the ID is then taken into account as described above.

> Not sure whether @everyone is on the 0 end of the list, if so, the <=comparison is wrong and needs to be turned into >[=]; after I checked the first thing.

`@everyone` role always has the lowest position. If someone has no other role `getHighestRole` will return the `@everyone` role, which is why the `orElseGet` is totally wrong. If `getHighestRole` does return an empty optional, the user is not a member of the server.

> The comparable algorithm doesn't matter here as it only matters whether the role is EQUAL position (= equal role) or HIGHER position; no need to check anything else than the position, yet a #compare call might look a slightly bit less funky

You are wrong. The position is **not** significant enough if it is equal, which is why `Role#compareTo` then takes the ID into account.

Just try it out yourself.
Create two roles, make sure they have the same position, then try from each other to manipulate the other. You will see that one works and the other does not.
If it works both ways or doesn't work both ways either you or Walker were right, but I don't think so.
That one of you says it works and one says it does not, just confirms what I said, that it might work or might not work, because it additionally depends on the ID if the position is equal.

burdoto: first of all, how do you put two roles on the same position?

burdoto:
Requested changes should be done 👍

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
Bastian: There's a difference between what is displayed in the UI and the `position` field of a role. It's possible that two roles have the same `position` field value in which case they are ordered by `id` in the UI. That's not what you would expect l, but it's the same for channel ordering as well. For channel ordering we have `getRawPosition()` and `getPosition()` though. I guess we should add this for roles as well.

I didn't test it, but I think you can produce this result by just creating two new roles.

Vampire: Afair if you simply create new ones they automatically get unique positions. Iirc you must make an update REST call that puts them on the same position and this outside Javacord as you can only give a list of roles through the Javacord API.

Bastian:
Thx 👍

Issue #360, opened 2018-08-06
Methods to set a proxy for REST-calls / the websocket


Bastian:
At the moment it is not possible to set a proxy for the REST-calls and the websocket.

Vampire:
Actually for the REST calls there is.
OkHttp uses the system default proxy selector if no proxy and no proxy selector is set explicitly, so the standard Java properties for setting a proxy should at least work fine with the REST part for which we use OkHttp.

Vampire:
Besides the standard system properties, the system default proxy selector should also be taken into account.
So as the system default proxy selector handles the standard system properties basically it should probably be:
- If proxy is set explicitly via `DiscordApiBuilder#setProxy`, use it
- If not, but proxy selector is set via `DiscordApiBuilder#setProxySelector`, use it
- If not, use the system default proxy selector from `ProxySelector.getDefault()`
This should probably be all that is necessary to have it correctly, besides some docs.

Vampire:
I'm working on this, you can assign me here

PR #403, opened 2018-11-27
Proxy Support

Vampire:
This fixes / implements #360.

It adds and tests as far as possible proxies of various types, with and without authentication where specified.
Unfortunately the NV WebSocket library is particularly bad in supporting proxies.
Due to that solely HTTP proxies are supported, with and without authentication.

SOCKS4 proxies are not supported as OkHttp has a bug they refuse to fix or accept a PR for
and also because NV WebSocket supports neither SOCKS4 nor SOCKS5 proxies at the moment.

If we switch away from NV WebSocket to another WebSocket lib (like OkHttp) that has better proxy support,
we might be able to support at least SOCKS5 in the future.

If we switch away from both OkHttp and NV WebSocket and for example require at least Java 9 and use the
new http client and web socket standard implementation in the JRE, then we might be able to support all
kinds of proxies.

Anyway for all possible combinations there are tests, just the ones that are currently not supported due to
libraries are marked with `@PendingFeature` which means they are ignored as long as they fail and will
start failing as soon as they are working as designed, so that you see that the feature is working now
and the test can be un-annotated and put into production.


Bastian:
Closes #360

Issue #353, opened 2018-08-06
Add listeners before logging in


Bastian:
In some cases it might be useful to register listeners before logging in (aka. in the `DiscordApiBuilder`), e.g. the `ServerBecomesAvailableListener`. Otherwise you might miss some events.

PR #407, opened 2018-12-08
Allow event listener registration during api creation

Saladoc:
Untested for now. Since the `GloballyAttachableListenerManager` interface is a) autogenerated and b) huge, I've settled for a single method requiring the class also passed.

Implements #353

### DIFF
--- a/gradle/listener-manager-generation.gradle
+++ b/gradle/listener-manager-generation.gradle
@@ -13,6 +13,8 @@ import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeS
 
 import javax.annotation.Generated
 import javax.naming.ConfigurationException
+import java.util.function.Function
+import java.util.function.Supplier
 import java.util.stream.Collectors
 import java.util.stream.Stream
 
@@ -114,6 +116,8 @@ task generateListenerManagers {
 
         def discordApiName =
                 typeSolver.solveType('org.javacord.api.DiscordApi').qualifiedName
+        def discordApiBuilderName =
+                typeSolver.solveType('org.javacord.api.DiscordApiBuilder').qualifiedName
         def discordApiImplName =
                 typeSolver.solveType('org.javacord.core.DiscordApiImpl').qualifiedName
         def listenerManagerName =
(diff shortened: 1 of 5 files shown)


Vampire:
Besides that we have similar methods that are named `addListener`, not `addListenerFor`, one of the reasons to have the listener manager interfaces was / is, that `DiscordApi` and `DiscordApiBuilder` can both implement it.
I don't think that this should he merged without the builder implementing the interface, sorry.

Saladoc:
The `DiscordApiBuilder` is unable to satisfy the contract of the `GloballyAttachableListenerManager` as it requires `ListenerManager`s to be returned which is impossible since at the moment of adding the listener there is no `DiscordApi` to create the manager with. Unless you want to duplicate the entire interface (or change it to return `CompletableFuture<ListenerManager<T>>`) you have exactly zero chances of implementing it.

Vampire:
Hm, didn't think about that, but actually the `ListenerManager` interface nowhere has the `DiscordApi` instance in its interface, so there could well be a second implementation that defers any calls that need the api until the api is available. and works on all shards or maybe it could even be considered to return `null` in that case or refactor to return `Optional<ListenerManger>` instead but that would be a breaking change. But as I said, maybe it could be considered returning `null` in this special case, especially as you can easily get hold of the `ListenerManager` later when the actual `DiscordApi` instance is available by just trying to register the same listener again.
@Bastian what do you think?

Vampire:
I'm just not sure how Bastian will like if the listener manager handles multiple shards as he didn't want something like a "shard manager".

Bastian:
To summarize what we discussed in the Discord server:
* The api builder should still be chainable
* For global listeners, the `ListenerManager` is not that important, so it's no big deal to not have it.
* A user should be able to have different listener instances per shard

The methods may look like this:
```java
// One listener for all shards
default ApiBuilder addXyzListener(XyzListener listener) {
    addXyzListener(api -> listener);
    return this;
}

// A new listener instance per shard
default ApiBuilder addXyzListener(Supplier<XyzListener> supplier) {
    addXyzListener(api -> supplier.get());
    return this;
}

// A new listener instance per shards with access to the api object in the constructor
ApiBuilder addXyzListener<Function<DiscordApi, XyzListener> supplier);
```

The methods that accept a supplier or function can be used like this:
```java
apiBuilder
    .addXyzListener(MyXyzListener::new)
    .login();
```
with
```java
public class MyXyzListener implements XyzListener {

    // Can have the api as parameter (but does not have to)
    public MyXyzListener(DiscordApi api) {
        System.out.println("Created xyz listener for shard " + api.getCurrentShard());
    }

    @Override
    onXyzEvent(XyzEvent event) {
        // ...
    }
}
```

The method that just accepts a listener can be used like all other listeners:
```java
apiBuilder
    .addXyzListener(event -> ...)
    .addXyzListener(new MyXyzListener())
    .login();
```


The implementation should be pretty straight-forward:
- Write the code to generate the required classes and interfaces 
- Register the listeners directly **after** the `READY`-packet was received. Do not dispatch or handle any packets until the listeners are registered.

Saladoc:
I have finished the early event registration. Also, since the `DiscordApiBuilderDelegate` is shipped with the api but needs to implement `ChainableGloballyAttachableListenerManager`, I have made the interface generic so that the delegate can fulfil the method contracts.

Listeners are created and registered immediatly after receiving the ready packet, even before the application info is set. Please check if that would be a problem. 

So far the latest changes are untested, but this PR is ready for review.

Bastian:
Works like a charm 👌 
Only the JavaDocs of `DiscordApiImpl`'s constructor needs to be cleaned up.

PR #416, opened 2018-12-29
ServerChannel should extend Comparable

burdoto:
A ServerChannel should extend Comparable<ServerChannel>, because it can be sorted by position

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannel.java
@@ -26,7 +26,8 @@ import java.util.concurrent.CompletableFuture;
 /**
  * This class represents a server channel.
  */
-public interface ServerChannel extends Channel, Nameable, ServerChannelAttachableListenerManager {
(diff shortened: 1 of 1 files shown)


Bastian:
It does take categories into account.
`getPosition()` takes already care about everything.

Bastian:
I'm also unsure about the `IllegalArgumentException`.

Saladoc:
If channels are on different servers, they should not be able to be compared. What other way do we have of dealing with this?

Bastian:
Ok, in `Role` we also throw an exception: https://github.com/Javacord/Javacord/blob/development/javacord-core/src/main/java/org/javacord/core/entity/permission/RoleImpl.java#L266

Vampire:
Yes, comparable defines the natural order of things and roles or channels from different servers just have no natural order, so there is no meaningful result you could return.

burdoto:
Changes applied 👍

Bastian:
Thanks 👍 
I've added some JavaDoc comments similar to the ones we use in `Role#compareTo(..)` to your commit. I hope you don't mind :)

PR #421, opened 2019-01-15
Remove object listeners on channel/server/message deletion.

### LINKED ISSUE #365: Cleanup message listeners on message deletion
Someone told me, that message listeners - that are attached to a message - don't get removed when the message get's deleted.
If this is the case, it might be a good idea to automatically remove them, as a message cannot receive any new events after the `MessageDeleteEvent`.

Saladoc:
Fixes #365


Bastian:
It's possible to attach remove-handlers to listener managers. These should be called when you remove the listener.
See [JavaDocs](https://docs.javacord.org/api/v/3.0.1/org/javacord/api/util/event/ListenerManager.html#addRemoveHandler-java.lang.Runnable-)

PR #425, opened 2019-01-17
Allow to attach MessageCreateListeners to Webhooks

Vampire:


Bastian:
I guess it would also make sense to be able to attach listeners to `MessageAuthors` then?
But this can easily be done in a separate PR.

Vampire:
What would that do? Just forward to the user or webhook and throw if webhook but not message create listener? Usually we don't do such stuff but first have to cast or use as....

Bastian:
I'm only talking about message create listeners.
And no, don't "forward" them to the User or Webhook object, as you can only access the Webhook object if you have the appropriate permissions (+ it would be an asynchronous REST-Request).

Vampire:
Hm, I'm still unsure about the `MessageAuthor`.
The question remains, should it "forward" (logically, not physically) to the `User` and `Webhook`, so that if the listeners for the user are retrieved it is also returned, or should `MessageAuthor` be a first-class citizen and so on.

Bastian:
I think it would be best to forward it logically, but this should be done in another PR.

PR #426, opened 2019-01-20
Support for mobile indicator

Saladoc:
With the last update, a `PRESENCE_UPDATE` packet might include the `client_status` node with up to three of the nodes `web`, `mobile` and `desktop`. They have string values similar to the `status` node, except that `offline` is indicated by them being absent.

These mobile and desktop statuses, if they are both present, seem to always have the same value as discord persists the user status over all platforms. So, if we take that assumption for granted, they could get replaced with methods returning a `boolean`.

Also: Should the connection / disconnection of another device trigger an event even if it doesn't change the global online state of a user? I've gone with "no" so far.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/DiscordClient.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/DiscordClient.java
@@ -1,0 +1,46 @@
+package org.javacord.api.entity;
+
+/**
+ * Represents the three different Discord clients.
+ */
+public enum DiscordClient {
+
+    /**
+     * The Discord desktop client.
+     *
+     * <p>Does not differentiate between operating systems like Windows or Linux.
+     */
+    DESKTOP("desktop"),
+
+    /**
+     * The Discord mobile client.
+     *
+     * <p>Does not differentiate between the iOS app and the Android app.
+     */
+    MOBILE("mobile"),
+
+    /**
+     * The Discord web client.
+     */
+    WEB("web");
+
+    private final String name;
+
+    /**
+     * Creates a new Discord Client.
+     *
+     * @param name The name for the client, as it appears in the Discord REST-api.
+     */
+    DiscordClient(String name) {
+        this.name = name;
+    }
+
+    /**
+     * Gets the name of client, as it appears in the Discord REST-api.
+     *
+     * @return The name of the client.
+     */
+    public String getName() {
+        return name;
+    }
+}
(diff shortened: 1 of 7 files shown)


Bastian:
There are still some open questions / missing implementations:
* Does it also exist in the `presences` field of a server? Most likely yes (see https://github.com/Javacord/Javacord/blob/3.0.1/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java#L315-L337).
* Can you update the status? Or is it a client-only feature?
* Is this API considered stable (on Discord's side)? In the worst case we implement it now and 2 days later it no longer works because Discord changed something. As long as it's not in the official docs, it's not guaranteed to stay the same.

If the status is really always the same across devices, I would go with a simple boolean flag instead. But then I'm curious what's Discord reason to implement it this way.

About events:
We should definitely have some kind of event that gets triggered when a user changes its status on one device (e.g. switched from mobile to desktop). I'm just unsure what event it should be. I guess we can/should just extend the `UserChangeStatusEvent` with this new information, but I'm unsure about the method names. When we use the enum, we can just keep the pattern `getOldDesktopStatus()`, `getNewDesktopStatus()` etc. but with boolean flags it's harder (`getOldDesktopOnlineFlag()`? 😕 ).

Saladoc:
- I have not verified if it's sent with servers as well. 
- I don't believe it will be a bot feature, although I have not yet thought to verify the status for a bot user.
- As long as it's not in the official API docs, we have no way of knowing. Everything in this PR is found out via hints and experimental verification. I have found the connection status to always be the same (and it change in my desktop client when I changed it on my mobile device) but I'm not sure if that's something we should rely on.

Bastian:
I feel like we cannot expect an update of the official documentation soon, so here's my suggestion (which is very similar to your first suggestion on Discord):
* Add new enum `DiscordClient` with the values `WEB`, `MOBILE` and `DESKTOP`
* Add a method `User#getStatusOnClient(DiscordClient)` that returns a `UserStatus`
* Keep `User#getStatus()`
* Extend the `UserChangeStatusEvent` with `getOldStatusOnClient(DiscordClient)` and `getNewStatusOnClient(DiscordClient)`, as well as `hasStatusChangeOn(DiscordClient)`

And of course:
* Verify if it's send in the server JSON as well (I'm very sure that it is) and handle it
* Update the `PresenceUpdateHandler` to fire a `UserChangeStatusEvent` on device-status changes as well.

PR #431, opened 2019-01-23
[WIP] Rework Embed Structure

burdoto:
Closes https://github.com/Javacord/Javacord/issues/358

If there are any suggestions, ideas, requests, etc., let me know.
I chose a backwards compatible and intuitive implementation, that will be able to create new `EmbedDraft` instances from any kind of `BaseEmbed`.
In my idea, `EmbedDraft` extends `EmbedBuilder` so that you are still able to modify the `EmbedDraft` right before sending it.
I am modifying things like the `EmbedBuilderDelegate` because I require extra getter methods with the `EmbedDraft` class that I am working on.


burdoto:
*forcepushed to be able to split the rework in several commits*

javacord-api/src/main/java/org/javacord/api/entity/message/embed/Embed.java
```diff
@@ -102,6 +104,7 @@
      * @return The fields of the embed.
      */
     List<EmbedField> getFields();
+@Deprecated
```
Saladoc: Deprecation annotation needs to be on the deprecated Type, method or member, not at a random position in the body.

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedAuthor.java
```diff
@@ -31,4 +33,5 @@
      */
     Optional<URL> getProxyIconUrl();
 
+@Deprecated
```
Saladoc: Same as on `Embed.java`

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedField.java
```diff
@@ -21,4 +23,5 @@
      */
     boolean isInline();
 
+@Deprecated
```
Saladoc: Same as on `Embed.java`

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedFooter.java
```diff
@@ -28,5 +30,6 @@
      * @return The proxy url of the footer icon.
      */
     Optional<URL> getProxyIconUrl();
+@Deprecated
```
Saladoc: Same as on `Embed.java`

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedImage.java
```diff
@@ -35,4 +37,5 @@
      */
     int getWidth();
 
+@Deprecated
```
Saladoc: Same as on `Embed.java`

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedProvider.java
```diff
@@ -16,4 +18,5 @@
      */
     URL getUrl();
 
+@Deprecated
```
Saladoc: Same as on `Embed.java`

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedThumbnail.java
```diff
@@ -34,5 +36,6 @@
      * @return The width of the thumbnail.
      */
     int getWidth();
+@Deprecated
```
Saladoc: Same as on `Embed.java`

javacord-api/src/main/java/org/javacord/api/entity/message/embed/EmbedVideo.java
```diff
@@ -28,4 +30,5 @@
      */
     int getWidth();
 
+@Deprecated
```
Saladoc: Same as on `Embed.java`

burdoto:
> 
> 
> What purpose does the `Delegated` interface serve? Why do none of the other classes that internally use delegates implement / extend it?
> IMO the use of delegates is an implementation detail that should not be exposed unless strictly necessary.

I want to be able to get the delegate from within the interfaces, otherwise the implementations must implement `#asEmbedDraft` etc methods. 
@Bastian What do you think about this?

burdoto:
> 
> 
> Why is the old Embed deprecated? Wouldn't it be better to use it instead of `BaseEmbed` and try to get as many getters as possible into this interface?
The old embed structure cannot be used as the base and still maintain backwards compatibility, as the old Embed structure is made for already sent embeds, not for EmbedDrafts - I can NOT have methods like `#getVideo` accessible in EmbedDraft - What am i gonna do? Throw an Exception? Bad Idea.
 
> Also it is good practise with git to let text files end on a line separator (= have an empty line at the bottom of a file). None of your files have this.
This is why the PR has a large, uppercase `[WIP]` in the title. no need to be picky.

javacord-api/src/main/java/org/javacord/api/entity/message/embed/Embed.java
burdoto: grafik

burdoto: grafik

Saladoc:
> The old embed structure cannot be used as the base and still maintain backwards compatibility, as the old Embed structure is made for already sent embeds, not for EmbedDrafts - I can NOT have methods like `#getVideo` accessible in EmbedDraft - What am i gonna do? Throw an Exception? Bad Idea.

It's a breaking change either way, so no backwards compatibility is needed. Functions that no longer can get served by all implementations simply will be removed from the interface. 

> This is why the PR has a large, uppercase `[WIP]` in the title. no need to be picky.

What is the point of having the PR up if not for feedback? If it's up and I spot an issue that is not mentioned yet, I point it out. This has nothing to do with being picky, I'm just assuming you're looking for feedback and offering what I can. It is up to you to decide if, how and when to address it.

Bastian:
Feel free to take a look at this commit to get a basic idea of what I was thinking of: https://github.com/Bastian/Javacord/commit/c1ae61ea9b098b0fe944d99bf38ce96b5440b758

I didn't work on it for quite a while, but iirc the main interfaces are down.

burdoto:
@Bastian#0001 I have one single important question:
Should an `EmbedDraft` be modifyable after building it or not; moreso: Should `EmbedDraft` extend `EmbedBuilder`?

Pros:
- You can send the Draft, modify it, then edit the message with the modified `EmbedDraft` without creating a new Builder from the `SentEmbed` that you recieve with the message or keeping the Builder around.
- A "Draft" named Structure seems "temporary" enough (to me) to justify being able to modify it.

Cons:
- The `EmbedBuilder` would basically become useless. This might not be a con; but the overall Builder-Object structure gets kinda obsolete if you can just modify an `EmbedDraft` instead. However, the `EmbedDraft` can only be instantiated with an `EmbedBuilder` object, not without parameters.

burdoto:
I am really uncomfortable with the current URL solution:
- The `EmbedBuilder` accepts a `String` as URL, but due to the fact that the `EmbedDraft` members require a built `URL` for modifying. 
- I have to build the `URL`s myself within the implementations and just set the `URL` to `null` when the provided `String` is a malformed URL.
- This is not very intuitive and will lead to confusion.

Proposal:
- The `EmbedBuilder` should either throw a `MalformedUrlException` on the `#setUrl(String)` methods AND/OR only accept `URL` as a parameter for such methods.
- This way the end user has the ability to handle malformed url inputs, e. G. from the user interfacing with the bot.
- The end user must provide a valid URL due to the constructor of `java.net.URL`.

Second proposal:
- Change `URL` to `URI` type. 
- This allows us, in case of an attachment, to use the legacy field for storing the reference to the attachment (which starts with `attachment://` which is an `URI` prefix).
- This may lead to confusion on the end user if we accept `URI`s instead of `URL`s. However, it is possible to accept `URL`s and build an `URI` from that, because every valid URL is also a valid URI.

burdoto:
In the current state, this Branch does compile.
It has not been tested yet.

JavaDocs are being added after testing has been done.

burdoto:
In the current state, this branch does compile.
The API should be finished at this point, now I will start testing and might be changing some things.
It has not been tested yet.

JavaDocs are being added after testing has been done.

burdoto:
You might want to consider moving `FileContainer` to the API, or create another version of that for the end user.

The problem is that e.g. any `SentEmbedThumbnail` will contain an URL to the image, but not every `EmbedDraftThumbnail` points to an URL, it might point to an attachment; this is code from the [EmbedBuilderDelegateImpl](https://github.com/burdoto/Javacord/blob/embed-structure/javacord-core/src/main/java/org/javacord/core/entity/message/embed/EmbedBuilderDelegateImpl.java#L123-L130) that can apply a previously created `EmbedThumbnail` to an `EmbedBuilder`:
```java
    @Override
    public void setThumbnail(BaseEmbedThumbnail thumbnail) {
        setThumbnail(thumbnail.getUrl().orElse(null));
        // if the url is empty, thumbnail is an EmbedDraftThumbnail with an attachment
        if (thumbnail.getUrl().isEmpty()) {
            this.thumbnailContainer = ((EmbedDraftFileContainerAttachableMember) thumbnail).getContainer();
        }
    }
```

As you can see; I need to use an `Optional<URL>` as the return value for `BaseEmbedThumbnail#getUrl` due to the fact that an `EmbedDraftThumbnail` may not point to an URL.
This can lead to situations in which the end user might call `#getUrl` on an `EmbedDraftThumbnail` and gets an empty optional.
We have two options here:
- Document `#getUrl` to say that if the returned Optional is empty that the `EmbedThumbnail` is linked to an attachment
- Add an `EmbedAttachmentMember` interface (or something similar) that contains a method `#getAttachment` or `#getFileContainer` to enable the end user to recieve the attachment. Such an interface could also have methods like `#hasAttachment`, `#hasURL` and `FileState getFileState` with ```enum FileState { ATTACHMENT, URL, NONE }```.
All those are just possibilities that I managed to get from my head. I will not implement any of those yet, please talk to me about what solution you want to target.

Bastian:
I see the problem, but I neither want to expose the `FileContainer` nor use `URI`s.
I would suggest to just use `String` instead of `URL` everywhere embed-related.

PR #432, opened 2019-01-28
Add isBotUser() and isRegularUser() to MessageAuthor

Saladoc:


Vampire:
Hm, I remember we talked about this method previously (Bastian and I afair, but in some channel) and came to the conclusion not to add it, but I don't remember the reasoning.

Vampire:
I think we also considered those and decided against them. :-D

Saladoc:
Not having a method to check if a `MessageAuthor` is a bot account is just plain counterintuitive though as it is one of the most common use cases so we should include some way.

I'd probably do all three `isUser()`, `isBotUser()`, and `isHumanUser()`, although I dislike the latter name.

Vampire:
Please read https://discordapp.com/channels/151037561152733184/354648262453624833/467999839683215372 until 17.07. for reference.

PR #435, opened 2019-02-01
Support for Spoilers

Saladoc:
As of today's update, both textual and attachment spoilers are live in the discord client.

Since adding a "spoiler: true/false" field to the json would be the sensible thing to do, discord decided to instead prefix file names with `SPOILER_` to indicate a file should be considered a spoiler. However, this saves us waiting for a new documentation since the file name is something we can already control.

This PR implements
 - [x] Checking the spoiler status of a `MessageAttachment`
 - [x] Adding spoiler-marked attachments via the `MessageBuilder`
 - [ ] Sending spoilerific messages and attachments directly from the `Messageable` interface

I'd like some input with regard to method names for `Messageable` since `sendMessageAsSpoiler` would be confusing as to whether the text or the files would be marked as spoilers.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageAttachment.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageAttachment.java
@@ -85,4 +85,16 @@ public interface MessageAttachment extends DiscordEntity {
      */
     CompletableFuture<BufferedImage> downloadAsImage();
 
+    /**
+     * Checks whether the attachment is marked as a spoiler.
+     *
+     * <p>Discord encodes the information on whether a file is considered a spoiler in the file name. Any file whose
+     * filename starts with {@code SPOILER_} is considered a spoiler.
+     *
+     * @return The spoiler status.
+     */
+    default boolean isSpoiler() {
+        return getFileName().startsWith("SPOILER_");
+    }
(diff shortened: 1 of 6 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/MessageAttachment.java
```diff
@@ -85,4 +85,11 @@ default boolean isImage() {
      */
     CompletableFuture<BufferedImage> downloadAsImage();
 
+    /**
+     * Checks whether the attachment is marked as a spoiler.
+     *
+     * @return The spoiler status.
+     */
+    boolean isSpoiler();
```
Bastian: I think it would be enough to have this as a default methods that uses the `getFileName()` method.
Would you also add some information in the JavaDoc, that it just checks if the file name starts with `SPOILER_`?

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -77,6 +78,7 @@ public MessageBuilder append(Mentionable entity) {
         return this;
     }
 
+
```
Bastian: why?

PR #437, opened 2019-02-06
Adds MessageAttachment#downloadAsInputStream

WalkerKnapp:
This is more performant and more memory efficient for large files than #downloadAsByteArray, and already has a precedent in Icon#asInputStream().

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageAttachment.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageAttachment.java
@@ -3,6 +3,8 @@ package org.javacord.api.entity.message;
 import org.javacord.api.entity.DiscordEntity;
 
 import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


Vampire:
Why does it need to be a `CompletableFuture` when it is a stream?
A stream should be streamed, so it wouldn't make much sense to put this in future I'd say.

WalkerKnapp:
I'm not sure why, but FileContainer#asInputStream sets up the connection on a separate thread and returns a CompletableFuture<InputStream>.

WalkerKnapp:
Also, the naming conventions here are a bit inconsistent. Icon has `#asByteArray`, `#asInputStream`, and `#asBufferedImage`, but MessageAttachment prefixes all of these methods with `download`. I'm not sure which is better, but one should be used everywhere over the other.

Vampire:
My guess would be historically grown / copied from v2 without too much thought.
I would have guessed there was one `download` method that returned a byte array and then two methods `downloadAsByteArray` and `downloadAsImage` for convenience.
This could make sense when only thinking about receiving messages with attachments where you actually need to download the attachment before you can get it as byte array or image.

But for consistency and also because you could probably have created the attachment locally for sending, I'd vote for deprecating the `downloadAs...` methods and instead provide just `as...` methods like on `Icon`.

That `FileContainer` and `Icon` return `CompletableFuture<InputStream>` does not make much sense imho, as a stream is a stream and should simply not deliver bytes if none are available and not be wrapped in a completable future until the file is completely put to memory which is a bit counter-productive in regards to streams.

For `FileContainer` this would easily be changeable as it is a core class, but unfortunately for `Icon` it will be a breaking change without much chance to deprecate and replace. Yet it makes more more sense imho to return an `InputStream`. What do you think @Bastian?

Bastian:
I'm not quite sure about the internals of `InputStream`s.
https://github.com/Javacord/Javacord/blob/master/javacord-core/src/main/java/org/javacord/core/util/FileContainer.java#L268
Not sure if this line is already blocking. The JavaDocs are contradictory: https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#openConnection()

> It should be noted that a URLConnection instance does not establish the actual network connection on creation. This will happen only when calling URLConnection.connect().

This says no, but there's also:
> Throws: IOException - if an I/O exception occurs.

How can it throw an `IOException` without any I/O?

Vampire:
It only talks about "network connection".
Not every I/O operation needs network connection, like file operations for example.
Also a `MalformedURLException` is an `IOException` that you for example get with an `http` URL that contains a newline character or a `jar` URL that does not contain a `!/` separator. Those you will get on `openConnection` already.
Also the `getInputStream()` will not block until the download is done, that's why it is a stream. It will block until the network connection is established though of course.
I sent #447 that changes the `FileContainer` accordingly and #448 which breakingly changes the signature of the `Icon` method.

Bastian:
I've merged #447 which changes the return type of `FileContainer#asInputStream()`.

Can you please rebase on the latest version of the `development` branch and change the return type of `MessageAttachment#downloadAsInputStream()` to return the input stream directly :-)

Also sorry for the late response...

Vampire:
What do you think about my suggested naming change Bastian?

Bastian:
Thanks for reminding me. :)
I will create an issue for it.

Bastian:
https://github.com/Javacord/Javacord/issues/552

PR #441, opened 2019-02-25
Added ban request methods to ServerMemberBanEvent

burdoto:
This change was [suggested on the Discord server](https://discordapp.com/channels/151037561152733184/151326093482262528/549020384343752725).

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
+++ b/javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
@@ -1,7 +1,29 @@
 package org.javacord.api.event.server.member;
 
+import org.javacord.api.entity.server.Ban;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
 /**
  * A server member ban event.
  */
 public interface ServerMemberBanEvent extends ServerMemberEvent {
+
+    /**
+     * Requests the Ban object of the event.
+     *
+     * @return The ban object.
+     */
+    CompletableFuture<Optional<Ban>> requestBan();
+
+    /**
+     * Requests the reason of the ban.
+     *
+     * @return The reason of the ban.
+     */
+    default CompletableFuture<Optional<String>> requestReason() {
+        return requestBan().thenApply(ban -> ban.flatMap(Ban::getReason));
+    }
+
 }
--- a/javacord-core/src/main/java/org/javacord/core/event/server/member/ServerMemberBanEventImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/event/server/member/ServerMemberBanEventImpl.java
@@ -1,8 +1,12 @@
 package org.javacord.core.event.server.member;
 
+import org.javacord.api.entity.server.Ban;
 import org.javacord.api.entity.server.Server;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.event.server.member.ServerMemberBanEvent;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * The implementation of {@link ServerMemberBanEvent}.
@@ -19,4 +23,11 @@ public class ServerMemberBanEventImpl extends ServerMemberEventImpl implements S
         super(server, user);
     }
 
+    @Override
+    public CompletableFuture<Optional<Ban>> requestBan() {
+        return getServer().getBans()
+                .thenApply(bans -> bans.stream()
+                        .filter(ban -> ban.getUser().equals(getUser()))
+                        .findAny());
+    }
 }


javacord-core/src/main/java/org/javacord/core/event/server/member/ServerMemberBanEventImpl.java
```diff
@@ -19,4 +23,14 @@ public ServerMemberBanEventImpl(Server server, User user) {
         super(server, user);
     }
 
+    @Override
+    public CompletableFuture<Ban> requestBan() {
+        return getServer().getBans()
```
burdoto: Proposal:
Fields of all classes that are being extended somewhere should be marked `protected` instead of `private`, in order for extending classes to be able to access the fields directly.

javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
```diff
@@ -1,7 +1,35 @@
...
+
 /**
  * A server member ban event.
  */
 public interface ServerMemberBanEvent extends ServerMemberEvent {
+
+    /**
+     * Requests the Ban object of the event.
+     * The returned future will complete exceptionally with a
+     * {@link NoSuchElementException} if the ban could not be requested.
```
Vampire: Should be `CompletableFuture<Optional<Ban>>` then

javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
```diff
@@ -1,7 +1,35 @@
...
+     *
+     * @return The ban object.
+     */
+    CompletableFuture<Ban> requestBan();
+
+    /**
+     * Requests the reason of the ban.
+     * The returned future will complete exceptionally with a {@link NoSuchElementException}
```
Vampire: Should be `CompletableFuture<Optional<String>>` then

javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
burdoto: Haven't we had some discussion before, that returning `CompletableFuture<Optional<?>>` is bad?

javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
burdoto: See other discussion

javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMemberBanEvent.java
Vampire: Not that I remember, or maybe it was just your opinion?
We have it at other places too.

burdoto: If we got this in other places, its better to have it here, too, of course.
Maybe I talked about this in another context. :)

javacord-core/src/main/java/org/javacord/core/event/server/member/ServerMemberBanEventImpl.java
Vampire: What would be the great benefit?
The JIT compiler should probably inline those calls anyway, shouldn't it?

burdoto: From what I know, it does not; I read about this kind of methods stacking registry entries. However, _this is not_ information that I am 100% sure about.

burdoto: Related:
- https://stackoverflow.com/questions/15896152/java-protected-getters-vs-nested-classes
- http://tutorials.jenkov.com/java/access-modifiers.html
- https://dev.to/scottshipp/avoid-getters-and-setters-whenever-possible-c8m

Vampire: Why do you think it does not?
It is the job of the JIT compiler to optimize code that is often called.
If this method gets called once a day, there is no real performance issue and if this called once a second, I'd epxect the JIT compiler to optimize the call by inlining the access.

burdoto: As I said, I don't really know.
Plus, It is also something that I would say it's nicer than methods, because that would show that it's just fields, better style imo.
After all, it's just a proposal

Vampire: Well, they are properties of the class and it is quite common to encapsulate these.
We could think about exposing them to subclasses, but that is up to Bastian to decide I'd say.

Bastian:
Lgtm 👍

PR #443, opened 2019-03-03
Added Message#deleteAfter methods

burdoto:
This change was proposed in the [Discord channel](https://discordapp.com/channels/151037561152733184/151326093482262528/551828037612601370).

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -27,6 +27,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -50,6 +52,35 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
      */
     default MessageBuilder toMessageBuilder() {
         return MessageBuilder.fromMessage(this);
+    }
+
+    /**
+     * Deletes the message after the given delay.
+     *
+     * @param time The time to be deleted after.
+     * @param unit The unit to measure the given {@code time} in.
+     * @return A future that completes once the message is deleted.
+     */
+    default ScheduledFuture<Void> deleteAfter(long time, TimeUnit unit) {
+        return deleteAfter(time, unit, null);
+    }
+
+    /**
+     * Deletes the message after the given delay with the given reason.
+     *
+     * @param time The time to be deleted after.
+     * @param unit The unit to measure the given {@code time} in.
+     * @param reason The reason why the message has been deleted.
+     * @return A future that completes once the message is deleted.
+     */
+    default ScheduledFuture<Void> deleteAfter(long time, TimeUnit unit, String reason) {
+        return getApi()
+                .getThreadPool()
+                .getScheduler()
+                .schedule(() -> {
+                    this.delete(reason);
+                    return null;
+                }, time, unit);
     }
 
     /**


burdoto:
The errors shown by CircleCI are not caused by this commit.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -102,6 +104,41 @@ default MessageBuilder toMessageBuilder() {
...
+     * @return A future that completes once the message is deleted.
+     */
+    default CompletableFuture<Void> deleteAfter(long time, TimeUnit unit, String reason) {
+        CompletableFuture<Void> future = new CompletableFuture<>();
+
+        try {
+            getApi().getThreadPool()
+                    .getScheduler()
+                    .schedule(() -> this.delete(reason), time, unit)
+                    .get();
```
Vampire: You shouldn't block the caller thread but properly use completable futures.

burdoto: yes, i noticed that on  the last commit but i didnt have time to fix that yet

burdoto: Fixed 👍

Vampire: I wouldn't call that fixed, it is like if you call `invokeLater` just because you have strange behavior in Swing or like calling `Thread.sleep` just because a multi-threaded program behaves strange. Wrapping this stuff in yet another thread is no just a workaround, not a proper solution.

Besides that, you also ignore any exceptions happening in the `delete`.

burdoto: I cannot properly handle checked exceptions in SupplyAsync without rethrowing them underlying a runtime exception..... What do you propose I do

Vampire: There is no need for checked exceptions?

Vampire: And I don't see any supply sync

burdoto: How else do I wait for the ScheduledFuture to be completed after deletion than using #get? I could return the ScheduledFuture instead of the CompletableFuture, that might be even better.

Vampire: The better question is, why should you wait? It is an asynchronous process.

Saladoc: Because `ScheduledFuture` offers no callbacks and the least disgusting way (that I'm aware of) of getting from a `ScheduledFuture` to a `CompletableFuture` is wrapping the former in the latter via `CompletableFuture#supplyAsync` or similar means.

burdoto: I'm going with just returning the `ScheduledFuture` as that is the most logical thing to do with a scheduled task.

burdoto:
Ready for new revision 👍

Vampire:
There is still no way to handle exceptions during deleting. Besides that I'm not sure whether we should mix the kind of futures we return in the API.

burdoto:
What do you suggest how to do exception handling?
- I cannot wait for the ScheduledFuture to finish / exceptionally, because that would mean using `#get` which throws checked exceptions
- The checked exceptions cannot be handled properly within a `CompletableFuture#supplyAsync` block without being wrapped in an unchecked exception
- [What Saladoc said](https://github.com/Javacord/Javacord/pull/443#discussion_r262020344)

Vampire:
Again, why should you wait?
Just react like usual.
Complete the future when the delete is done, complete it exceptionally when it completed exceptionally, cancel the scheduled future when the returned future is cancelled.

Saladoc:
Ooooh now I finally get it. Like this... 

```java
CompletableFuture<Void> future = new CompletableFuture<>();
scheduler.schedule( () -> message.delete().thenRun(future::complete).exceptionally(future::completeExceptionally), time, timeUnit);
return future;
```

Now excuse me I'm gonna facepalm for not having understood what Vampire meant earlier.

Vampire:
Something like that, yes. Plus the cancel logic.

burdoto:
I'm sorry but I do not see _any_ advantage in basically copying the `ScheduledFuture`, which _already is_ linked to the scheduled task, into another `CompletableFuture` object, other than that would not mix up future return types in the class, which _personally_, I would deem as a price low enough to be paid.

However, I will make the changes in a second commit and wait for @Bastian's opinion on this.

Bastian:
Before you invest more time into this PR: I'm not a big fan of the idea.

It exists in JDA, because something like this is trivial to implement with their RestActions. In our case however, we only have three options:
• Implement a `xyzAfter(...)` variant for some methods where we think it makes sense
• Implement a `xyzAfter(...)` variant for every method
• Do not implement it

I greatly dislike the second option, as it would be a lot of additional work. I'm also not a huge fan of the first option, as this could make sense for almost any other method as well, depending on the use case (`sendMessageAfter(...)`, `deleteChannelAfter(...)`, `updateNicknameAfter(...)`, `unbanUserAfter(...)`, ... just to name a few). In the end we would either end with option 2 or get questions like "Why is there `xyzAfter(...)` but not `abcAfter(...)`" regularly.

In the end, it's best to not add that much boilerplate and/or inconsistency for a feature that will be hardly used and can be done with just two additional lines by the user itself:
```java
api.getThreadPool().getScheduler().schedule(() -> {
  // Delayed code goes here
}, 15, TimeUnit.SECONDS);
```
If a user needs it very frequently, he could even write a utility method, that makes it even shorter (0 additional lines, just a few characters):
```java
FutureUtil.delay(() -> message.delete(), 15, TimeUnit.SECONDS)
    .thenApply(() -> System.out.println("Deleted!")); // The "delay" method could return a Future itself
```
(Disclaimer: Such a method would be out of scope of Javacord, just in case someone wants to PR it)

burdoto:
That is a fair opinion. _Closed_

PR #445, opened 2019-03-03
Added static method to build Permissions from permission Bitmasks

burdoto:
This change was proposed in the [Discord channel](https://discordapp.com/channels/151037561152733184/151326093482262528/551843038947966986).

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/permission/Permissions.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/permission/Permissions.java
@@ -82,4 +82,44 @@ public interface Permissions {
      */
     boolean isEmpty(); // We could check it in a default method, but it's faster to just check it in the implementation
 
+    /**
+     * Creates a {@code Permissions} object from the given bitmask.
+     * Permissions that are not included are marked as {@link PermissionState#UNSET}.
+     *
+     * @param bitmask The bitmask of allowed permissions.
+     * @return A {@code Permissions} object created from the given bitmask.
+     */
+    static Permissions fromBitmask(int bitmask) {
+        return fromBitmask(bitmask, 0);
+    }
+
+    /**
+     * Creates a {@code Permissions} object from the given bitmasks.
+     *
+     * <p>Permissions that are not included in any bitmap are marked as {@link PermissionState#UNSET}.
+     * If a permission is included in both the allowed and denied bitmap,
+     * it is marked as {@link PermissionState#UNSET}.
+     *
+     * @param allowedBitmask The bitmask of allowed permissions.
+     * @param deniedBitmask The bitmask of denied permissions.
+     * @return A {@code Permissions} object created from the given bitmasks.
+     */
+    static Permissions fromBitmask(int allowedBitmask, int deniedBitmask) {
+        PermissionsBuilder permissionsBuilder = new PermissionsBuilder();
+
+        for (PermissionType permissionType : PermissionType.values()) {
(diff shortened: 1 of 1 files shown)


Saladoc:
I think this would better be suited as a `Permissions#fromBitmask(int)` method.

burdoto:
Moved the method to `Permissions` 👍

Saladoc:
You could have made it a static constructor method directly creating a `Permissions` object... Which was the whole point of me suggesting it.

burdoto:
I'm not sure if that is better than returning a plain `Set`, as a `Permissions` object is more useful for a Collective of allowed, neutral and denied permissions.
However if that is requested, I can change it to be such a method.

Saladoc:
When someone has a bitmask, where do they want to go from there? They will, in 99% of use cases, want to arrive at a `Permissions` object. So that's what should be provided. You can even extend the signature to include a denied bitmask as well, with 0 as a default parameter.

burdoto:
That is a good idea. I will make those changes asap 👍

burdoto:
Ready for new revision 👍

javacord-api/src/main/java/org/javacord/api/entity/permission/Permissions.java
```diff
@@ -82,4 +82,43 @@ default PermissionsBuilder toBuilder() {
      */
     boolean isEmpty(); // We could check it in a default method, but it's faster to just check it in the implementation
 
+    /**
+     * Creates a {@code Permissions} object from the given bitmask.
+     * Permissions that are not included are marked as {@link PermissionState#UNSET}.
+     *
+     * @param bitmask The bitmask of allowed permissions.
+     * @return A {@code Permissions} objects created after the given bitmask.
```
Bastian:
```suggestion
     * @return A {@code Permissions} object created after the given bitmask.
```
Just a small typo

javacord-api/src/main/java/org/javacord/api/entity/permission/Permissions.java
```diff
@@ -82,4 +82,43 @@ default PermissionsBuilder toBuilder() {
...
+
+    /**
+     * Creates a {@code Permissions} object from the given bitmasks.
+     * Permissions that are not included in any bitmap are marked as {@link PermissionState#UNSET}.
+     * If a permission is included in both the allowed and denied bitmap,
+     * it is marked as {@link PermissionState#UNSET}.
+     *
+     * @param allowedBitmask The bitmask of allowed permissions.
+     * @param deniedBitmask The bitmask of denied permissions.
+     * @return A {@code Permissions} objects created after the given bitmasks.
```
Bastian:
```suggestion
     * @return A {@code Permissions} object created after the given bitmasks.
```

javacord-api/src/main/java/org/javacord/api/entity/permission/Permissions.java
```diff
@@ -82,4 +82,43 @@ default PermissionsBuilder toBuilder() {
...
+     *
+     * @param bitmask The bitmask of allowed permissions.
+     * @return A {@code Permissions} objects created after the given bitmask.
+     */
+    static Permissions fromBitmask(int bitmask) {
+        return fromBitmask(bitmask, 0);
+    }
+
+    /**
+     * Creates a {@code Permissions} object from the given bitmasks.
```
Bastian:
```suggestion
     * Creates a {@code Permissions} object from the given bitmasks.
     * <p>
```

burdoto:
All changes have been applied 👍

Bastian:
Thx

PR #451, opened 2019-03-17
Add tts flag for received messages

Saladoc:
- tts Status can't be edited
 - sending tts messages already is possible so no need for change there.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -724,6 +724,13 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     boolean isPinned();
 
     /**
+     * Checks if this message is text-to-speech.
+     *
+     * @return Whether this message is text-to-speech or not.
+     */
+    boolean isTts();
+
+    /**
      * Checks if the message mentions everyone.
      *
      * @return Whether the message mentions everyone or not.
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
@@ -69,6 +69,11 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
     private volatile boolean pinned;
 
     /**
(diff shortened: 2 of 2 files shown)


Bastian:
Is `tts` a well-known acronym?
I think, I would prefer `isTextToSpeech()`. What do you think?

Also notice, that we have the following method ordering convention:
- Overridden method must be in the same order than it is in the interface
- Overridden method must not be separated by non-overridden methods between them

Saladoc:
`tts` is consistent with both the client command and the `MessageBuilder` class. I fixed the method order.

Bastian:
Agreed. Maybe we should consider renaming the method in the `MessageBuilder`, but this can be done in a separate commit, if we decide so.

xaanit:
`tts` is a very well known acronym. Just googling it you'll see all sorts of "text to speech". The client itself is inconsistent. The command is tts while the settings is Text-To-Speech, so technically either are "correct". But tts is shorter while losing no information. Renaming it is just adding unnecessary verbosity.

Bastian:
I'm not sure if I would agree. Of course you will get a lot of results with `text to speech` when you google for `tts`, but this does not mean that everyone knows what it means. 

I'm personally not a big fan of acronyms in method/class names, unless the acronym is better known than the actual full word (e.g. `http`, `url`, ...).
Someone who does not know `tts` will be confused when seeing it. Someone who does know `tts` will not be confused by `text to speed` though. Even though I know the acronym, I still find
```java
if (message.isTextToSpeed()) {
  // Do something
}
```
more pleasing to read than
```java
if (message.isTts()) {
  // Do something
}
```
but this is more of a personal preference.

xaanit:
"Someone who does not know `tts` will be confused when seeing it."

Someone who doesn't know `pvp` will be confused when they start up a game, despite it being very well known in general. "If someone doesn't know it it'll be confusing" is somewhat of a redundant argument. If someone doesn't know something of course they'll be confused, that's kind of implied. But it's a well known enough acronym that Discord thought it'd be good enough and you see it even when googling "text to speech". Both `http` and `url` are very specific to programming (i.e they're a more rare outside of programming talk). 

I wouldn't say it's "more" known but I'd say it's on par.

PR #436, opened 2019-02-03
[WIP] Performance improvements

Bastian:
There are some old code parts, that have a very bad performance (e.g. by too excessive use of streams). This PR tries to clean them up.

PR #455, opened 2019-03-23
Improve performance of getters for channels

Bastian:
This PR is a cherry-pick of https://github.com/Javacord/Javacord/pull/436, as this PR became a bit of a mess with several other changes, too.

This PR mainly improves the performance of the `getChannelXyzById(...)` methods. As these are used internally by the packet handlers, improving their performance has a huge impact on Javacord#s overall performance.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
@@ -13,7 +13,7 @@ import org.javacord.api.Javacord;
 import org.javacord.api.entity.ApplicationInfo;
 import org.javacord.api.entity.activity.Activity;
 import org.javacord.api.entity.activity.ActivityType;
-import org.javacord.api.entity.channel.GroupChannel;
+import org.javacord.api.entity.channel.Channel;
 import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.emoji.CustomEmoji;
 import org.javacord.api.entity.emoji.KnownCustomEmoji;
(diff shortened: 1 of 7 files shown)


javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelDeleteHandler.java
```diff
@@ -42,6 +42,7 @@ public ChannelDeleteHandler(DiscordApi api) {
     @Override
     public void handle(JsonNode packet) {
         int type = packet.get("type").asInt();
+        api.removeChannelFromCache(packet.get("id").asLong());
```
Vampire: Maybe better remove from the cache after dispatching the event?

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -571,10 +571,10 @@ public void purgeCache() {
                 .map(Cleanupable.class::cast)
                 .forEach(Cleanupable::cleanup);
         servers.clear();
-        groupChannels.values().stream()
+        channels.values().stream()
+                .filter(Cleanupable.class::isInstance)
                 .map(Cleanupable.class::cast)
                 .forEach(Cleanupable::cleanup);
-        groupChannels.clear();
```
Vampire: Shouldn't you add `channels.clear();`?

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -663,9 +663,9 @@ public void addUserToCache(User user) {
      *
```
Vampire: JavaDoc is wrong now

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -675,9 +675,11 @@ public void addGroupChannelToCache(GroupChannel channel) {
...
```
Vampire: JavaDoc is wrong now

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelDeleteHandler.java
Bastian: Events are dispatched in another thread anyway, so this does not matter.
But you are correct: Removing it before calling the methods will break the `handleGroupChannel()` because it uses calls `api.getGroupChannelById(channelId)`.
Fixed :)

PR #461, opened 2019-03-30
Quickfix for new channel types

Bastian:
This commit only adds very basic support for the new channel types.
News channels get treated as normal server text channels and store channels get completely ignored.

It also cleans up the switch-statements a little bit by using the enum instead of hard-coding the integer values.

Better support will be added when we get more information from Discord in https://github.com/discordapp/discord-api-docs/issues/881.


Bastian:
I don't think that's necessary, as this is only a temporary solution.
The worst that can happen is that the CompletableFuture fails exceptionally, which you should always be prepared for anyway.

PR #472, opened 2019-04-26
Add the ability to move AudioConnections between voice channels

### LINKED ISSUE #470: Add method to move an audio connection to another voice channel
It should be possible to move an audio connection to another voice channel on the same server.
```java
AudioConnection connection = ...;
connection.moveTo(otherChannel); // Moves the bot to another channel. Throws an IllegalArgumentException when the channel is from another server
otherChannel.connect(); // Closes any potential existing audio connections on the server and creates a new one. It does not reuse the existing one
```

Mysterypotatoguy:
Closes #470

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
@@ -41,7 +41,7 @@ public class AudioConnectionImpl implements AudioConnection, InternalAudioConnec
     /**
      * The voice channel of the audio connection.
      */
-    private final ServerVoiceChannel channel;
+    private ServerVoiceChannel channel;
 
     /**
      * A lock to ensure we don't accidentally poll two elements from the queue.
@@ -52,6 +52,11 @@ public class AudioConnectionImpl implements AudioConnection, InternalAudioConnec
      * A future that finishes once the connection is fully established.
      */
     private final CompletableFuture<AudioConnection> readyFuture;
+
+    /**
+     * A future that finishes once the connection has been moved to a different channel.
+     */
+    private CompletableFuture<Void> movingFuture;
 
     /**
      * The source that gets played at the moment.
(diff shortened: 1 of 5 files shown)


javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -41,7 +41,7 @@
     /**
      * The voice channel of the audio connection.
      */
-    private final ServerVoiceChannel channel;
+    private ServerVoiceChannel channel;
```
Bastian: This should be marked as `volatile` then.

javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -265,6 +283,29 @@ public boolean dequeue(AudioSource source) {
...
+        movingFuture = new CompletableFuture<>();
+        if (!destChannel.getServer().equals(channel.getServer())) {
+            movingFuture.completeExceptionally(
+                    new IllegalArgumentException("Cannot move to a voice channel not in the same server!"));
+            return movingFuture;
+        }
+        if (destChannel.equals(channel)) {
+            movingFuture.complete(null);
+        } else {
+            movingFuture.thenRun(() -> setChannel(destChannel));
```
Bastian: You should return the result of `thenRun()`, otherwise it's not guaranteed to be executed before any user-provided callbacks.

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerVoiceChannelImpl.java
```diff
@@ -111,6 +111,7 @@ public void removeConnectedUser(long userId) {
 
     @Override
     public CompletableFuture<AudioConnection> connect() {
+        getServer().getAudioConnection().ifPresent(AudioConnection::close);
```
Bastian: Does this work every time?
I can imagine that it might cause race conditions when the new connection gets established before the old one got closed. Or does Discord closes an existing connection automatically when creating a new one?

Bastian:
Any update here? :)

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerVoiceChannelImpl.java
Bastian: If this is what's blocking this PR from being finished, feel free to ignore it for now.
We can create an issue for it and fix it in another commit.

Mysterypotatoguy:
Updated with requests :)

Bastian:
Lgtm 👍

PR #473, opened 2019-04-27
Remove the requirement for AudioSourceBase.

Saladoc:
The AudioUdpSocket will assume that each AudioSource manages their own directly attached listeners, making it more robust towards custom Implementations. The AudioSourceBase class remains as a convenient utility for implementors to depend on and a limited-usecase implementation of a ListenerManager removes the need to depend on javacord-core.

I have tested this with a slightly altered version of Bastians Lavaplayer-Adapter and verified that the `AudioSourceFinishedListener` gets called. Further testing is pending, but everything looks well enough that I feel comfortable putting this up for review.


Bastian:
I get why casting to `AudioSourceBase` internally feels wrong, but I really feel like this PR is moving in the wrong direction. 

It's really hard for me to come up with any good reason why anyone would want to implement everything - especially the listener logic - themselves. The only "reason" would be some abuse like an AudioSource queue that is itself just an AudioSource.

However, I can easily come up with a lot of downsides of custom implementations:
- It's extremely likely for users to mess something up
- We limit ourself with future changes. Adding any new non-default method to the `AudioSource` interface will be a breaking change, as it breaks all existing AudioSource implementations. Hell, we can't even change related interfaces like the `ListenerManager` anymore without having a breaking change.
- It encourages bad designs like the mentioned queue


Instead of removing the need to extend `AudioSourceBase`, I think it would be better to just make it more explicit by throwing an `InvalidArgumentException` when setting an audio source that does not extend `AudioSourceBase` instead of just letting it fail somewhere internally with a `ClassCastException`. There are just too many downsides with custom implementations but close to zero benefits.

Vampire:
I didn't look at the code changes, but my standpoint on the philosophical side is:
1. If the API demands to get an `AudioSource`, be able to cope with not getting an `AudioSourceBase`
2. If you want to require people to use `AudioSourceBase` as implementation, then require it in the API.
3. If you are totally against 1. and 2., then throw an exception immediately in the setter for fail-fast behavior.

I personally greatly dislike 3. as it lies to the user in saying "I accept any `AudioSource`" but then actually requires an `AudioSourceBase` and I don't like that.

If you go with 2. you can still have the `AudioSource` interface, so that a user can declare his variables as `AudioSource` and use only that interface, but the API already makes sure you only get `AudioSourceBase` implementations supplied.

Bastian:
I just don't see how 2. should work without having the "normal" user of audio sources need to deal with the `AudioSourceBase` as it is not an interface but a class.
This means that a structure like 
Prefered
this will not work but only
Reality
which will mean that a user cannot just do
```java
AudioConnection connection = ...;
AudioSource source = new YoutubeAudioSource(...);
connection.queue(source);
```
but is limited to
```java
AudioConnection connection = ...;
YoutubeAudioSource source = new YoutubeAudioSource(...);
connection.queue(source);
```
or worse
```java
AudioConnection connection = ...;
AudioSourceBase source = new YoutubeAudioSource(...);
connection.queue(source);
```

Option 1. comes with the issues I described in my previous message.

Another option is to get rid of the `AudioSource` interface and rename `AudioSourceBase` to `AudioSource`. This will however come with all the restrictions of classes.
E.g. `PausableAudioSource extends AudioSource` will no longer be possible which comes with a lot of other problems and probably break our complete design.

You should always remember that most people will never write their own audio sources. This means that nearly all audio sources will be provided by us or a few amount of experienced contributors. With this in mind, I still believe that option 3 is the best option.

Vampire:
And why can't we just check whether it is an `AudioSourceBase` before we cast to it and handle it differently if it is not an `AudioSourceBase`?

Bastian:
> Instead of removing the need to extend AudioSourceBase, I think it would be better to just make it more explicit by throwing an InvalidArgumentException when setting an audio source that does not extend AudioSourceBase instead of just letting it fail somewhere internally with a ClassCastException.
That's exactly what I have suggested.

Saladoc:
The whole point of the PR is us talking in Discord about whether it is possible for `AudioSource` to be implemented without depending on a delegate introduced in `javacord-core`. Bastian told me he was unsure if it could be done and I was welcome to give it a try, so I did. He told me to feel free to PR if I manage to do it, so I did.
I do not think this code is beautiful, technically it could lose a lot of fat since there is only one type of listener that can be attached to an `AudioSource` it technically could be coded with that assumption since any event addition would warrant a new API version anyway. But it is proof that a delegate is not needed.

The point is: I'm generally unhappy with assumptions being made about class hierarchies being made that are not directly apparent from the interfaces used. However, in many cases that behavior is justifiable as it can be reasonably assumed that everything passed into that class is created by javacord itself and therefore we control the ~~universe~~ class hierarchy. 
However, with Audio Sources it is a different thing. Audio sources *can* be created by third party libraries and even end users so it would arguably be bad api design if there were any requirements to these implementations that are not immediatly apparent from the required interface (`AudioSource` in this case) - but that Vampire already pointed out above.

The restrictions we had with `DiscordEntity`s that *forced* us to work with delegates and a `(type,id)->listeners` mapping instead of storing the listeners at the entities do not apply here. It is safe to assume the instances will stay valid over the whole runtime. How we proceed from here is not dictated by technical necessity but a concious API design decision and this is merely proof that "Plan B" can work.

Bastian:
I'm sorry, but I am not going to merge this PR.
There are just too many downsides and imo zero benefits in allowing audio sources that are not extending the `AudioSourceBase` class. I see why it feels "wrong" to enforce this restriction but in practice this will not be a problem and I'm willing to accept this small theoretical imperfection in the api.

PR #476, opened 2019-05-31
Added Messageable#getChannel to get the channel that the messageable will send to

burdoto:
This change is suggested by me.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Messageable.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Messageable.java
@@ -1,6 +1,8 @@
 package org.javacord.api.entity.message;
 
+import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.message.embed.EmbedBuilder;
+import org.javacord.api.entity.user.User;
 
 import java.io.File;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


Vampire:
What would be the use-case for this method?

burdoto:
In a case, where there is only a `Messageable` passed around, e.g. for Message Utilities, to recieve the actual channel that will be afflicted by the `Messageable`.

Reference: https://github.com/burdoto/javacord-dialogue/blob/master/src/main/java/de/kaleidox/javacord/dialogue/model/SingleInputAction.java#L20

Vampire:
But what for? 
If you need a channel, pass around a channel.
A messageable does not need to have a channel.
The property of a messageable is, that you can send it a message, no matter whether there is a channel underneath or not.
That the two messageable that exist currently have an underlying channel does not necessarily mean all future ones will have one or that it should be enforced via the interface imho. That would be tight coupling that should not be done.

Vampire:
Btw. **if** this will get accepted, please move the implementation to the actual implementations instead of doing `instanceOf` checks in a default method.

Vampire:
And also the JavaDoc has a typo and should not have information about the implementation details.

burdoto:
The main goal with having it like this is to make it simpler to get the Channel ID that a Messageable is "attached" to.

I'll leave the decision up to @Bastian, if this is not aimed for, just close the PR.

Vampire:
Well, as I said, the messageable interface does not and should not have any knowledge about channels. But sure, let's Bastian decide, I stated my PoV.

Saladoc:
A `Messageable` is something that a message can be sent to. No more, no less. If you pass one around, all the information you give is "you can send messages to this". While it is true that it is at the moment de facto limited to something backed by channels, but that should not be an assumption made, especially not by the interface itself. It would be perfectly acceptable to have a Webhook client class implement `Messageable` as well.
Inheritance hierarchies should be a one-way street. The dependency goes from `User` and `TextChannel` to `Messageable`, so `Messageable` should not have any knowledge about them. 

And finally this specific use case can easily be resolved in your library, I'm not at all convinced that there is a general use case here.

PR #477, opened 2019-06-03
Added a method for getting a message based on a channel and a message ID

burdoto:


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -722,6 +723,25 @@ default AccountUpdater createAccountUpdater() {
...
+     * @return The message from those IDs.
+     */
+    default CompletableFuture<Message> getMessageById(long messageId, long channelId) {
+        return getChannelById(channelId)
+                .flatMap(Channel::asTextChannel)
+                .map(textChannel -> getMessageById(messageId, textChannel))
+                .orElseGet(() -> {
+                    CompletableFuture<Message> failingFuture = new CompletableFuture<>();
+                    failingFuture.completeExceptionally(
+                            new NoSuchElementException("No channel with id " + channelId + " found"));
```
Vampire: We don't throw `NoSuchElementException`s, I guess if this will get added, it should return `CompletableFuture<Optional<Message>>`, probably something like
```java
        return getChannelById(channelId)
                .flatMap(Channel::asTextChannel)
                .map(textChannel -> textChannel.getMessageById(messageId));
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -722,6 +723,25 @@ default AccountUpdater createAccountUpdater() {
         }
     }
 
+    /**
+     * Gets a message by a channel and message id.
+     *
+     * @param messageId The id of the message.
+     * @param channelId The id of the channel.
+     * @return The message from those IDs.
```
Vampire:
```suggestion
     * @return The message with the given id from the text channel with the given id.
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
burdoto: This would instead result in an `Optional<CompletableFuture<Message>>`, and I have no idea how to turn these generics around.

Is this kind acceptable, though?

Vampire: Ah, your right, I wrote that from mind.
I'll leave this decision to @Bastian, but actually we already have an according return type for `org.javacord.api.entity.message.MessageAuthor#asWebhook` where it is `Optional<CompletableFuture<Webhook>>`.

The other way around would then probably be something like
```java
        Optional<TextChannel> textChannel = getTextChannelById(channelId);
        if (textChannel.isPresent()) {
            return textChannel.get()
                    .getMessageById(messageId)
                    .thenApply(Optional::of);
        } else {
            return CompletableFuture.completedFuture(Optional.empty());
        }
```

Vampire:
You also miss the `String`s variant of the method where you give both IDs as `String`.

Bastian:
Neither throwing a `NoSuchElementException` nor returning a `Optional<CompletableFuture<Message>>` would be the default "Javacord style". We usually just send a request to Discord and let Discord do the work. This is also possible here:
Instead of forwarding `DiscordApi#getMessageById(long, long)` to `TextChannel#getMessageById(long)`, we can do it the other way around:
- Move the code from [InternalTextChannel#getMessageById(...)](https://github.com/Javacord/Javacord/blob/master/javacord-core/src/main/java/org/javacord/core/entity/channel/InternalTextChannel.java#L55-L59) to `DiscordApi#getMessageById(long, long)`.
- Now make `TextChannel#getMessage(long)` a default method that calls the method in `DiscordApi`

It's now Discord that handled the response (not found, no permissions, ...) and we can just return a `CompletableFuture<Message>`.

Only problem: What happens when someone requests a message from another shard? `DiscordApiImpl#getOrCreateMessage(TextChannel, JsonNode)` needs a text channel as the first parameter. This most likely the place where we can throw the exception:
```java
public CompletableFuture<Message> getMessageById(long messageId, long channelId) {
    return getApi().getCachedMessageById(id)
            .map(CompletableFuture::completedFuture)
            .orElseGet(() -> new RestRequest<Message>(getApi(), RestMethod.GET, RestEndpoint.MESSAGE)
            .setUrlParameters(Long.toUnsignedString(channelId), Long.toUnsignedString(messageId))
            .execute(result -> ((DiscordApiImpl) getApi()).getOrCreateMessage(getTextChannelById(channelId).orElseThrow(() -> new IllegalStateException("Message from unknown text channel. Does it belong to another shard?")), result.getJsonBody())));
}
```
Something like this.

Vampire:
> In the long run, we might want to support messages from different shards.

It would also be nice to support servers from other shards.
Currently it is e. g. pretty hard to get the amount of servers you are in for example to send this to some statistics server or bot list service.
Or it is pretty hard to impossible cleanly to get a command in a private message and affect a server if the server is on another shard.

> Move the code from InternalTextChannel#getMessageById(...) to DiscordApi#getMessageById(long, long).

This is not possible, that would use core-code in the api-artifact, it probably needs to be in the `DiscordApiImpl` or so.

Bastian:
> This is not possible, that would use core-code in the api-artifact, it probably needs to be in the DiscordApiImpl or so.

Ups, sorry. That's what I meant.

> Or it is pretty hard to impossible cleanly to get a command in a private message and affect a server if the server is on another shard.

I don't want a `ShardManager` or a similar solution that other libraries have, because it does no longer work as soon as you have shards on multiple machines. Implementing a local solution is pretty trivial and everyone should be able to develop this on it's own. You only have to keep all your DiscordApi instances somewhere together (e.g. a list), that's all.

Here's our [sharding-example from the wiki](https://javacord.org/wiki/advanced-tutorials/sharding/) extended with the functionality you mentioned. That's ~10 additional lines of code.
```java
public class Main {

    private static final List<DiscordApi> shards = new ArrayList<>();

        new DiscordApiBuilder()
            .setToken("top secret")
            .setRecommendedTotalShards().join()
            .loginAllShards()
            .forEach(shardFuture -> shardFuture
                .thenAccept(Main::onShardLogin)
                .exceptionally(ExceptionLogger.get())
            );
    }

    private static void onShardLogin(DiscordApi api) {
        System.out.println("Logged in Shard " + api.getCurrentShard() + "/" + api.getTotalShards());
        synchronized(shards) {
            shards.add(api);
            if (shards.size() >= api.getTotalShards()) {
                int totalServers = shards.map(shard -> shard.getServers().size()).reduce(0, (acc, amount) -> acc + amount));
                // Send to server list
            }
        }

        api.onMessageCreate(event -> {
            if (!event.isPrivateMessage()) {
                return;
            }
            // You can access the "shards" array here to access something from another shard
        });
    }

}
```

This will fail, as soon as the bot runs on multiple machines. Implementing a network solution is out-of-scope of Javacord and might look different depending on the user's needs. 

Btw: Afaik most server lists allow you to send a request per shard and calculate the total server count themselves.

Vampire:
I didn't suggest a shard manager, neither local nor networked.

Ok, if bot lists have the sharded possibility they were a bad example, then let's say one wants to mention how many servers the bot is in in the bot activity.

Or if you e. g. have a server ID or name and want to do something on that server due to a private message, like a secret vote where none of the voters should see who gave which vote. You would either have to have some way for your distributed shards to talk to each other to be able to get the votes in a DM and then update the vote count in the server on a another shard, or you have to create one vote channel per voter where only the voter has access or something similar, while it would be trivial to do something to the server in question directly. Foreign-shard entities could just never be cached, but always operated on live.

But we are misusing this PR, this should be discussed in a separate issue I'd say.

PR #482, opened 2019-06-16
Implemented a method to get a DiscordEntity of any type by it's respective ID

burdoto:
This change was proposed [by me in the Discord server](https://discordapp.com/channels/151037561152733184/151326211581280256/589761459551076380).

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -1,6 +1,7 @@
 package org.javacord.api;
 
 import org.javacord.api.entity.ApplicationInfo;
+import org.javacord.api.entity.DiscordEntity;
 import org.javacord.api.entity.Icon;
 import org.javacord.api.entity.activity.Activity;
 import org.javacord.api.entity.activity.ActivityType;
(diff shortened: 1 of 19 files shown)


Bastian:
IDs are not unique across different entity types.
E.g. the initial channel has (had?) the same id as the server.
If I remember correctly, the @everyone role also has the same id as the server.

burdoto:
This is only true for the `@everyone` role;
grafik
grafik

And that is a case that can be handled back to default behaviour, as the `@everyone` role is not "classically" accessible via it's ID, as seen in the screenshot.

Bastian:
I've just checked and the `#info` channel on the Javacord server has the same id as the server. This is probably legacy behavior, but it still needs to be considered.

Can you show me a use-case where you need these kind of method? I'm failing to find one, but I guess you have one as you suggested this method?

I'm unsure about the `DiscordEntityType` especially considering that we already have `ChannelType`. It might confuse people if there is `Channel#getType()` and `Channel#getEntityType()`. What's the benefit of having the enum? When I think about it, we should probably even deprecate and remove the `ChannelType` enum. Especially with the new channel types it could lead to errors (e.g. checking if the channel type is `SERVER_TEXT_CHANNEL` and forgetting about news channels). But that's another topic.

burdoto:
Initial motivation was this part of a `response` method to respond to any kind of parameter:
grafik

burdoto:
Handling duplicate types is very niche in my opinion. In a case where there are multiple entities with one ID, like with the legacy channels, we might just print a warning and default to the server object, since that is rather the exact object.

Vampire:
Uniqueness is not guaranteed by Discord over entity types, so they can give duplicate IDs for any entity type any time.
That they are currently almost unique accross entity types is an implementation detail we should imho not depend upon and returning an arbitrary one if multiple match is not a good idea either. The former would be tight coupling to an API that can change any time and break our or our users code, the latter just feels very wrong.

burdoto:
We could also introduce another result object type that will carry over any results of such a search. IMO this would be pretty shabby, though.

Syntax suggestion:
```java
// ByIdSearchResults DiscordApi#getEntitiesById(long);
ByIdSearchResults results = api.getEntitiesById(123L);

// Optional<T> ByIdSearchResults#getType(Class<T extends DiscordEntity>)
Optional<ServerTextChannel> foundServerTextChannelOpt = results.getType(ServerTextChannel.class);
Optional<Role> mightBeEveryoneRole = results.getType(Role.class);

// boolean ByIdSearchResults#hasResult(DiscordEntityType);
boolean hasServerResult = results.hasResult(DiscordEntityType.SERVER);
```

Vampire:
I'm still not sure whether this should be in Javacord core or in some utility library at most.
Is the use-case for this really that common?

burdoto:
The use-case might not be extremely common, but given the existing `ChannelType` enum, it might be better to extend that one to an entity-wide system. Plus, personally I'd say an addition of ~200 effective lines, given we would deprecate the `ChannelType` enum would really not be a bloating change.

Vampire:
An additional comment from xaanit from the DAPI channel also arguing against this PR:
grafik

PR #486, opened 2019-07-02
Role performance & synchronization enhancements

coaxial1:
This PRs fixes some bugs I encountered:

- `RoleImpl#getUsers` does now create a defensive copy (before, it only wrapped it into a unmodifiable view)
- Synchronize access to the `users` hashset of `RoleImpl` using a `ReentrantReadWriteLock` as it caused frequent `ConcurrentModificationException`s for me because there was a change to the set while creating a defensive copy

This PR also aims to further improve the internal handling related to `RoleImpl`, so I added a convenience method `RoleImpl#hasUser` (also added to the public interface) that allows to check whether an user has a role without first creating a defensive copy. 

Additionally, I changed `Server#getRoles(User user)` to use that method instead of using `contains()` to increase performance.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/permission/Role.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/permission/Role.java
@@ -64,6 +64,14 @@ public interface Role extends DiscordEntity, Mentionable, Nameable, Permissionab
     Collection<User> getUsers();
 
     /**
+     * Checks whether the specified users has this role.
+     *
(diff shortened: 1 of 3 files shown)


Bastian:
@coaxial1 Can you please give me write access to your fork? I need it to rebase to the latest version of the `development` branch.

Alternatively, if you don't want to give me access, please rebase the PR yourself :)

Bastian:
Thank you!

PR #487, opened 2019-07-02
Remove unnecessary shard-id based reconnect waiting time

coaxial1:
This PR removes the addtional reconnect wait time per shard ( `+ (currentShard * 6)` ) in the default `reconnectDelayProvider` as it is not necessary.

The goal of it was to delay the reconnect of each shard by 6 seconds if all websockets disconnect simultaneously to prevent hitting the identify ratelimit. However, this causes inconvenient delays if only a single shard with a large shard ID disconnects, e.g. shard 10 - then Javacord is currently waiting 60 unnecessary seconds before even attempting to reconnect.

`+ (currentShard * 6)`  can be removed as `DiscordWebSocketAdapter` prevents hitting the identify rate limit using a Semaphore anyways. Relevant code:

DiscordWebSocketAdapter.java L391
```java
            waitForIdentifyRateLimit(); //Javacord ensures to not hit the limit BEFORE (re)connecting
            websocket.connect();
```

The `waitForIdentifyRateLimit()` method prevents that more than one connect attempt per 5100ms can happen on the same token.

DiscordWebSocketAdapter.java L413
```java
    /**
     * Identification is rate limited to once every 5 seconds,
     * so don't try to more often per account, even in different instances.
     * This method waits for the identification rate limit to be over, then returns.
     */
    private void waitForIdentifyRateLimit() {
        String token = api.getPrefixedToken();
        connectionDelaySemaphorePerAccount.computeIfAbsent(token, key -> new Semaphore(1)).acquireUninterruptibly();
        for (long delay = 5100 - (System.currentTimeMillis() - lastIdentificationPerAccount.getOrDefault(token, 0L));
                delay > 0;
                delay = 5100 - (System.currentTimeMillis() - lastIdentificationPerAccount.getOrDefault(token, 0L))) {
            logger.debug("Delaying connecting by {}ms", delay);
            try {
                Thread.sleep(delay);
            } catch (InterruptedException ignored) { }
        }
    }
```

Thus, we can safely remove the additional waiting time included in the default `reconnectDelayProvider`.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
@@ -444,7 +444,7 @@ public class DiscordApiImpl implements DiscordApi, DispatchQueueSelector {
         this.proxyAuthenticator = proxyAuthenticator;
         this.trustAllCertificates = trustAllCertificates;
         this.reconnectDelayProvider = x ->
-                (int) Math.round(Math.pow(x, 1.5) - (1 / (1 / (0.1 * x) + 1)) * Math.pow(x, 1.5)) + (currentShard * 6);
+                (int) Math.round(Math.pow(x, 1.5) - (1 / (1 / (0.1 * x) + 1)) * Math.pow(x, 1.5));
 
         if ((proxySelector != null) && (proxy != null)) {
             throw new IllegalStateException("proxy and proxySelector must not be configured both");


Bastian:
While `waitForIdentifyRateLimit()` does not work across multiple machines, I do agree that it might be more sensible to not take the shard count into account in the default delay.
If someone runs into issues with a larger bot that runs on multiple machines, it is still possible to adjust the reconnect delay.

Bastian:
Can you please rebase the PR to be up to date with the development branch?
I cannot rebase it myself, as I don't have write permissions on your fork.

coaxial1:
I guess that should've worked.

I'll think about a good distributed solution for `waitForIdentifyRateLimit()` once that becomes a problem for Yunite in the future; maybe it'll possible to just make the implementation of that method replaceable so custom solutions can be implemented.

Bastian:
It would most likely be easier by providing another reconnect delay that gets its delay from a centralized webservice or something similar. At least that's how I would do it.
Another option would be to take the amount of machines into account instead of the amount of shards. This would result in slightly longer reconnect delays, but would work without the need of a web service.

coaxial1:
Yeah, that would also be an option. Whether you add the delay before Javacord starts the reconnect attempt or directly before the websocket connect is only a little difference, right?
However, in my case, a centralized service would be required anyways to forward DMs to the correct machine (depends on which server caused the DM interaction), so that would be a doable addition.

Bastian:
> I guess that should've worked.

Not quite:

Bastian:
```
git checkout remove-unnecessary-wait-time-2 // If not already done
git reset --hard 024b2972892fe4e8696714b66bdb2006309bada5
git remote add https://github.com/Javacord/Javacord.git upstream // If not already done
git pull upstream
git rebase upstream/development
git push -f
```

PR #488, opened 2019-07-04
Added a documentation note for #isMove methods on ServerVoiceChannelEvents

burdoto:


Vampire:
That's the definition of move, user leaves one channel and joins a different one => move. Whether he initiated the move himself or was moved by a different user is not significant for this. Imho this addition is unnecessary.

Bastian:
I have to agree with Vampire: Moving from one channel to another is the same as switching.
Thank you for your contribution anyway :)

PR #491, opened 2019-08-11
Fix speaking update logic, send false speaking update on ready packet

Mysterypotatoguy:
This fixes speaking updates always being sent as 1 (speaking), also fixes the speaking indicator being stuck on at connect.
This is a basic fix and I plan to submit another PR implementing speaking flags properly


Bastian:
Thx :-)

PR #492, opened 2019-08-16
add stream (go live) voice channel permission

coaxial1:
This small PR adds the `STREAM` permission for voice channels that will be rolled out over the next weeks. It's already there on some verified and partnered servers.


Bastian:
Thx :)

PR #496, opened 2019-09-24
Fix cached message retrieval

Vampire:


Bastian:
Some context for this PR:


* The exception:
image

* The code that caused the exception (basically just calling the method): 
image

PR #499, opened 2019-10-26
Implement priority speaker

Mysterypotatoguy:
This PR implements priority speaker for audio connections and adds some basis for other speaking flags (i.e. soundshare which is not currently useful to bots)


Bastian:
Thanks! <3

PR #503, opened 2019-10-30
Replace AudioConnection#queue methods with AudioConnection#set methods

### LINKED ISSUE #469: Replace AudioConnection#queue(...) with AudioConnection#set(...)
Instead of having a primitive built-in queue, it might better to keep it simple and only allow users to set a single audio source. We can still provide a utility class that supports the user with the queue logic.

Mysterypotatoguy:
Resolves #469 

Unsure on the naming of these methods, perhaps they should just be `setAudioSource`, `removeAudioSource` etc..

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
@@ -9,6 +9,7 @@ import org.javacord.api.entity.channel.ServerVoiceChannel;
 import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.entity.server.ServerImpl;
 import org.javacord.core.listener.audio.InternalAudioConnectionAttachableListenerManager;
+import org.javacord.core.util.concurrent.BlockingReference;
 import org.javacord.core.util.gateway.AudioWebSocketAdapter;
 import org.javacord.core.util.logging.LoggerUtil;
 
@@ -16,14 +17,9 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.ReentrantLock;
 
 public class AudioConnectionImpl implements AudioConnection, InternalAudioConnectionAttachableListenerManager {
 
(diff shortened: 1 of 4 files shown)


javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -71,7 +71,7 @@
     /**
      * A queue with all audio sources for this connection.
      */
-    private final BlockingQueue<AudioSource> queue = new LinkedBlockingQueue<>();
+    private final BlockingQueue<AudioSource> queue = new LinkedBlockingQueue<>(1);
```
Bastian, line 74: Is there a reson to still have the queue?

Mysterypotatoguy: It's been some time, but I believe to maintain the ability to block while waiting for a source without resorting to wait/notify

Bastian: For reference
image
image

Tldr: Please create a helper class that abstracts this behaviour.

javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -307,38 +299,7 @@ public synchronized boolean tryConnect() {
...
-        }
-        return source;
-    }
-
-    /**
-     * Removes the current audio source.
-     */
-    public void removeCurrentSource() {
-        currentSource.set(null);
+        return currentSource.get();
```
Bastian, line 302:
```suggestion
        return currentSource.get(timeout, unit);
```
This completely ignores the `timeout` and `unit` parameter.

javacord-core/src/main/java/org/javacord/core/util/concurrent/BlockingReference.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @param <V> The type of object referred to by this reference.
+ */
+public class BlockingReference<V> {
+
+    private final Lock lock = new ReentrantLock();
+
+    private final Condition hasValue = lock.newCondition();
+
+    private V value;
```
Bastian, line 19:
```suggestion
    private volatile V value;
```

javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -307,38 +299,7 @@ public synchronized boolean tryConnect() {
      * @throws InterruptedException If interrupted while waiting.
      */
     public AudioSource getCurrentAudioSourceBlocking(long timeout, TimeUnit unit) throws InterruptedException {
-        AudioSource source;
-        currentSourceLock.lock();
```
Bastian, line 311: The `currentSourceLock` field can be removed as it is no longer used anywhere.

javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -68,14 +59,9 @@
...
-     * The source that gets played at the moment.
-     */
-    private volatile AtomicReference<AudioSource> currentSource = new AtomicReference<>();
-
-    /**
-     * A queue with all audio sources for this connection.
+     * The source that is currently being played.
      */
-    private final BlockingQueue<AudioSource> queue = new LinkedBlockingQueue<>();
+    private volatile BlockingReference<AudioSource> currentSource = new BlockingReference<>();
```
Bastian, line 64:
```suggestion
    private final BlockingReference<AudioSource> currentSource = new BlockingReference<>();
```

Bastian:
Thx :)

PR #517, opened 2019-12-13
Convenience method for EmbedThumbnail image downloading.

robotman2412:
This pull request is about the convenience method i would like to add to the EmbedThumbnail class.
The method functions similarly to the method MessageAttachment#downloadAsImage().

This does not need any major API or source changes.

Thanks in advance.


Bastian:
Thank you for the pull request and sorry for the late response.
I've cleaned up your commits and squashed them into a single commit. I've also added download methods for the embed image and author icon.

PR #525, opened 2020-04-10
Allowed mentions

KILLEliteMaste:
Adds the ability to control the mention behavior

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
@@ -5,6 +5,7 @@ import org.javacord.api.entity.Mentionable;
 import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.message.embed.EmbedBuilder;
 import org.javacord.api.entity.message.internal.MessageBuilderDelegate;
+import org.javacord.api.entity.message.mention.AllowedMentions;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.util.internal.DelegateFactory;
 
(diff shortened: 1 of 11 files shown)


KILLEliteMaste:
> You should not follow the bad pattern of the embed builder being pushed around as argument, for embeds this should be changed too. (see #358)

Oh okay. I just began to improve my knowledge and I have choosen the EmbedBuilder as a template to get better understanding how the things work. 

> And second, why all the "custom" in the methods?

I named it like this because I thought there are general mentions like "users". This will mention all user and "custom" seemed to me more understandable that it will mention specific users. But after thinking about it again it sound a bit weird

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -423,6 +424,17 @@ public MessageBuilder addAttachmentAsSpoiler(InputStream stream, String fileName
         return this;
     }
 
+    /**
+     * Controls who will be mentioned if mentions exist in the message.
+     *
+     * @param allowedMentions The mention object.
+     * @return The current instance in order to chain call methods.
+     */
+    public MessageBuilder addAllowedMentions(AllowedMentions allowedMentions) {
```
Bastian, line 433: As this overrides the previous allowed mentions, the method should be called `setAllowedMentions(...)`.
(The same ofc also applies to the delegate, etc.)

javacord-core/src/main/java/org/javacord/core/entity/message/mention/AllowedMentionsImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    private final boolean mentionEveryoneAndHere;
+
+    /**
+     * Creates a new mention.
+     *
+     * @param mentionAllRoles        Whether it mentions all roles.
+     * @param mentionAllUsers        Whether it mentions all users.
+     * @param mentionEveryoneAndHere Whether it mentions @everyone and @here.
+     * @param allowedRoleMentions    Mentions added role id's.
+     * @param allowedUserMentions    Mentions added user id's.
```
Bastian, line 31:
```suggestion
     * @param allowedRoleMentions    Mentions added role ids.
     * @param allowedUserMentions    Mentions added user ids.
```

PR #531, opened 2020-04-24
Added method to get reaction tag of custom emoji

Joshix-1:
This method allows removing some duplicate code and helps users to get the reaction tag of a custom emoji. This is really useful, when trying to add UnicodeEmojis and CustomEmojis in the same line.


Bastian:
Thank you!

Issue #522, opened 2020-04-04
Add support for intents


Bastian:
https://discordapp.com/developers/docs/topics/gateway#gateway-intents

Bastian:
![image](https://user-images.githubusercontent.com/5033001/78458808-82de0280-76b4-11ea-94b5-a96d873bc635.png)
> Intents are optionally supported on the v6 gateway. They will become mandatory in a future gateway version.

PR #532, opened 2020-04-25
Support for intents

KILLEliteMaste:
This is a WIP feature because intents are in the current gateway version optional and Javacord can break when not using this correctly.

Things which are affected of not being able to receive events need to be taken care of in order to not break Javacord.

Current known issues with this implementation:
- When not providing the GUILDS and GUILD_PRESENCES intent some events and other GUILD intents do not work anymore. There's also a long startup time

In regard of issue #522 this provides the implementation of intents.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -14,6 +14,7 @@ import org.javacord.api.entity.channel.ServerVoiceChannel;
 import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.channel.VoiceChannel;
 import org.javacord.api.entity.emoji.KnownCustomEmoji;
+import org.javacord.api.entity.intent.Intent;
 import org.javacord.api.entity.message.Message;
 import org.javacord.api.entity.message.MessageSet;
 import org.javacord.api.entity.message.UncachedMessageUtil;
@@ -41,6 +42,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import java.util.regex.Matcher;
@@ -66,6 +68,13 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
      * @return The prefixed, used token.
      */
     String getPrefixedToken();
+
+    /**
+     * Gets the intents used to receive some or all events specified in {@code Intent}.
+     *
+     * @return The intents for the events.
+     */
+    Set<Intent> getIntents();
 
     /**
      * Gets the thread pool which is internally used.
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
@@ -1,5 +1,6 @@
 package org.javacord.api;
 
+import org.javacord.api.entity.intent.Intent;
 import org.javacord.api.event.server.ServerBecomesAvailableEvent;
 import org.javacord.api.internal.DiscordApiBuilderDelegate;
 import org.javacord.api.listener.ChainableGloballyAttachableListenerManager;
@@ -11,11 +12,13 @@ import org.javacord.api.util.ratelimit.Ratelimiter;
 
 import java.net.Proxy;
 import java.net.ProxySelector;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import java.util.function.IntPredicate;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.IntStream;
 
@@ -320,6 +323,72 @@ public class DiscordApiBuilder implements ChainableGloballyAttachableListenerMan
      */
     public boolean isShutdownHookRegistrationEnabled() {
         return delegate.isShutdownHookRegistrationEnabled();
+    }
+
+    /**
+     * Sets intent for the events which should be received.
+     *
+     * @param intents One or more intents from {@link Intent}.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setIntents(Intent... intents) {
+        setAllIntentsWhere(intent -> Arrays.asList(intents).contains(intent));
+        return this;
+    }
+
+    /**
+     * Sets all intents.
+     *
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllIntents() {
+        setAllIntentsWhere(intent -> true);
+        return this;
+    }
+
+    /**
+     * Sets all non privileged intents.
+     *
+     * <p>This is the default behavior if no intents are set in the builder.
+     *
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllNonPrivilegedIntents() {
+        setAllIntentsWhere(intent -> !intent.isPrivileged());
+        return this;
+    }
+
+    /**
+     * Sets all intents except the given intents.
+     *
+     * @param intentsToOmit One or more {@code Intent}s which should be omitted.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllIntentsExcept(Intent... intentsToOmit) {
+        setAllIntentsWhere(intent -> !Arrays.asList(intentsToOmit).contains(intent));
+        return this;
+    }
+
+    /**
+     * Sets all non privileged intents except the given intents.
+     *
+     * @param intentsToOmit One or more {@code Intent}s which should be omitted.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllNonPrivilegedIntentsExcept(Intent... intentsToOmit) {
+        setAllIntentsWhere(intent -> !intent.isPrivileged() && !Arrays.asList(intentsToOmit).contains(intent));
+        return this;
+    }
+
+    /**
+     * Sets the intents where the given predicate matches.
+     *
+     * @param condition Whether the intent should be added or not.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllIntentsWhere(Predicate<Intent> condition) {
+        delegate.setAllIntentsWhere(condition);
+        return this;
     }
 
     /**
--- a/javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
@@ -1,0 +1,214 @@
+package org.javacord.api.entity.intent;
+
+/**
+ * Represents an intent.
+ *
+ * @see <a href="https://discordapp.com/developers/docs/topics/gateway#list-of-intents">Discord docs</a>
+ */
+public enum Intent {
+
+    /**
+     * The following event are received.
+     * <ul>
+     *     <li>GUILD_CREATE</li>
+     *     <li>GUILD_UPDATE</li>
+     *     <li>GUILD_DELETE</li>
+     *     <li>GUILD_ROLE_CREATE</li>
+     *     <li>GUILD_ROLE_UPDATE</li>
+     *     <li>GUILD_ROLE_DELETE</li>
+     *     <li>CHANNEL_CREATE</li>
+     *     <li>CHANNEL_UPDATE</li>
+     *     <li>CHANNEL_DELETE</li>
+     *     <li>CHANNEL_PINS_UPDATE</li>
+     * </ul>
+     */
+    GUILDS(0, false),
+
+    /**
+     * The following events are received.
+     * <ul>
+     *     <li>GUILD_MEMBER_ADD</li>
+     *     <li>GUILD_MEMBER_UPDATE</li>
+     *     <li>GUILD_MEMBER_REMOVE</li>
+     * </ul>
+     *
+     * <p>Note: This is a privileged intent which must be enabled in the bots application page in the Developer portal
+     *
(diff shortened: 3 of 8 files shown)


javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -703,6 +703,10 @@ private void sendIdentify(WebSocket websocket) {
                 .put("$device", "Javacord")
                 .put("$referrer", "")
                 .put("$referring_domain", "");
+        //TODO: If Intents are mandatory change this to add an empty string if no Intent is set
+        if (api.getIntent() != 0) {
+            data.put("intents", api.getIntent());
+        }
```
Bastian, line 709: Just set it directly. I would expect Discord to interpret `0` correctly as "no intents" (no intents = do not send any of the gateway events).
Can you please test this?
```suggestion
        data.put("intents", api.getIntent());
```

javacord-core/src/main/java/org/javacord/core/DiscordApiBuilderDelegateImpl.java
```diff
@@ -103,6 +104,12 @@
      */
     private volatile boolean waitForServersOnStartup = true;
 
+    /**
+     * The intent count.
+     */
+    private final AtomicInteger intent = new AtomicInteger();
```
Bastian, line 110: The default should be all non-privileged intents.

javacord-core/src/main/java/org/javacord/core/DiscordApiBuilderDelegateImpl.java
```diff
@@ -333,6 +340,15 @@ public boolean isWaitingForServersOnStartup() {
         return waitForServersOnStartup;
     }
 
+    @Override
+    public void setIntent(IntentType... intentType) {
+        int intentCount = 0;
+        for (IntentType intentTypeValue : intentType) {
+            intentCount += (1 << intentTypeValue.getId());
+        }
```
Bastian, line 348: Better move the calculation into a static method in `IntentType`:
```java
public static int calculateBitmask(IntentType... intentType) { /* ... */ }
```

javacord-core/src/test/groovy/org/javacord/core/DiscordApiImplTest.groovy
```diff
@@ -320,7 +320,7 @@ class DiscordApiImplTest extends Specification {
             MockProxyManager.setSocks4SystemProperties()
 
         and:
-            def api = new DiscordApiImpl(AccountType.BOT, 'fakeBotToken', 0, 1, false, null, null, null, null, true,
+            def api = new DiscordApiImpl(AccountType.BOT, 'fakeBotToken', 0, 1,0, false, null, null, null, null, true,
```
Bastian, line 323: Please don't call me nitpicky 🙈 
```suggestion
            def api = new DiscordApiImpl(AccountType.BOT, 'fakeBotToken', 0, 1, 0, false, null, null, null, null, true,
```

javacord-api/src/main/java/org/javacord/api/entity/intent/IntentType.java
```diff
@@ -0,0 +1,182 @@
+package org.javacord.api.entity.intent;
+
+/**
+ * Represents an intent type.
+ *
+ * @see <a href="https://discordapp.com/developers/docs/topics/gateway#list-of-intents">Discord docs</a>
+ */
+public enum IntentType {
```
Bastian, line 8: All of these are `Intent`s and not `IntentType`s.
The getter and setters should also be adjusted accordingly (I suggested changes to some of them to give you an idea of the naming).

```suggestion
public enum Intent {
```

javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
```diff
@@ -297,6 +298,18 @@ public boolean isWaitingForServersOnStartup() {
...
     }
 
+    /**
+     * WIP: to work properly you have to use at least {@link IntentType#GUILD_PRESENCES} and {@link IntentType#GUILDS}
+     * Sets intent for the events which should be received.
+     *
+     * @param intentType One or more intents from {@link IntentType}.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setIntent(IntentType... intentType) {
```
Bastian, line 308: Please also update all other setters accordingly.
```suggestion
    public DiscordApiBuilder setIntents(Intent... intent) {
```

javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
```diff
@@ -297,6 +298,18 @@ public boolean isWaitingForServersOnStartup() {
         return delegate.isWaitingForServersOnStartup();
     }
 
+    /**
+     * WIP: to work properly you have to use at least {@link IntentType#GUILD_PRESENCES} and {@link IntentType#GUILDS}
```
Bastian, line 302: We will probably forget to remove this WIP message when we fix it.
As we have the development branch I don't think it is necessary to add this comment. 
```suggestion
```

javacord-api/src/main/java/org/javacord/api/entity/intent/IntentType.java
```diff
@@ -0,0 +1,182 @@
...
+     */
+    GUILD_EMOJIS(3),
+
+    /**
+     * The following events are received.
+     * <ul>
+     *     <li>GUILD_INTEGRATIONS_UPDATE</li>
+     * </ul>
+     */
+    GUILD_INTEGRATIONS(4),
```
Bastian, line 64:
```suggestion

    GUILD_INTEGRATIONS(4),
```

javacord-api/src/main/java/org/javacord/api/entity/intent/IntentType.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * The following events are received.
+     * <ul>
+     *     <li>WEBHOOKS_UPDATE</li>
+     * </ul>
+     */
+    GUILD_WEBHOOKS(5),
```
Bastian, line 71:
```suggestion

    GUILD_WEBHOOKS(5),
```

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
KILLEliteMaste: Since we use as default all non privileged intents (from one of your requested changes) it will always have a valid number.

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -66,6 +68,13 @@
      */
     String getPrefixedToken();
 
+    /**
+     * Gets the intent used to receive some or all events specified in {@link Intent}.
+     *
+     * @return The intent for the events.
+     */
+    Set<Intent> getIntent();
```
Bastian, line 76:
```suggestion
    /**
     * Gets the intents used to receive some or all events specified in {@code Intent}.
     *
     * @return The intents for the events.
     */
    Set<Intent> getIntents();
```

javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
```diff
@@ -0,0 +1,219 @@
...
+     * @return Whether this intent is privileged or not.
+     */
+    public boolean isPrivileged() {
+        switch (this) {
+            case GUILD_MEMBERS:
+            case GUILD_PRESENCES:
+                return true;
+            default:
+                return false;
+        }
```
Bastian, line 202:
```suggestion
        return required;
```

javacord-api/src/main/java/org/javacord/api/internal/DiscordApiBuilderDelegate.java
```diff
@@ -129,6 +130,13 @@
      */
     boolean isWaitingForServersOnStartup();
 
+    /**
+     * Sets the intent.
+     *
+     * @param intent One or more {@link Intent}.
+     */
+    void setIntent(Intent... intent);
```
Bastian, line 138:
```suggestion
    /**
     * Sets the intents.
     *
     * @param intent One or more {@code Intent}s.
     */
    void setIntents(Intent... intents);
```

javacord-core/src/main/java/org/javacord/core/DiscordApiBuilderDelegateImpl.java
```diff
@@ -103,6 +106,11 @@
      */
     private volatile boolean waitForServersOnStartup = true;
 
+    /**
+     * The intents. Default are all intents except the privileged
+     */
+    private final Set<Intent> intent = new HashSet<>();
```
Bastian, line 112:
```suggestion
    private final Set<Intent> intents = Arrays.stream(Intent.values())
                      .filter(intent -> !intent.isPrivileged())
                      .collect(Collectors.toCollection(HashSet::new));
```
(not written in IDE, might not be 100% correct)

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -189,6 +190,11 @@
      */
     private final int totalShards;
 
+    /**
+     * The intents to be set.
+     */
+    private final Set<Intent> intent;
```
Bastian, line 196:
```suggestion
    private final Set<Intent> intents;
```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -362,6 +369,7 @@ public DiscordApiImpl(
             String token,
             int currentShard,
             int totalShards,
+            Set<Intent> intent,
```
Bastian, line 372:
```suggestion
            Set<Intent> intents,
```

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -703,6 +704,10 @@ private void sendIdentify(WebSocket websocket) {
                 .put("$device", "Javacord")
                 .put("$referrer", "")
                 .put("$referring_domain", "");
+
+        data.put("intents",api.getIntent().isEmpty()
+                ? 32509 : Intent.calculateBitmask(api.getIntent().toArray(new Intent[0])));
```
Bastian, line 709: * What is `32509`? If that's the bitmask for all non-privileged intents: See my suggestion in `DiscordApiBuilderDelegateImpl`.
* If you default `api.getIntent().isEmpty()` to `32509`, it is impossible for the user to specify "I don't want any intents". Just set whatever the result of `api.getIntents()` is.
```suggestion
        data.put("intents", Intent.calculateBitmask(api.getIntent().toArray(new Intent[0])));
```

javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
```diff
@@ -0,0 +1,219 @@
...
+    private final int id;
+
+    /**
+     * Class constructor.
+     *
+     * @param id The id of the activity type
+     */
+    Intent(int id) {
+        this.id = id;
+    }
```
Bastian, line 176: Better have it as a parameter + field instead of using a switch statement.
```suggestion
    Intent(int id, boolean privileged) {
        this.id = id;
        this.privileged = privileged;
    }
```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1094,6 +1106,11 @@ public String getPrefixedToken() {
         return accountType.getTokenPrefix() + token;
     }
 
+    @Override
+    public Set<Intent> getIntents() {
+        return intents;
```
Bastian, line 1111: Oops, I missed this one in earlier reviews, sorry.
All collection returned by Javacord should be immutable. In this case, no defensive copy is necessary, as we never modify the set internally.
```suggestion
        return Collections.unmodifiableSet(intents);
```

KILLEliteMaste:
@Bastian It is still untested what happens if we pass `0` as an intent argument to discord because Javacord wont login properly without the GUILD_PRESENCES and GUILDS intent

Bastian:
That's fine.
I'm pretty sure that it just means that we won't receive any events besides passthrough events.

Bastian:
I've tested it and it seems to work fine 👍 

Just one more suggestion, because I noticed that it is very annoying: We should provide some more helper methods to construct intents. A simple use-case like "Set all intents except xyz" currently looks like this:
```java
DiscordApi api = new DiscordApiBuilder()
        .setToken(token)
        .setIntents(Arrays.stream(Intent.values()).filter(intent -> intent != Intent.GUILD_MESSAGES).toArray(Intent[]::new))
        .login()
        .join();
```

Wouldn't it be more convenient to have
```java
DiscordApi api = new DiscordApiBuilder()
        .setToken(token)
        .setAllIntentsExcept(Intent.GUILD_MESSAGES)
        .login()
        .join();
```
?

My suggestion would be to have the following methods:
```java
setAllIntents();
setAllIntentsExcept(Intent... intentsToOmit);
setAllNonPrivilegedIntents(); // This is our default behaviour, but maybe someone wants to make it explicit
setAllNonPrivilegedIntentsExcept(Intent... intentsToOmit);
setAllIntentsWhere(Predicate<Intent> condition); // You can use this method to easily implement all the other ones
```

Here's a small example of `setAllIntentsWhere` usage because it is a bit more complicated:
```java
DiscordApi api = new DiscordApiBuilder()
        .setToken(token)
        .setAllIntentsWhere(intent -> !intent.isPrivileged() && intent == Intent.GUILD_MESSAGES)
        .login()
        .join();
```
or to implement the `setAllNonPrivilegedIntentsExcept` method:
```java
DiscordApiBuilder setAllNonPrivilegedIntentsExcept(Intent... intentsToOmit) {
    setAllIntentsWhere(intent -> !intent.isPrivileged() && !Arrays.asList(intentsToOmit).contains(intent));
    return this;
}
```

PR #535, opened 2020-05-01
Message#getLink method.

Joshix-1:
It is really helpful to have such a method. Some don't know how the links are composed, and for others it is helpful do just use the Javacord method instead of creating an own.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -689,6 +689,20 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     }
 
     /**
+     *  Gets the link leading to this message.
+     *
+     * @return The message link.
+     */
+    default String getLink() {
+        return "https://discordapp.com/channels/"
+                + getServer().map(DiscordEntity::getIdAsString).orElse("@me")
+                + "/"
+                + getChannel().getIdAsString()
+                + "/"
+                + getIdAsString();
+    }
+
+    /**
      * Gets a list of all custom emojis in the message.
      *
      * @return The list of custom emojis in the message.
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -20,6 +20,8 @@ import org.javacord.api.entity.user.User;
 import org.javacord.api.listener.message.MessageAttachableListenerManager;
 import org.javacord.api.util.DiscordRegexPattern;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -688,6 +688,25 @@ default String getReadableContent() {
         return ESCAPED_CHARACTER.matcher(content).replaceAll("${char}");
     }
 
+    /**
+     *  Gets the link leading to this message.
+     *
+     * @return The message link.
+     */
+    default String getLink() {
```
Vampire, line 696: Should this and the others probably return `URL` instead?

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -688,6 +688,25 @@ default String getReadableContent() {
...
+        StringBuilder link = new StringBuilder("https://discordapp.com/channels/");
+        if(getServer().isPresent()) {
+            link.append(getServer().get().getIdAsString());
+        } else {
+            link.append("@me");
+        }
```
Vampire, line 702: I'd suggest instead
```java
getServer()
    .map(DiscordEntity::getIdAsString)
    .orElse("@me")
```

Vampire:
Also you should change the target branch of the PR to `development`

Vampire:
And you should execute `check` task, or at least look what our CI said, as you violated the defined checkstyle rules. Please fix up your commits, so that each one results in a green build.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
Joshix-1: Good Idea :)

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
Joshix-1: I'm not sure if that's better. I would think it is more convenient to leave it as String. I don't see any use case where it would be better to have it as an URL.

Vampire: Well, it **is** an `URL`, that could be reason enough.
But I'll let that decision for @Bastian, I just mentioned it to be thought about. :-)

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -688,6 +688,20 @@ default String getReadableContent() {
...
     }
 
+    /**
+     *  Gets the link leading to this message.
+     *
+     * @return The message link.
+     */
+    default String getLink() {
+        return "https://discordapp.com/channels/"
+                + getServer().map(Server::getIdAsString).orElse("@me")
```
Vampire, line 698:
```suggestion
                + getServer().map(DiscordEntity::getIdAsString).orElse("@me")
```

Vampire:
Please rebase your branch instead of merging in `development` and fix up each commit so that it has a green checkmark at https://github.com/Javacord/Javacord/pull/535/commits meaning each commit individually results in a successful build.
You might even consider making one commit of them, the changes are pretty atomic alltogether I'd say. :-)

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
Bastian: I have to agree with Vampire. It is a url and thus should return a `URL` object.
This is also consistent with all the other methods of Javacord that return a url.

Bastian:
Besides the method return type, this looks good to me.
Thanks! :)

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
Joshix-1: Then I will change that :-)

Joshix-1: How should I handle the MalformedURLException?

Bastian: In other places we just return `null` and log a error, but this would be better:
```java
try {
    URL url = new URL("");
} catch (MalformedURLException e) {
    throw new AssertionError("Message link is malformed", e);
}
```

Vampire: I would have also said catch and transform to `AssertionError`.
As we construct the URL and know it is valid, it is a programming error if a MURLE is thrown, so translating it to an `AssertionError` is the way to go.

PR #536, opened 2020-05-01
method to get Message by link

Joshix-1:
Helpful method to get messages by their link.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -26,6 +26,7 @@ import org.javacord.api.entity.user.User;
 import org.javacord.api.entity.user.UserStatus;
 import org.javacord.api.entity.webhook.Webhook;
 import org.javacord.api.listener.GloballyAttachableListenerManager;
+import org.javacord.api.util.DiscordRegexPattern;
 import org.javacord.api.util.concurrent.ThreadPool;
 import org.javacord.api.util.ratelimit.Ratelimiter;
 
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -754,6 +756,24 @@ default AccountUpdater createAccountUpdater() {
...
+     * Gets a message by its link.
+     *
+     * @param link The link of the message.
+     * @return The message with the given link.
+     */
+    default CompletableFuture<Message> getMessageByLink(String link) {
+        Matcher matcher = DiscordRegexPattern.MESSAGE_LINK.matcher(link);
+
+        if (!matcher.matches()) {
+            return CompletableFuture.completedFuture(null);
```
Vampire, line 769: Javacord does not return `null`, except for `Void`, it should either be `CompletableFuture<Optional<Message>>` or `Optional<CompletableFuture<Message>>` which we both have in the codebase already. I guess here `Optional<CompletableFuture<Message>>` is more appropriate as the text channel is resolved first.

Vampire:
But what if you have a valid link, but the channel id does not match a text channel.
That is where previously `null` was given back in the future and that is not right.

Bastian:
Ohh right, I didn't cosider this case.
In that case, we should keep the `Optional<CompletableFuture<Message>> getMessageByLink(String link)` method (but still add the `getCachedMessageByLink(...)` method and the `IllegalArgumentException` for invalid links).

Bastian:
Thank you! 👍

Issue #526, opened 2020-04-18
http status codes being ignored


Pepich:
JavaCord will attempt to parse Discord's repsonse on certain actions (e.g. sendMessage()) without checking the http response code. This results in HTML being parsed as JSON, and an exception being thrown. Attached is a stacktrace of when the API returned an HTTP 502, but JavaCord tried parsing the response as JSON anyway.

https://hastebin.com/zodofazefi.php

Bastian:
Thank you for reporting! 👍 
Do you still have the error log? From my experience, hastebin has been pretty unreliable in the past 2 years. At the moment it just displays a blank page for me :(

Pepich:
Let‘s try a slightly more reliable service the. https://gist.github.com/Pepich/a7bbb9cc5f8cc8bc634aa0df64b75437#file-exception-log
The exception was thrown upon joining a message send. If you need more info lemme know :)

Vampire:
As far as I remember when we last talked about this on Discord, the outcome was, that the HTTP status is checked, but only after the content is parsed as JSON already.

Pepich:
[Here's another similar instance](https://gist.github.com/Pepich/cf388c5474e8fc898614aaf36b5cb546) - this time it has nothing to do with an error message but instead was caused by the connection being dropped shortly before/after it. The message that Discord returned is unknown to me as the stacktrace didn't reveal any info about it :/

PR #550, opened 2020-05-04
Fix parsing exception for invalid Discord response

Bastian:
Fixes https://github.com/Javacord/Javacord/issues/526

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/rest/RestRequestResult.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/rest/RestRequestResult.java
@@ -1,10 +1,13 @@
 package org.javacord.core.util.rest;
 
+import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.NullNode;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
+import org.apache.logging.log4j.Logger;
+import org.javacord.core.util.logging.LoggerUtil;
 
 import java.io.IOException;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


javacord-core/src/main/java/org/javacord/core/util/rest/RestRequestResult.java
```diff
@@ -37,7 +45,14 @@ public RestRequestResult(RestRequest<?> request, Response response) throws IOExc
         } else {
             stringBody = body.string();
             ObjectMapper mapper = request.getApi().getObjectMapper();
-            JsonNode jsonBody = mapper.readTree(stringBody);
+            JsonNode jsonBody;
+            try {
+                jsonBody = mapper.readTree(stringBody);
+            } catch (JsonParseException e) {
```
Vampire, line 51: Wouldn't it be better to instead only parse the body if the answer had a 2xx response code,
or maybe even better parse the body lazily only if it is requested the first time?

Bastian: > Wouldn't it be better to instead only parse the body if the answer had a 2xx response code,

Actually, Discord should answer every request with a JSON response, not only 2xx. A non-JSON response only happens if something broke on Discord's side (in this case a `502 - Bad Gateway`).

> or maybe even better parse the body lazily only if it is requested the first time?
I don't know if there are even any cases where we don't need the body.

Vampire: > Actually, Discord should answer every request with a JSON response, not only 2xx. A non-JSON response only happens if something broke on Discord's side (in this case a 502 - Bad Gateway).

Actually it could also be some proxy or whatnot.
Assuming a valid JSON-formatted response for a non 2xx response is fragile.

Bastian: > Assuming a valid JSON-formatted response for a non 2xx response is fragile.

The API even says, that it may a JSON response for errors:
> Along with the HTTP error code, our API can also return more detailed error codes through a code key in the JSON error response. The response will also contain a message key containing a more friendly error string.

https://discordapp.com/developers/docs/topics/opcodes-and-status-codes#json

The 5xx status codes are probably the only ones that **may** not return a JSON response.

Bastian:
Thanks for the review :)

Issue #390, opened 2018-10-19
Audio support


Bastian:
## Api design
This is a description about Javacord's upcoming audio api. It is "work in progress" and might change significantly during development. Feedback is very appreciated.

### Joining a voice channel
Joining a voice channel is nothing special. You call `#connect()` on a `VoiceChannel` and receive an `CompletableFuture<AudioConnection>`. If the bot is already connected to another voice channel on the same server, it will leave the old channel and join the new one (aka. move).
```java
VoiceChannel channel = ...;
channel.connect().thenAcceptAsync(connection -> {
    // Do stuff
});
```

### AudioSources
To be flexible with all various types of inputs, there will be an interface `AudioSource` with different implementations. Creating an audio source for a YouTube video might look like this:
```java
AudioSource source = new YoutubeAudioSourceBuilder()
    .setUrl("https://youtu.be/NvS351QKFV4")
    .build();

// Maybe, there might be a "shortcut" to skip the builder for "simpler" sources
AudioSource source = YouTubeAudiSource.of("https://youtu.be/NvS351QKFV4");
```
An audio source from an mp3 file, might look like this:
```java
AudioSource source = new FileAudioSourceBuilder()
    .setFile(new File("C:/Users/Bastian/Music/moskau.mp3"))
    .build();
```

Javacord will not provide any implementation for `AudioSource`s, but just interfaces like
- `AudioSource` 
- `PauseableAudioSource` -> Has additional methods like `#pause()` and `#unpause()`
- `SeekableAudioSource`/`NavigableAudioSource` -> Has additional methods like `#jumpToPosition(int, TimeUnit)`
- `...` -> ...

The actual implementation will be moved into an own repository or module.

### Sending audio
Sending audio is also very simple. Audio connections will have a queue and just consume the audio sources one after the other. You can also queue an audio source multiple times.
```java
AudioConnection connection = ...;
AudioSource source = ...;
connection.queue(source);
```

### Handling exceptions
Exceptions will be treated like normal "events". 
```java
AudioConnection connection = ...;
connection.addAudioSourceExceptionListener(event -> {
    event.getException().printStackTrace();
    // Skip the audio source
    event.getConnection().dequeue(event.getAudioSource());
});
```
Like every other Javacord listener, it can also be attached to other objects. In this case you can attach the listener to `DiscordApi`, `Server`, `VoiceChannel`, `AudioConnection`, `AudioSource`.

### Interacting with the queue
[...]

PR #553, opened 2020-05-04
Improve readme template

Bastian:
Improves the layout of the README.

## 📋 Changelog

* Some sections are reworded, especially the introduction
* More GIFs and images
* Emojis everywhere 🤯 (I will have a poll if we should keep them. I like them, but I can see how it can be a little bit to much for some)
* Examples, examples, and even more examples 
* "Large bots using Javacord" section
* "License" section

## 🔨 TODO

- [x] Do not use a third-party website to host images -> Either have them inside of the `.github` folder or upload them to `javacord.org` like we already did it with the banner)
- [x] Poll if we should keep the emojis 🗳
- [x] Proof-read for spelling mistakes and typos
- [x] Replace the link to my private [Lavaplayer-Wrapper](https://github.com/Bastian/Lavaplayer-Wrapper) with our offical one (which does not exist yet)

## 💡 Tipp

Its best to view the README here to see how it will look like when being merged: 
- [README_template.md](https://github.com/Bastian/Javacord/blob/improve-readme/.github/README_template.md)

## 🔒 When to merge

~~The new README already assumes that we have audio support, so we have to wait for #390~~
The audio example is no longer part of the PR. It can still be found at https://gist.github.com/Bastian/d3eacea6c2e030633cde5b9c0e132860 for later usage.


Bastian:
Please use this message to vote if you
* 👍 Like the emojis and think we should keep them in the README
* 👎 Dislike the emojis and think we should get rid of them
* 👀 Don't care

Bastian:
I've removed the audio example to allow the PR to get merged before audio is released.
The example can still be found here: https://gist.github.com/Bastian/d3eacea6c2e030633cde5b9c0e132860

We can re-add it with the release of audio.

Issue #554, opened 2020-05-04
Change endpoints from discordapp.com to discord.com


Bastian:
![image](https://user-images.githubusercontent.com/5033001/81005169-cbbcde80-8e4d-11ea-944b-2629cb5aa7cc.png)

PR #555, opened 2020-05-04
Change the occourences of discordapp.com to discord.com 

Joshix-1:
see #554!
I changed everything I could find. I don't know if it is complete.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/BotInviteBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/BotInviteBuilder.java
@@ -12,7 +12,7 @@ public class BotInviteBuilder {
      * The base link of a bot invite.
      */
     public static final String BASE_LINK =
-            "https://discordapp.com/oauth2/authorize?client_id=%s&scope=bot&permissions=%s";
+            "https://" + Javacord.DISCORD_DOMAIN + "/oauth2/authorize?client_id=%s&scope=bot&permissions=%s";
 
     /**
      * The client id of the bot's application.
--- a/javacord-api/src/main/java/org/javacord/api/Javacord.java
+++ b/javacord-api/src/main/java/org/javacord/api/Javacord.java
@@ -32,6 +32,16 @@ public class Javacord {
     public static final Instant BUILD_TIMESTAMP;
 
     /**
+     * The domain of Discord.
+     */
+    public static final String DISCORD_DOMAIN = "discord.com";
+
+    /**
+     * The domain of Discord's CDN.
+     */
+    public static final String DISCORD_CDN_DOMAIN = "cdn.discordapp.com";
+
+    /**
      * The display version of the current Javacord version.
      * If the current Javacord version is a release version, it is equal to {@link #VERSION}.
      * If the current Javacord version is a snapshot version, it consists of the {@link #VERSION}, the
@@ -98,14 +108,14 @@ public class Javacord {
     /**
      * The gateway version from Discord which we are using.
      * A list with all gateway versions can be found
-     * <a href="https://discordapp.com/developers/docs/topics/gateway#gateways-gateway-versions">here</a>.
+     * <a href="https://discord.com/developers/docs/topics/gateway#gateways-gateway-versions">here</a>.
      */
     public static final String DISCORD_GATEWAY_VERSION = "6";
 
     /**
      * The API version from Discord which we are using.
      * A list with all API versions can be found
-     * <a href="https://discordapp.com/developers/docs/reference#api-versioning-api-versions">here</a>.
+     * <a href="https://discord.com/developers/docs/reference#api-versioning-api-versions">here</a>.
      */
     public static final String DISCORD_API_VERSION = "6";
 
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -1,6 +1,7 @@
 package org.javacord.api.entity.message;
 
 import org.javacord.api.DiscordApi;
+import org.javacord.api.Javacord;
 import org.javacord.api.entity.DiscordEntity;
 import org.javacord.api.entity.UpdatableFromCache;
 import org.javacord.api.entity.channel.ChannelType;
@@ -698,7 +699,7 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
      */
     default URL getLink() throws AssertionError {
         try {
-            return new URL("https://discordapp.com/channels/"
+            return new URL("https://" + Javacord.DISCORD_DOMAIN + "/channels/"
                     + getServer().map(DiscordEntity::getIdAsString).orElse("@me")
                     + "/"
                     + getChannel().getIdAsString()
--- a/javacord-api/src/main/java/org/javacord/api/util/DiscordRegexPattern.java
+++ b/javacord-api/src/main/java/org/javacord/api/util/DiscordRegexPattern.java
@@ -78,18 +78,18 @@ public class DiscordRegexPattern {
                             + ">                   # '>' \n");
 
     /**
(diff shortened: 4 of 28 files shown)


Joshix-1:
I read somewhere, that the cdn.discordapp.com won't change.

Bastian:
image
You missed one in the `mockserver.properties` file.

Bastian:
> I read somewhere, that the cdn.discordapp.com won't change.

Ineed. Here's the reason:
image

Bastian:
> Maybe a good solution will be to store the main url (discord.com) in a properties. load it on the begin of an app anreplace all real occuration (not javadoc) by this value

Agreed. Even though I think that it should be enough to have it as a `public static final String` constant somewhere:
```java
public static String DISCORD_DOMAIN = "discord.com";
public static String DISCORD_CDN_DOMAIN = "cnd.discordapp.com";
```

Maybe even in the API package in the `DiscordApi` interface?

Vampire:
> You missed one in the mockserver.properties file.

Just have both in there.
It is for the generated SSL certificate for MockServer based tests.
Doesn't hurt to simply have both in there.

javacord-api/src/main/java/org/javacord/api/Javacord.java
```diff
@@ -98,14 +98,14 @@
     /**
```
CharlesLgn, line 98: you could put the two constant here

`public static final String URL_DISCORD = "discord.com"` ...

and then call this constants.

javacord-api/src/main/java/org/javacord/api/BotInviteBuilder.java
```diff
@@ -12,7 +12,7 @@
      * The base link of a bot invite.
      */
     public static final String BASE_LINK =
-            "https://discordapp.com/oauth2/authorize?client_id=%s&scope=bot&permissions=%s";
+            "https://discord.com/oauth2/authorize?client_id=%s&scope=bot&permissions=%s";
```
CharlesLgn, line 15: `"https://"+ DISCORD_URL +"/oauth2/authorize?client_id=%s&scope=bot&permissions=%s"`
or using `String.formatMessage`

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -698,7 +698,7 @@ default String getReadableContent() {
      */
     default URL getLink() throws AssertionError {
         try {
-            return new URL("https://discordapp.com/channels/"
+            return new URL("https://discord.com/channels/"
```
CharlesLgn, line 701: `"https://"+ DISCORD_URL +"/channels/"`
or using `String.formatMessage`

javacord-api/src/main/java/org/javacord/api/util/DiscordRegexPattern.java
```diff
@@ -78,18 +78,18 @@
...
-                            + "\\Qdiscordapp.com/channels/\\E # 'discordapp.com/channels/' \n"
-                            + "(?:(?<server>[0-9]++)|@me)     # '@me' or the server id as named group \n"
-                            + "/                              # '/' \n"
-                            + "(?<channel>[0-9]++)            # the textchannel id as named group \n"
-                            + "/                              # '/' \n"
-                            + "(?<message>[0-9]++)            # the message id as named group \n");
+            Pattern.compile("(?x)                              # enable comment mode \n"
+                            + "(?i)                            # ignore case \n"
+                            + "(?:https?+://)?+                # 'https://' or 'http://' or '' \n"
+                            + "discord(?:app)?\\.com/channels/ # 'discord(app).com/channels/' \n"
```
CharlesLgn, line 87: why adding `(?:app)?\\` the url will only be `discord.com/channels/` no ?

javacord-core/src/main/java/org/javacord/core/util/rest/RestEndpoint.java
```diff
@@ -153,7 +153,7 @@ public void setGlobal(boolean global) {
      * @return The full url of the endpoint.
      */
     public String getFullUrl(String... parameters) {
-        StringBuilder url = new StringBuilder("https://discordapp.com/api/v" + Javacord.DISCORD_API_VERSION + getEndpointUrl());
+        StringBuilder url = new StringBuilder("https://discord.com/api/v" + Javacord.DISCORD_API_VERSION + getEndpointUrl());
```
CharlesLgn, line 156: same remark as other file (maybe have a constant for `https://discord.com`

javacord-api/src/main/java/org/javacord/api/util/DiscordRegexPattern.java
Joshix-1: Currently both is possible. And I would remove the (?:app)? only when only discordapp.com/channels is possible.

Vampire: I'd also keep it, someone could post an old link. But add `+` after `?`, there is no need to backtrack.

Vampire: And now the alignment again. :-)

Bastian:
@Joshix-1 Do you plan to add the constants in this PR?

I would be ok if you don't. In this case I would create an issue for it and someone else (maybe @CharlesLgn ;-) ) can create a separate PR for it.

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -49,6 +49,8 @@
  * This class is the most important class for your bot, containing all important methods, like registering listener.
  */
 public interface DiscordApi extends GloballyAttachableListenerManager {
+    public static final String DISCORD_DOMAIN = "discord.com";
```
Bastian, line 52: If it's in the public api, it should have Javadoc comments.
```suggestion

    /*
     * The domain of Discord.
     */
    public static final String DISCORD_DOMAIN = "discord.com";
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -49,6 +49,8 @@
...
+    public static final String DISCORD_CDN_DOMAIN = "cnd.discordapp.com";
```
Bastian, line 53:
```suggestion
    /*
     * The domain of Discord's CDN.
     */
    public static final String DISCORD_CDN_DOMAIN = "cnd.discordapp.com";
```

javacord-api/src/main/java/org/javacord/api/BotInviteBuilder.java
Bastian, line 15: When we have the constants, we should also use them.
```suggestion
            "https://" + DiscordApi.DISCORD_DOMAIN + "/oauth2/authorize?client_id=%s&scope=bot&permissions=%s";
```

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
Bastian, line 701:
```suggestion
            return new URL("https://" + DiscordApi.DISCORD_DOMAIN + "/channels/"
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
DoNotSpamPls, line 53: cnd != cdn

javacord-core/src/main/java/org/javacord/core/util/rest/RestEndpoint.java
Bastian, line 156:
```suggestion
        StringBuilder url = new StringBuilder("https://" + DiscordApi.DISCORD_DOMAIN + "/api/v" + Javacord.DISCORD_API_VERSION + getEndpointUrl());
```

PR #557, opened 2020-05-07
Enhance ChannelType with more information

Bastian:
I will need this for https://github.com/Javacord/Javacord/pull/556 but I feel like this is better suited to be in its own pull request, as this is an independent change.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
@@ -1,31 +1,55 @@
 package org.javacord.api.entity.channel;
 
+import java.util.Arrays;
+
 /**
  * A enum with all different channel types.
  */
 public enum ChannelType {
 
-    SERVER_TEXT_CHANNEL(0),
-    PRIVATE_CHANNEL(1),
-    SERVER_VOICE_CHANNEL(2),
-    GROUP_CHANNEL(3),
-    CHANNEL_CATEGORY(4),
-    SERVER_NEWS_CHANNEL(5),
-    SERVER_STORE_CHANNEL(6),
-    UNKNOWN(-1);
+    SERVER_TEXT_CHANNEL(0, true, false, true),
+    PRIVATE_CHANNEL(1, true, true, false),
+    SERVER_VOICE_CHANNEL(2, false, true, true),
+    GROUP_CHANNEL(3, true, true, false),
+    CHANNEL_CATEGORY(4, false, false, true),
+    SERVER_NEWS_CHANNEL(5, true, false, true),
+    SERVER_STORE_CHANNEL(6, true, false, true),
+    UNKNOWN(-1, false, false, false);
+
+    private static final ChannelType[] textChannelTypes = Arrays.stream(ChannelType.values())
+            .filter(ChannelType::isTextChannelType)
+            .toArray(ChannelType[]::new);
+
+    private static final ChannelType[] voiceChannelTypes = Arrays.stream(ChannelType.values())
+            .filter(ChannelType::isVoiceChannelType)
+            .toArray(ChannelType[]::new);
+
+    private static final ChannelType[] serverChannelTypes = Arrays.stream(ChannelType.values())
+            .filter(ChannelType::isServerChannelType)
+            .toArray(ChannelType[]::new);
 
     /**
      * The id of the channel type.
      */
     private final int id;
 
+    private final boolean textChannelType;
+    private final boolean voiceChannelType;
+    private final boolean serverChannelType;
+
     /**
      * Creates a new channel type.
      *
      * @param id The id of the channel type.
+     * @param textChannelType Whether this type is a text channel type or not.
+     * @param voiceChannelType Whether this type is a voice channel type or not.
+     * @param serverChannelType Whether this type is a server channel type or not.
      */
-    ChannelType(int id) {
+    ChannelType(int id, boolean textChannelType, boolean voiceChannelType, boolean serverChannelType) {
         this.id = id;
+        this.textChannelType = textChannelType;
+        this.voiceChannelType = voiceChannelType;
+        this.serverChannelType = serverChannelType;
     }
 
     /**
@@ -35,6 +59,33 @@ public enum ChannelType {
      */
     public int getId() {
         return id;
+    }
+
+    /**
+     * Checks if this type is a text channel type.
+     *
+     * @return Whether this type is a text channel type or not.
+     */
+    public boolean isTextChannelType() {
+        return textChannelType;
+    }
+
+    /**
+     * Checks if this type is a voice channel type.
+     *
+     * @return Whether this type is a voice channel type or not.
+     */
+    public boolean isVoiceChannelType() {
+        return voiceChannelType;
+    }
+
+    /**
+     * Checks if this type is a server channel type.
+     *
+     * @return Whether this type is a server channel type or not.
+     */
+    public boolean isServerChannelType() {
+        return serverChannelType;
     }
 
     /**
@@ -52,4 +103,30 @@ public enum ChannelType {
         return UNKNOWN;
     }
 
+    /**
+     * Gets an array with all types that are text channel types.
+     *
+     * @return All types that are text channel types.
+     */
+    public static ChannelType[] getTextChannelTypes() {
+        return textChannelTypes;
+    }
+
+    /**
+     * Gets an array with all type that are voice channel types.
+     *
+     * @return All types that are voice channel types.
+     */
+    public static ChannelType[] getVoiceChannelTypes() {
+        return voiceChannelTypes;
+    }
+
+    /**
+     * Gets an array with all types that are server channel types.
+     *
+     * @return All types that are server channel types.
+     */
+    public static ChannelType[] getServerChannelTypes() {
+        return serverChannelTypes;
+    }
 }


javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -1,31 +1,55 @@
...
+    SERVER_TEXT_CHANNEL(0, true, false, true),
+    PRIVATE_CHANNEL(1, true, true, false),
+    SERVER_VOICE_CHANNEL(2, false, true, true),
+    GROUP_CHANNEL(3, true, true, false),
+    CHANNEL_CATEGORY(4, false, false, true),
+    SERVER_NEWS_CHANNEL(5, true, false, true),
+    SERVER_STORE_CHANNEL(6, true, false, true),
+    UNKNOWN(-1, false, false, false);
+
+    private static final ChannelType[] textChannels = Arrays.stream(ChannelType.values())
```
Vampire, line 19: I guess you tested these work?
I mean to remember I sometimes had problems with statically initializing some fields in enums derived from the constants,
but I don't remember under which circumstances right now.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -37,6 +61,33 @@ public int getId() {
         return id;
     }
 
+    /**
+     * Checks if this given type is a text channel.
```
Vampire, line 65: -`given`
There is nothing given.
Given usually refers to arguments *given* to the method.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -37,6 +61,33 @@ public int getId() {
...
+     *
+     * @return Whether this type is a text channel or not.
+     */
+    public boolean isTextChannel() {
+        return textChannel;
+    }
+
+    /**
+     * Checks if this given type is a voice channel.
```
Vampire, line 74: -`given`

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -37,6 +61,33 @@ public int getId() {
...
+     *
+     * @return Whether this type is a voice channel or not.
+     */
+    public boolean isVoiceChannel() {
+        return voiceChannel;
+    }
+
+    /**
+     * Checks if this given type is a server channel.
```
Vampire, line 83: -`given`

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -52,4 +103,30 @@ public static ChannelType fromId(int id) {
         return UNKNOWN;
     }
 
+    /**
+     * Gets an array with all types that are text channels.
+     *
+     * @return All text channels.
+     */
+    public static ChannelType[] getTextChannels() {
```
Vampire, line 111: Maybe better `getTextChannelTypes`? also for the other two methods and the three variables of course.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -52,4 +103,30 @@ public static ChannelType fromId(int id) {
         return UNKNOWN;
     }
 
+    /**
+     * Gets an array with all types that are text channels.
+     *
+     * @return All text channels.
```
Vampire, line 109: +`types that are`

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -52,4 +103,30 @@ public static ChannelType fromId(int id) {
...
+     */
+    public static ChannelType[] getTextChannels() {
+        return textChannels;
+    }
+
+    /**
+     * Gets an array with all type that are voice channels.
+     *
+     * @return All voice channels.
```
Vampire, line 118: +`types that are`

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -52,4 +103,30 @@ public static ChannelType fromId(int id) {
...
+     */
+    public static ChannelType[] getVoiceChannels() {
+        return voiceChannels;
+    }
+
+    /**
+     * Gets an array with all types that are server channels.
+     *
+     * @return All server channels.
```
Vampire, line 127: +`types that are`

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
Bastian: Oops, ofc you are right

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
Bastian: Agreed

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
Bastian: I've tested it and it works like intended.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -1,31 +1,55 @@
...
+    CHANNEL_CATEGORY(4, false, false, true),
+    SERVER_NEWS_CHANNEL(5, true, false, true),
+    SERVER_STORE_CHANNEL(6, true, false, true),
+    UNKNOWN(-1, false, false, false);
+
+    private static final ChannelType[] textChannels = Arrays.stream(ChannelType.values())
+            .filter(ChannelType::isTextChannel)
+            .toArray(ChannelType[]::new);
+
+    private static final ChannelType[] voiceChannels = Arrays.stream(ChannelType.values())
```
Vampire, line 23: These should also be `...Type` according to the new setter names, shouldn't they?

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -37,6 +61,33 @@ public int getId() {
         return id;
     }
 
+    /**
+     * Checks if this type is a text channel.
+     *
+     * @return Whether this type is a text channel or not.
+     */
+    public boolean isTextChannel() {
```
Vampire, line 69: Do we also want these named differently?
After all it is still no text channel.
So maybe `isTextChannelType()` or `isTextType()` or similar? (The fields and parameters accordingly of course)

Bastian: I would be fine with both (`isTextChannel()` and `isTextChannelType()`). If you think the latter is better, I'll change it.

Vampire: I think so, because purely from the wording it feels wrong as it is not a channel, though I'm not too hard on this either.
If in doubt, make a poll :-D

Bastian: > If in doubt, make a poll :-D

That's way too unimportant to justify a poll.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -1,31 +1,55 @@
...
+    private static final ChannelType[] serverChannelTypes = Arrays.stream(ChannelType.values())
+            .filter(ChannelType::isServerChannelType)
+            .toArray(ChannelType[]::new);
 
     /**
      * The id of the channel type.
      */
     private final int id;
 
+    private final boolean isTextChannelType;
```
Vampire, line 36: Was it intentional that you now also `is`-prefixed the fields and parameters?
I guess not.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -52,4 +103,30 @@ public static ChannelType fromId(int id) {
         return UNKNOWN;
     }
 
+    /**
+     * Gets an array with all types that are text channels.
```
Vampire, line 107: And while we are nitpicking "that are ... channel types" :-D

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
Bastian: Yes, this was intentional ;)

I know that we usually don't have this prefix in Javacord, but I feel like the name `textChannelType` or before just `textChannel` doesn't describe the type really good.

I even did a quick Google search for best practices before committing and found this comment pretty convincing. 
[image](https://stackoverflow.com/a/5326487)

There are some other opionions out there, too, like
[image](https://stackoverflow.com/a/5322730)


To be honest, I'm a bit undecided which convention is better. When I see a variable called `textChannelType` without its type, I wouldn't expect it to be a boolean value but a `ChannelType`.

In practice, it doesn't really matter and a simple "what name fits best" is probably enough as a rule of thumb.

Vampire: I know of that first mentioned practice but I dislike it almost as much as omitting `get` on getter, but just having `foo()` to get the value for the `foo` property.

Both just feels horribly wrong to me and afaik most devs.
And to my knowledge, the most common practice indeed does not use the `is` on the fields,
but to name the field like the property it represents.
Property being the term used in the JavaBeans conventions.

I just hoped you would have done this by accident.
If not I find it sad, but ok.
A name alone often is not enough to conclude a meaning.
Take the german "Boxen find ich klasse", do you know whether it means the boxes or boxing without context?
The same is here, context is just important and that's what documentation and IntelliSense is for and there are for sure many other places like this in Javacord.

But ok, if that was a delibarte decision, fine.
Not the first point where I disagree with the final state. :-)

Bastian: I have no strong feelings about the notation.
If you dislike it, I'm fine with reverting it. :-)

Bastian:
Having used half half a dozen different programming languages in the past 1-2 years, all of them with their own conventions, I got a bit more relaxed with some conventions.

> lgtm now :-)

Finally! 😄 
I didn't expect this PR to need 5 iterations. I wasn't even sure if I should request your review for such a small change. Glad I did.

Vampire:
Ah, there's always some nits to pick and for such a small change what else should I review than nits? :-D

PR #561, opened 2020-05-08
Housekeeping

Saladoc:
Some minor fixes I noticed while working on other stuff but found out of scope for the respective PRs

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/logging/ExceptionLoggerDelegateImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/logging/ExceptionLoggerDelegateImpl.java
@@ -42,8 +42,8 @@ public class ExceptionLoggerDelegateImpl implements ExceptionLoggerDelegate {
     public UncaughtExceptionHandler getUncaughtExceptionHandler() {
         return (thread, throwable) -> logger.error(
                 "Caught unhandled exception on thread '{}'!",
-                thread::getName,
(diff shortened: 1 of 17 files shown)


Bastian:
Unused imports? O.o
Should our checkstyle checks detect  those?

Saladoc:
It probably should, I've also added the check.

javacord-core/src/main/java/org/javacord/core/util/logging/ExceptionLoggerDelegateImpl.java
```diff
@@ -41,9 +41,9 @@
     @Override
     public UncaughtExceptionHandler getUncaughtExceptionHandler() {
         return (thread, throwable) -> logger.error(
-                "Caught unhandled exception on thread '{}'!",
-                thread::getName,
-                () -> ExceptionLoggerDelegate.unwrapThrowable(throwable));
+                "Caught unhandled exception on thread '{}}'!",
```
Bastian, line 44: Was the second `}` intended?

Saladoc: It was not. IntelliJ autocompleted one while I typed the other.

Bastian:
Thank you!

Fyi: The common convention for commit subjects is to not have a period at the end. See [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/). I don't enforce strict commit message rules for Javacord, but I thought you might be interested in it 🙂

PR #567, opened 2020-05-19
Vavr set view

Bastian:
Adds a view for Vavr's sets.

> **Note:** This PR is built upon #556. Only the last commit is relevant.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
@@ -1,0 +1,241 @@
+package org.javacord.core.util;
+
+import io.vavr.collection.HashSet;
+import io.vavr.collection.LinkedHashSet;
+import io.vavr.collection.TreeSet;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * An immutable view that maps a Vavr set to a Java set.
+ *
+ * @param <E> The type of elements maintained by this set.
+ */
+public class VavrSetView<E> implements Set<E> {
+
+    private io.vavr.collection.Set<E> vavrSet;
+    private final boolean unmodifiable;
+
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+
+    /**
+     * Creates a new Vavr set view.
+     *
+     * @param vavrSet The Vavr set.
+     * @param unmodifiable Whether the view should be modifiable or not.
+     */
+    public VavrSetView(io.vavr.collection.Set<E> vavrSet, boolean unmodifiable) {
+        this.vavrSet = vavrSet;
+        this.unmodifiable = unmodifiable;
+    }
+
+    @Override
+    public int size() {
+        lock.readLock().lock();
+        try {
+            return vavrSet.size();
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    @Override
+    public boolean isEmpty() {
+        lock.readLock().lock();
+        try {
+            return vavrSet.isEmpty();
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean contains(Object o) {
+        lock.readLock().lock();
+        try {
+            return vavrSet.contains((E) o);
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p><b>Implementation Note:</b> Modifications to the underlying set do not affect the {@link Iterator#hasNext()}
+     * and {@link Iterator#next()} methods.
+     */
+    @Override
+    public Iterator<E> iterator() {
+        lock.readLock().lock();
+        try {
+            return new VavrSetViewIterator();
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    @Override
+    public Object[] toArray() {
+        lock.readLock().lock();
+        try {
+            return vavrSet.toJavaArray();
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> T[] toArray(T[] a) {
+        lock.readLock().lock();
+        try {
+            if (a.length < size()) {
+                return (T[]) Arrays.copyOf(toArray(), size(), a.getClass());
+            }
+            System.arraycopy(toArray(), 0, a, 0, size());
+            if (a.length > size()) {
+                a[size()] = null;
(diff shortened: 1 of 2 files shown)


javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Set;
+
+/**
+ * An immutable view that maps a Vavr set to a Java set.
+ *
+ * @param <E> The type of elements maintained by this set.
+ */
+public class VavrSetView<E> implements Set<E> {
+
+    private volatile io.vavr.collection.Set<E> vavrSet;
```
Vampire, line 15: I understand your intention with your usage of `volatile` and `synchronized`,
that the read-only methods should not block each other and leverage the `volatile` while the write methods block for read-change-write.
Unfortunately I think this idiom will not work correctly.
Writes should also block reads.

Maybe it is better and also clearer to use a `ReentrantReadWriteLock` instead of `volatile` and `synchronized`.
This will also fix the potential problem that if someone for whatever reason - legit or evilly trying to harm something - synchronizes on the set instance this will block the write methods that cannot obtain the lock.

If I really want to use `synchronized` instead of some higher level API like `ReentrantReadWriteLock` or `Semaphore`, I at least use some internal `private final Object LOCK = new Object();` and synchronize on that instead.
This also allows to keep the synchronized scope as narrow as possible.
In your case it is unnecessary to synchronize for the modifiable check.
Only after that the synchronized block would be necessary, or the obtaining of the write lock from the `ReentrantReadWriteLock` I suggested.

Vampire:
Maybe you also want to PR this set view to the Vavr project when it is finished :-)

Bastian:
> Maybe you also want to PR this set view to the Vavr project when it is finished :-)

I'm not sure if this is what they want. Their current `ListView` implementation is pretty different compared to the view from this PR.
https://github.com/vavr-io/vavr/blob/master/src/main/java/io/vavr/collection/JavaConverters.java

Bastian:
I've pushed two new commits that adress your comments.
I'll squash them into one commit later, but I think the additional commits make it easier to review :)

javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
```diff
@@ -0,0 +1,127 @@
...
+    private final boolean unmodifiable;
+
+    /**
+     * Creates a new Vavr set view.
+     *
+     * @param vavrSet The Vavr set.
+     * @param unmodifiable Whether the view should be modifiable or not.
```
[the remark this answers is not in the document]
Bastian: I've created a custom iterator to support the `remove()` method if the set is modifiable.

javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
```diff
@@ -0,0 +1,127 @@
...
+        throwIfUnmodifiable();
+        io.vavr.collection.Set<E> oldVavrSet = vavrSet;
+        vavrSet = vavrSet.removeAll((Collection<E>) c);
+        return oldVavrSet.size() != vavrSet.size();
+    }
+
+    @Override
+    public synchronized void clear() {
+        throwIfUnmodifiable();
+        vavrSet = vavrSet.removeAll(vavrSet);
```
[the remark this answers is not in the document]
Bastian: While this would probably has a better performance, I specifically wrote it this way to keep the underlying implementation the same (e.g., not change a `TreeSet` to a `HashSet`).

Another possibility would be to use some `instranceof` checks and use the `vavrSet.removeAll(vavrSet);` as a fallback.

Bastian:
And now a thrid commit that improves the `clear()` method.

javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Set;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * An immutable view that maps a Vavr set to a Java set.
+ *
+ * @param <E> The type of elements maintained by this set.
+ */
+@SuppressWarnings("NonAtomicOperationOnVolatileField") // We use locks to synchronize access on the vavrSet field
```
Vampire, line 19: The question is more why you still keep the field volatile.
The read/write locks should already guarantee memory consistency.
So you should remove this and the `volatile` if I'm not completely mistaken.

javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean contains(Object o) {
+        return vavrSet.contains((E) o);
+    }
+
+    @Override
+    public Iterator<E> iterator() {
```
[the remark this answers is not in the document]
Vampire: Did you resolve this accidentally or as "I don't want to do it"? :-)
This has nothing to do with the iterator now supporting removal.

javacord-core/src/main/java/org/javacord/core/util/VavrSetView.java
Bastian: I wasn't aware that high-level locks also provide memory synchronization.
https://stackoverflow.com/questions/12429818/does-explicit-lock-automatically-provide-memory-visibility/12429966#12429966
But looks like you are correct. The volatile keyword is not neccessary here.

Vampire: Just for you to know.
What is written in the documentation linked in that answer is not enough.
There is only described `Lock` per se, so one lock that is used.
The more important question is whether the separate read lock and write lock do guarantee the memory synchronization as pair.
But given there is no bug in the implementation, the docs of `ReadWriteLock` define it as:
> All ReadWriteLock implementations must guarantee that the memory synchronization effects of writeLock operations (as specified in the Lock interface) also hold with respect to the associated readLock. That is, a thread successfully acquiring the read lock will see all updates made upon previous release of the write lock.

Bastian: https://docs.oracle.com/javase/specs/jvms/se6/html/Threads.doc.html#21294
> Less formally: if a thread is to perform an unlock operation on any lock, it must first copy all assigned values in its working memory back out to main memory.

and

> Less formally: a lock operation behaves as if it flushes all variables from the thread's working memory, after which the thread must either assign them itself or load copies anew from main memory.

By that logic, the documentation of `Lock` should already be enough, shouldn't it?
If the cache gets flushed on `monitorEnter` and written back to main memory on `monitorLeave`, it should already ensure the memory synchronization of a read-write lock.

Vampire: Hm, good point.
I'd tend to say "maybe", I'm not an expert in Java memory model yet. :-D

PR #568, opened 2020-05-21
Add a refrence to KTX

Doomsdayrs:
- Adds a reference to KTX extensions for Kotlin
- Puts the word Kotlin in the ReadMe, so ya get all the Kotlin users dropping by

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,13 +88,13 @@ The optional KTX Dependency provides extensions in Kotlin for javacord
 In Gradle
 ```groovy
 repositories { maven { url 'https://jitpack.io' } }
-dependencies { implementation 'com.github.doomsdayrs:javacord-ktx:v0.1.0' }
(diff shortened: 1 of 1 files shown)


Saladoc:
I agree with Vampire 100% here. I'd not even put 3rd party libs anywhere on this repo, but rather, if anything, put a "List of third party modules" on https://javacord.org/wiki somewhere

Bastian:
First of all, thank you for the PR and your work on the Kotlin extensions!

I have to agree with the others, that I don't want to feature your—or any 3rd party library—that prominently in the README.

> I'd not even put 3rd party libs anywhere on this repo

To be fair, we already have two 3rd party command libraries mentioned in the README.

Saladoc:
> To be fair, we already have two 3rd party command libraries mentioned in the README.

Since both of those are made by maintainers of this repository I would not consider them to be third party.

Issue #463, opened 2019-04-09
Add methods to measure/get latency


Bastian:
Measuring the "ping" is a frequently requested feature.
While there is not "the ping", we could add some utility methods:

### `Duration DiscordApi#getGatewayLatency()`

Returns the time it took to receive a heartbeat ack after we sent a heartbeat. See https://discordapp.com/developers/docs/topics/gateway#heartbeating

### `CompletableFuture<Duration> DiscordApi#getRestLatency()`

Performs a REST request that has no impact (we have to look for a good one) and returns the time it took to get an answer.
I'm unsure about the method name because of https://github.com/Javacord/Javacord/issues/377

PR #569, opened 2020-05-24
Add DiscordApi#measureRestLatency()

Bastian:
Partially fixes https://github.com/Javacord/Javacord/issues/463.
A method to measure the REST latency is added in https://github.com/Javacord/Javacord/pull/570.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -34,6 +34,7 @@ import java.awt.image.BufferedImage;
 import java.io.File;
 import java.io.InputStream;
 import java.net.URL;
+import java.time.Duration;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -94,6 +95,30 @@
...
+     * Javacord will need. The latency is influenced by many non-network related factors like your system's current
+     * workload.
+     *
+     * <p>This method uses the {@code GET /users/@me} endpoint to test the latency.
+     * The method ensures that only one ping measurement is performed at once by the same {@code DiscordApi} instance.
+     *
+     * <p><b>Warning</b>: This method does not bypass the global ratelimit check.
+     * If your bot gets ratelimited globally, the latency will appear higher that it really is!
+     *
+     * <p><b>Expected latency</b>: Usually, you can expect a latency between {@code 50ms} and {@code 500ms} with a good
```
Vampire, line 113:
```suggestion
     * <p><b>Expected latency</b>: Usually, you can expect a latency between {@code 50 ms} and {@code 500 ms} with a good
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -94,6 +95,30 @@
...
+     * internet connection. This value may vary, depending on your location, time of day, Discord's status, and the
+     * current workload of your system. A value above {@code 1000ms} is usually an indicator that something is wrong.
```
Vampire, line 115:
```suggestion
     * current workload of your system. A value above {@code 1000 ms} is usually an indicator that something is wrong.
```

PR #570, opened 2020-05-24
Add DiscordApi#getLatestGatewayLatency()

Bastian:
Partially fixes https://github.com/Javacord/Javacord/issues/463.
A method to measure the REST latency is added in https://github.com/Javacord/Javacord/pull/569.

#### Why `Duration` and not `Optional<Duration>`?

In nearly every case, the duration will be available when the `DiscordApi` future finishes. It is just more convenient to now have the value wrapped in an `Optional` when it is not necessary in 99.99% of all cases. Returning `-1` should be enough in this very very rare case.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -95,6 +95,25 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
     Optional<Ratelimiter> getGlobalRatelimiter();
 
     /**
+     * Gets the latest gateway latency.
+     *
+     * <p>To calculate the gateway latency, Javacord measures the time it takes for Discord to answer the gateway
+     * heartbeat packet with a heartbeat ack packet. Please notice, that this values does only get updated on every
+     * heartbeat and not on every method call. To calculate an average, you have to collect the latency over a period of
+     * multiple minutes.
+     *
+     * <p>In very rare cases, the latency will be {@code -1 ns} directly after startup, because the initial heartbeat
+     * was not answered yet. Usually, the heartbeat is answered before the bot finished loading.
+     *
+     * <p><b>Expected latency</b>: Usually, you can expect a latency between {@code 30 ms} and {@code 300 ms} with a
+     * good internet connection. This value may vary, depending on your location, time of day, Discord's status, and the
+     * current workload of your system. A value above {@code 1000 ms} is usually an indicator that something is wrong.
+     *
+     * @return The latest measured gateway latency.
+     */
+    Duration getLatestGatewayLatency();
+
+    /**
      * Measures, how long Javacord will need to perform a single REST call.
      *
      * <p>This method does not measure the "true" latency to Discord's REST endpoints because the request is handled by
--- a/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
@@ -208,6 +208,11 @@ public class DiscordApiImpl implements DiscordApi, DispatchQueueSelector {
     private final boolean waitForServersOnStartup;
 
     /**
+     * The latest gateway latency.
+     */
+    private volatile long latestGatewayLatencyNanos = -1;
+
+    /**
      * A lock that makes sure that there are not more than one ping attempt at the same time.
      * This ensures that the ping does not get affected by other requests that use the same bucket.
      */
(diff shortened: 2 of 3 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -87,6 +88,25 @@
...
 
+    /**
+     * Gets the latest gateway latency.
+     *
+     * <p>To calculate the gateway latency, Javacord measures the time it takes for Discord to answer the gateway
+     * heartbeat packet with a heartbeat ack packet. Please notice, that this values does only get updated on every
+     * heartbeat and not on every method call. To calculate an average, you have to collect the latency over a period of
+     * multiple minutes.
+     *
+     * <p>In very rare cases, the latency will be {@code -1ns} directly after startup, because the initial heartbeat was
```
Vampire, line 99: Definitely a nit, but you should put a space between number and unit.
It reads better and follows the SI standard. :-)

```suggestion
     * <p>In very rare cases, the latency will be {@code -1 ns} directly after startup, because the initial heartbeat was
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -87,6 +88,25 @@
...
+     * not answered yet. Usually, the heartbeat is answered before the bot finished loading.
+     *
+     * <p><b>Expected latency</b>: Usually, you can expect a latency between {@code 30ms} and {@code 300ms} with a good
```
Vampire, line 102:
```suggestion
     * <p><b>Expected latency</b>: Usually, you can expect a latency between {@code 30 ms} and {@code 300 ms} with a good
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -87,6 +88,25 @@
...
+     * internet connection. This value may vary, depending on your location, time of day, Discord's status, and the
+     * current workload of your system. A value above {@code 1000ms} is usually an indicator that something is wrong.
```
Vampire, line 104:
```suggestion
     * current workload of your system. A value above {@code 1000 ms} is usually an indicator that something is wrong.
```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -194,6 +196,11 @@
      */
     private final boolean waitForServersOnStartup;
 
+    /**
+     * The last gateway latency.
+     */
+    private final AtomicLong lastGatewayLatencyNanos = new AtomicLong(-1);
```
Vampire, line 202: Why `AtomicLong` here instead of `volatile long`?
As far as I have seen from a quick look, you don't use any of the `AtomicLong` benefits like atomic read and write in `updateAndGet` and so on.

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -774,6 +781,15 @@ private void removeUnavailableServerFromCache(long serverId) {
         unavailableServers.remove(serverId);
     }
 
+    /**
+     * Gets the last gateway latency in nano seconds.
+     *
+     * @return The last gateway latency.
+     */
+    public AtomicLong getLastGatewayLatencyNanos() {
```
Vampire, line 789: `getLatestGatewayLatency` vs. `getLastGatewayLatencyNanos` (`last` vs. `latest` and also in JavaDoc)

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -130,6 +131,9 @@
     private final AtomicReference<Future<?>> heartbeatTimer = new AtomicReference<>();
     private final AtomicBoolean heartbeatAckReceived = new AtomicBoolean();
 
+    // Used to calculate the gateway latency
+    private final AtomicLong lastHeartbeatSentTimeNanos = new AtomicLong();
```
Vampire, line 135: Same question as for the other `AtomicLong` :-)

Vampire:
And the commit message seems to be wrong

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
Bastian: I don't need a setter 😏

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -809,6 +814,24 @@ private void removeUnavailableServerFromCache(long serverId) {
...
+     * Gets the latest gateway latency in nano seconds.
+     *
+     * @return The latest gateway latency.
+     */
+    public long getLatestGatewayLatencyNanos() {
+        return latestGatewayLatencyNanos;
+    }
+
+    /**
+     * Gets the latest gateway latency in nano seconds.
```
Vampire, line 827: Nope :-)

Bastian:
Ups, I thought you were talking about the `Last` vs `Latest` in the commit message. I didn't even notice the `Nanos`... :see_no_evil:

Issue #422, opened 2019-01-16
Send messages directly to webhooks


Bastian:
It should be possible to send messages to webhooks with Javacord.

- The `Webhook` class should extend `Messageable`
- The `MessageBuilder#send()` method should support webhook urls (maybe `sendToWebhook()` to make it more clear?) and of course `Webhook` also objects
- `MessageBuilder#setEmbed(...)` should be replaced with `#addEmbed(...)` (deprecate, not remove `#setEmbed()`!)
- Webhooks should be properly ratelimited with automatic retries -> I don't know if the ratelimit headers are the same. Hopefully we can just use our normal `RestRequest` class without any adjustments.

Bastian:
Webhooks also allow to set custom display names and avatar when sending the message.
So we'll also need `MessageBuilder#setWebhookDisplayName(...)` and `#setWebhookAvatar(...)`.
I would like to have the webhook in the name, because otherwise it might be confusing if you don't know that `MessageBuilder` can also be used for webhooks.

Saladoc:
Maybe it would be better to have a separate `WebhookMessageBuilder` with those functions, have it extend `MessageBuilder` and have a `MessageBuilder#forWebsocket` function that basically calls a copy constructor on the subclass. 
That way people need to explicitly create a `WebhookMessageBuilder` or convert a `MessageBuilder` and every risk of confusion is gone.

Bastian:
With the introduction of slash commands, this feature became significantly more important.

PR #573, opened 2020-06-06
Sending messages to webhooks.

Joshix-1:
Like said in #422. It should work fine.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
@@ -483,9 +483,8 @@ public class MessageBuilder {
      *
      * @param webhook The webhook from which the message should be sent.
      * @return The sent message.
-     * @throws IllegalStateException if the defined Webhook's channel is not present.
      */
-    public CompletableFuture<Message> send(IncomingWebhook webhook) throws IllegalStateException {
+    public CompletableFuture<Message> send(IncomingWebhook webhook) {
         return delegate.send(webhook);
     }
 
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
@@ -122,7 +122,8 @@ public class MessageBuilder {
      * @return The current instance in order to chain call methods.
      */
     public MessageBuilder setEmbed(EmbedBuilder embed) {
-        delegate.setEmbed(embed);
+        delegate.removeAllEmbeds();
+        delegate.addEmbed(embed);
         return this;
     }
 
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
@@ -1,5 +1,6 @@
 package org.javacord.api.entity.message;
 
+import org.javacord.api.DiscordApi;
 import org.javacord.api.entity.Icon;
 import org.javacord.api.entity.Mentionable;
 import org.javacord.api.entity.channel.TextChannel;
@@ -8,6 +9,7 @@ import org.javacord.api.entity.message.internal.MessageBuilderDelegate;
 import org.javacord.api.entity.message.mention.AllowedMentions;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.entity.webhook.IncomingWebhook;
+import org.javacord.api.util.DiscordRegexPattern;
 import org.javacord.api.util.internal.DelegateFactory;
 
 import java.awt.image.BufferedImage;
@@ -15,6 +17,7 @@ import java.io.File;
 import java.io.InputStream;
 import java.net.URL;
 import java.util.concurrent.CompletableFuture;
+import java.util.regex.Matcher;
 
 /**
  * This class can help you to create messages.
@@ -496,6 +499,50 @@ public class MessageBuilder {
      */
     public CompletableFuture<Message> send(Messageable messageable) {
         return delegate.send(messageable);
+    }
+
+    /**
+     * Sends the message.
+     *
+     * @param api The api instance needed to send and return the message.
+     * @param webhookId The id of the webhook from which the message should be sent.
+     * @param webhookToken The token of the webhook from which the message should be sent.
+     * @return The sent message.
+     */
+    public CompletableFuture<Message> sendWithWebhook(DiscordApi api, long webhookId, String webhookToken) {
+        return delegate.sendWithWebhook(api, Long.toUnsignedString(webhookId), webhookToken);
+    }
+
+    /**
+     * Sends the message.
+     *
+     * @param api The api instance needed to send and return the message.
+     * @param webhookId The id of the webhook from which the message should be sent.
+     * @param webhookToken The token of the webhook from which the message should be sent.
+     * @return The sent message.
+     */
+    public CompletableFuture<Message> sendWithWebhook(DiscordApi api, String webhookId, String webhookToken) {
+        return delegate.sendWithWebhook(api, webhookId, webhookToken);
+    }
+
+    /**
+     * Sends the message.
+     *
+     * @param api The api instance needed to send the message.
+     * @param webhookUrl The url of the webhook from which the message should be sent.
+     *
+     * @return The sent message.
+     * @throws IllegalArgumentException If the link isn't valid.
+     */
+    public CompletableFuture<Message> sendWithWebhook(DiscordApi api, String webhookUrl)
+                                                        throws IllegalArgumentException {
+        Matcher matcher = DiscordRegexPattern.WEBHOOK_URL.matcher(webhookUrl);
+
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException("The webhook url has an invalid format");
+        }
+
+        return sendWithWebhook(api, matcher.group("id"), matcher.group("token"));
     }
 
 }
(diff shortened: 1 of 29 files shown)


Joshix-1:
I improved the JavaDoc comments a bit, but I don't know if that is enough.

Joshix-1:
I just realised a problem. The messages sent to a IncomingWebhook won't be send in order. 
I do this:
```Java
for (int i = 0; i < 20; i++) {
    webhook.sendMessage(i + ".");
}
```
And I get this result: 
[https://media.discordapp.net/attachments/701182154180395091/725262809033605120/unknown.png](image)

When I do the same with a channel instead of a webhook I get this: [https://cdn.discordapp.com/attachments/701182154180395091/725263609500008468/unknown.png](image)

But I don't know why this happens.

javacord-core/src/main/java/org/javacord/core/entity/webhook/WebhookImpl.java
```diff
@@ -42,24 +45,39 @@
...
      * Creates a new webhook.
      *
      * @param api The discord api instance.
      * @param data The json data of the webhook.
      */
-    public WebhookImpl(DiscordApi api, JsonNode data) {
+    protected WebhookImpl(DiscordApi api, JsonNode data) {
         this.api = (DiscordApiImpl) api;
 
+        this.type = data.get("type").asText().equals("1") ? WebhookType.INCOMING : WebhookType.CHANNEL_FOLLOWER;
```
Pepich, line 59: Move this into the WebhookType enum as a helper method for easier maintainability? E.g. WebhookType.of(int code)

javacord-api/src/main/java/org/javacord/api/entity/message/internal/MessageBuilderDelegate.java
```diff
@@ -69,11 +70,16 @@
     void setContent(String content);
 
     /**
-     * Sets the embed of the message.
+     * Adds the embed to the message.
```
Pepich, line 73: I'm not sure if I mentioned this in a different place already but maybe update the comment to mention that only webhooks support more than one embed?

javacord-core/src/main/java/org/javacord/core/entity/message/WebhookMessageBuilderDelegateImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public void removeEmbed(EmbedBuilder embed) {
+        this.embeds.remove(embed);
+    }
+
+    @Override
+    public void removeEmbeds(EmbedBuilder[] embeds) {
+        this.embeds.removeAll(Arrays.asList(embeds));
+    }
+
```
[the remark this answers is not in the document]
Joshix-1: As this extends MessageBuilderDelegateImpl this shouldn't be necessary

javacord-api/src/main/java/org/javacord/api/entity/message/internal/WebhookMessageBuilderDelegate.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    void removeEmbed(EmbedBuilder embed);
+
+    /**
+     * Removes the embeds from the message.
+     *
+     * @param embeds The embeds to remove.
+     */
+    void removeEmbeds(EmbedBuilder[] embeds);
+
```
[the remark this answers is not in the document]
Joshix-1: As this extends MessageBuilderDelegate this shouldn't be necessary

javacord-core/src/main/java/org/javacord/core/entity/webhook/WebhookImpl.java
Joshix-1: I don't think that that is necessary.

javacord-core/src/main/java/org/javacord/core/entity/webhook/IncomingWebhookImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public boolean isIncomingWebhook() {
+        return true;
+    }
+
+    @Override
+    public boolean isChannelFollowerWebhook() {
+        return false;
+    }
```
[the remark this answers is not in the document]
Joshix-1: Yes, but I thought this would be faster. I don't know.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -121,7 +122,8 @@ public MessageBuilder setContent(String content) {
      * @return The current instance in order to chain call methods.
      */
     public MessageBuilder setEmbed(EmbedBuilder embed) {
```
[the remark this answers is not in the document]
Joshix-1: if I do that the WebhookMessageBuilderDelegate would have the method setEmed too and I don't think that is better.

javacord-api/src/main/java/org/javacord/api/entity/message/internal/MessageBuilderDelegate.java
Joshix-1: I don't think this is the right place to mention this.

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
```diff
@@ -0,0 +1,653 @@
...
+
+        return send(api, matcher.group("id"), matcher.group("token"));
+    }
+
+    /**
+     * Sends the message without waiting for a response.
+     *
+     * @param webhook The webhook from which the message should be sent.
+     */
+    public void sendSilently(IncomingWebhook webhook) {
```
Bastian, line 611: This should return a `CompletableFuture<Void>` instead.
Setting `wait = false` just means that Discord does not answer with a Message and that the webhook execution might have failed. But you might still want to catch other exceptions (e.g., network issues) or wait for the request to finish.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
```diff
@@ -50,9 +61,9 @@
     private final StringBuilder strBuilder = new StringBuilder();
 
     /**
-     * The embed of the message. Might be <code>null</code>.
+     * The list of embed's of the message.
```
Bastian, line 64:
```suggestion
     * The list of embeds of the message.
```

javacord-api/src/main/java/org/javacord/api/entity/webhook/WebhookType.java
```diff
@@ -0,0 +1,15 @@
...
+     * used with Channel Following to post new messages into channels.
+     */
+    CHANNEL_FOLLOWER,
+
+    /**
+     * Incoming webhooks can post messages to channels with a generated token.
+     * It extents {@link org.javacord.api.entity.message.Messageable}.
+     */
+    INCOMING;
+}
```
Bastian, line 15: This should contain the type as a value.
```suggestion
public enum WebhookType {

    /**
     * Incoming webhooks can post messages to channels with a generated token.
     * It extents {@link org.javacord.api.entity.message.Messageable}.
     */
    INCOMING(1),
    
    /**
     * Channel follower webhooks are internal webhooks
     * used with Channel Following to post new messages into channels.
     */
    CHANNEL_FOLLOWER(2),

    UNKNOWN(-1);
    
    private final int value;

    WebhookType(int value) {
        this.value = value;
    }

    /**
     * Gets integer value that represents this type.
     *
     * @return The integer value that represents this type.
     */
    public int getValue() {
        return value;
    }

    /**
     * Gets an {@code WebhookType} by its value.
     *
     * @param value The value of the webhook type.
     * @return The webhooktype for the given value,
     *         or {@link WebhookType#UNKNOWN} if there's none with the given value.
     */
    public static WebhookType fromValue(int value) {
        for (InteractionType type : values()) {
            if (type.getValue() == value) {
                return type;
            }
        }
        return UNKNOWN;
    }
}
```

javacord-core/src/main/java/org/javacord/core/entity/webhook/WebhookImpl.java
Bastian: I agree with Pepich. That's how we have it for every other enum as well.

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1405,7 +1407,22 @@ public void selfUndeafen() {
...
+    }
+
+    @Override
+    public CompletableFuture<List<IncomingWebhook>> getIncomingWebhooks() {
+        return new RestRequest<List<IncomingWebhook>>(getApi(), RestMethod.GET, RestEndpoint.SERVER_WEBHOOK)
+                .setUrlParameters(getIdAsString())
+                .execute(result -> {
+                    List<IncomingWebhook> webhooks = new ArrayList<>();
+                    for (JsonNode webhookJson : result.getJsonBody()) {
+                        if (webhookJson.get("type").asText().equals("1")) {
```
Bastian, line 1423: Instead of having a [magic value](https://en.wikipedia.org/wiki/Magic_number_(programming)), it should use the `WebhookType` enum.
```suggestion
                        if (WebhookType.fromValue(webhookJson.get("type").asInt()) == WebhookType.INCOMING) {
```

Bastian (commented):
Besides a few minor things, it looks good to me :)

Sorry for the late review.

javacord-core/src/main/java/org/javacord/core/entity/message/WebhookMessageBuilderDelegateImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        send(webhookId, webhookToken, displayName, avatarUrl, false, api).whenComplete((message, throwable) -> {
+            if (throwable != null) {
+                future.completeExceptionally(throwable);
+            } else {
+                future.complete(null);
+            }
+        });
+        return future;
+    }
```
Bastian, line 92:
```suggestion
        return send(webhookId, webhookToken, displayName, avatarUrl, false, api)
                .thenApply(m -> null);
```

PR #576, opened 2020-06-12
Remove Client login support

### LINKED ISSUE #540: Remove support for non-bot users (selfbots/client accounts)

Saladoc:
Fixes #540


Bastian:
Do we want to deprecate it first or just remove it as it is pretty much broken anyway?

Without client accounts, we can most likely also remove group channels. Or is there any way for bots to interact with them?

Saladoc:
I could make a deprecation PR too and we could hold this off until 3.1.0, but since it's not a concious design decision but rather a reality brought to us by discord, I'd not bother with deprecation.

Saladoc:
@Bastian How do you want to proceed here?

Bastian (approved):
I agree that a deprecation is not neccessary.

Saladoc:
Since #847 will also do a cleanup with regards to deprecated methods, I added this PR to the same milestone. Since the functionality behind the user account methods has long been discontinued, prior deprecation is not necessary.

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -149,6 +147,11 @@
      */
     private static final Map<String, Ratelimiter> defaultGlobalRatelimiter = new ConcurrentHashMap<>();
 
+    /**
+     * The key prefix for bot accounts.
+     */
+    private static final String BOT_TOKEN_PREFIX = "Bot ";
+
```
KILLEliteMaste, line 154: Seems redundant for me to have it as a variable. Should be directly concated in the `getPrefixedToken` method

javacord-core/src/main/java/org/javacord/core/util/handler/ReadyHandler.java
```diff
@@ -52,7 +51,7 @@ public void handle(JsonNode packet) {
                         new PrivateChannelImpl(api, channelJson);
                         break;
                     case GROUP_CHANNEL:
-                        new GroupChannelImpl(api, channelJson);
+                        logger.warn("Received a group channel which should be unavailable for bots.");
                         break;
                     default:
                         logger.warn("Unknown or unexpected channel type. Your Javacord version might be out of date!");
```
KILLEliteMaste, line 57: Can't select everything (up to the if statement) but this now seems to be code that will never be reached. Therefore this should be removed

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelDeleteHandler.java
```diff
@@ -56,7 +51,7 @@ public void handle(JsonNode packet) {
                 handleServerStageVoiceChannel(packet);
                 break;
             case GROUP_CHANNEL:
-                handleGroupChannel(packet);
+                logger.info("Received an update for a group channel. This should be impossible.");
```
KILLEliteMaste, line 54:
```suggestion
                logger.info("Received CHANNEL_DELETE packet for a group channel. This should be impossible.");
```

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -100,7 +96,7 @@ public void handle(JsonNode packet) {
                 handleServerStageVoiceChannel(packet);
                 break;
             case GROUP_CHANNEL:
-                handleGroupChannel(packet);
+                logger.info("Received an update for a group channel. This should be impossible.");
```
KILLEliteMaste, line 99:
```suggestion
                logger.info("Received CHANNEL_UPDATE packet for a group channel. This should be impossible.");
```

javacord-core/src/main/java/org/javacord/core/util/handler/guild/VoiceStateUpdateHandler.java
```diff
@@ -63,8 +62,8 @@ public void handle(JsonNode packet) {
                 VoiceChannel voiceChannel = optionalChannel.get();
                 if (voiceChannel instanceof PrivateChannel) {
                     handlePrivateChannel(userId, ((PrivateChannelImpl) voiceChannel));
-                } else if (voiceChannel instanceof GroupChannel) {
-                    handleGroupChannel(userId, ((GroupChannelImpl) voiceChannel));
+                } else if (voiceChannel.getType() == ChannelType.GROUP_CHANNEL) {
+                    logger.info("Received an update for a group channel. This should be impossible.");
```
KILLEliteMaste, line 66:
```suggestion
                    logger.info("Received a VOICE_STATE_UPDATE packet for a group channel. This should be impossible.");
```

KILLEliteMaste (changes requested):
When searching for `group` in the project there are still some things left.
- CertainMessageEvent#isGroupMessage
- In MessageAuthor and VoiceChannel , TextChannel, Channel some Javadocs contain `(private channel or group channel)` which can be removed. In Addition VoiceChannel also check for group channels sometimes
- DiscordApi some Javadocs contain `Can be <code>null</code> to search for group channels without name` which can be removed

javacord-core/src/main/java/org/javacord/core/util/handler/ReadyHandler.java
Saladoc: Same goes for the ChannelCreate/Delete/Update Handlers, but we still keep the informative log messages. Only dead code here is the GROUP_CHANNEL case as bots still have private message capabilities.

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
Saladoc: I'm in favor of keeping magic values in constants and not directly in the code. I see no reason to inline this.

javacord-core/src/main/java/org/javacord/core/util/handler/ReadyHandler.java
KILLEliteMaste: According to the mentioned issue above the code the `private_channels` array is always empty for bots. And it has also been completely removed from the docs on the READY event so it makes no sense to keep it

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -100,7 +97,7 @@ public void handle(JsonNode packet) {
                 handleServerStageVoiceChannel(packet);
                 break;
             case GROUP_CHANNEL:
-                handleGroupChannel(packet);
+                logger.info("Received CHANNEL_UPDATE packet  for a group channel. This should be impossible.");
```
KILLEliteMaste, line 100:
```suggestion
                logger.info("Received CHANNEL_UPDATE packet for a group channel. This should be impossible.");
```

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -1583,27 +1562,23 @@ default String makeMentionsReadable(String content) {
...
-     * @param name The name of the channels. Can be <code>null</code> to search for group channels without name.
+     * @param name The name of the channels.
      * @return A collection with all channels with the given name.
      */
     default Collection<Channel> getChannelsByName(String name) {
-        Collection<Channel> channels = new HashSet<>();
-        channels.addAll(getServerChannelsByName(name));
-        channels.addAll(getGroupChannelsByName(name));
+        Collection<Channel> channels = new HashSet<>(getServerChannelsByName(name));
         return Collections.unmodifiableCollection(channels);
```
KILLEliteMaste, line 1570: Here and in the methods below it is not necessary to add the channels to a collection before we return an unmodifiable collection. So here `Collections.unmodifiableCollection(getServerChannelsByName(name));` should be sufficient.

javacord-api/src/main/java/org/javacord/api/entity/channel/Channel.java
```diff
@@ -134,8 +125,7 @@
 
     /**
      * Checks if the given user can see this channel.
-     * In private chats (private channel or group channel) this always returns <code>true</code> if the user is
-     * part of the chat.
+     * In private channels this always returns <code>true</code> if the user is  part of the chat.
```
KILLEliteMaste, line 128:
```suggestion
     * In private channels this always returns <code>true</code> if the user is part of the chat.
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageAuthor.java
```diff
@@ -476,8 +469,8 @@ default boolean canAddNewReactionsInTextChannel() {
 
     /**
      * Checks if the author can manage messages (delete or pin them or remove reactions of others) in the channel
-     * where the message was sent. In private chats (private channel or group channel) this always returns {@code true}
-     * if the user is part of the chat.
+     * where the message was sent.
+     * In private channels this always returns {@code true}  if the user is part of the chat.
```
KILLEliteMaste, line 473:
```suggestion
     * In private channels this always returns {@code true} if the user is part of the chat.
```

KILLEliteMaste (changes requested):
Just a few minor things :)

KILLEliteMaste (approved):
LGTM!
Please rebase on the latest version and feel free to merge it right away :)

PR #585, opened 2020-07-18
BoostLevel TIER_3 Id Bugfix

AyuAi:
Tier 3 ID was set to "2" when it should be "3"


Bastian (approved):
Oops...
Thanks for the PR! 🚀

PR #586, opened 2020-07-21
Audio Reconnecting

Mysterypotatoguy:
🎉

This is a mostly complete audio reconnect implementation, It would be appreciated if I could get people to test this to make sure the disconnect codes are being handled properly.

A few thoughts:
Should there be a limit on the number of reconnect retries?
Should the UDP sending thread be paused during disconnects?

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
@@ -188,7 +188,7 @@ public class AudioUdpSocket {
     }
 
     /**
-     * Starts polling frames from the audio connection.
+     * Stops polling frames from the audio connection.
      */
     public void stopSending() {
         shouldSend = false;
--- a/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioWebSocketAdapter.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioWebSocketAdapter.java
@@ -60,7 +60,7 @@ public class AudioWebSocketAdapter extends WebSocketAdapter {
     /**
      * Whether a {@link VoiceGatewayOpcode#RESUME} should be sent or a normal connect.
      */
-    private volatile boolean sendResume;
+    private volatile boolean resuming;
 
     // A reconnect attempt counter
     private final AtomicInteger reconnectAttempt = new AtomicInteger();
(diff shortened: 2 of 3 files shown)


Bastian:
I've tried to test it by disconnecting the internet of my PC. The reconnect failed with this debug log: https://gist.github.com/Bastian/90b069ce7452e338fee93b25e467228d

In Discord the bot successfully reconnected the voice channel, just to disconnect a second later. I guess the important part is this line in the log: https://gist.github.com/Bastian/90b069ce7452e338fee93b25e467228d#file-audio-reconnect-log-L1543
```
2020-09-10 19:22:22.634+0200 INFO org.javacord.core.util.gateway.AudioWebSocketAdapter Websocket closed with reason 'Disconnected.' and code DISCONNECTED (4014) by server for AudioConnection (channel: ServerVoiceChannel (id: 360712737778827265, name: testing), sessionId: 10fa7b5d6859e0d7d7b0496dcbfd361d, endpoint: eu-central4803.discord.media:80)! {shard=0} 
````

Do you know, when the `DISCONNECTED` close code is used by Discord?

Besides that, the code looks good to me. Unfortunately, errors like crashes on Discord's end and similar problems are hard to test.

**Comment by Mysterypotatoguy** [comment]
I believe we can stop the debug spam by pausing UDP sending during disconnects (and catching WebSocketExceptions), however I'm unsure whether it should skip over audio frames or stop playing entirely


`DISCONNECTED` is for when the channel you are in is deleted or you are kicked from the channel (according to their docs), I'm not sure why it would send in a case like this.
https://discord.com/developers/docs/topics/opcodes-and-status-codes#voice-voice-close-event-codes

On testing this myself I received a `4006` close code:
```
2020-09-10 19:05:19.566+0100 INFO org.javacord.core.util.gateway.AudioWebSocketAdapter Websocket closed with reason 'Session is no longer valid.' and code 4006 by server for AudioConnection (...)
```
This makes more sense as I left the network disconnected for a little while
I did notice that the `Heart` of the connection was not sending the correct close code as it was marked `Usage.NORMAL` so expect a fix for that.

Bastian:
> `DISCONNECTED ` is for when the channel you are in is deleted or you are kicked from the channel (according to their docs), I'm not sure why it would send in a case like this.

Strange...

> This makes more sense as I left the network disconnected for a little while

I only left it disconnected for a short time.

Bastian:
Thank you!
I still get some strange behavior, when I disconnect my internet. But this might be a bug in Javacord (we receive `4003 Not authenticated` on the main websocket). This might be related to the fact that the audio branch is far behind the development branch. I will merge this commit and then test again.

PR #587, opened 2020-07-23
Fix and cleanup of Allowed Mentions

Mysterypotatoguy:
A fix to prevent a NullPointerException on trying to use an empty allowed mentions and an incorrect sublist range, as well as some tweaks to make things a bit cleaner.

Unsure whether getAllowedRoleMentions/getAllowedUserMentions should return the user provided allowed mentions with a warning in javadoc, or the effective mentions based on whether or not those mention types are allowed (In which case the Optional<List> would probably be better)

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/mention/AllowedMentions.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/mention/AllowedMentions.java
@@ -1,7 +1,7 @@
 package org.javacord.api.entity.message.mention;
 
+import java.util.EnumSet;
 import java.util.List;
-import java.util.Optional;
 
 /**
  * This interface represents a mention.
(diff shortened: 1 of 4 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/mention/AllowedMentions.java
```diff
@@ -1,32 +1,36 @@
...
 
 /**
  * This interface represents a mention.
  */
 public interface AllowedMentions {
 
     /**
-     * Gets the allowed user mentions from the message.
+     * Gets the explicitly allowed user mentions from the message.
+     * This could differ from the actual mentioned users if {@link AllowedMentionsBuilder#setMentionRoles(boolean)} has
```
Joshix-1, line 13: Couldn't this also differ if no allowed mentions are set?

javacord-api/src/main/java/org/javacord/api/entity/message/mention/AllowedMentions.java
```diff
@@ -1,32 +1,36 @@
...
-     * @return The allowed mentions for users of the message.
+     * @return The explicitly allowed mentions for users of the message.
      */
-    Optional<List<Long>> getAllowedRoleMentions();
+    List<Long> getAllowedRoleMentions();
 
     /**
-     * Gets the allowed role mentions from the message.
+     * Gets the explicitly allowed role mentions from the message.
+     * This could differ from the actual mentioned users if {@link AllowedMentionsBuilder#setMentionUsers(boolean)} has
```
Joshix-1, line 22: Couldn't this also differ if no allowed mentions are set?

Bastian (approved):

PR #590, opened 2020-08-08
#getReadableContent() Bugfix

AyuAi:
Fix for Illegal group reference exception when a username or a role name has a non-regex friendly character (like $'s)


Bastian (approved):

Bastian:
Thank you!

Just as a tip/suggestion for the future:
image

Better put the information in the commit message's subject and not in the body.
A commit message like
```
Fix for Illegal group reference exception

Happens when a username or a role name has a non-regex friendly character (like $'s)
```
is much better than
```
bugfix

Fix for Illegal group reference exception when a username or a role name has a non-regex friendly character (like $'s)
```

CLS-creator:
Unsubscribr

PR #594, opened 2020-08-21
Fix can manage role

cap5lut:
`Server#canManageRole(User, Role)` doesnt work currently if you pass the highest role the user has.


Bastian (approved):

Bastian:
Thanks! 👍

Issue #596, opened 2020-08-29
Reaction Remove Handler incorrectly assumes that users are cached


Pepich:
In the [MessageReactionRemoveHandler](https://github.com/Javacord/Javacord/blob/029384adbfb7a1ea8a15870ef549008e68fe8096/javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionRemoveHandler.java#L37), Javacord assumes that a user removing a reaction must be cached. However given the following scenario, an uncached user can remove a reaction:

- User "user" reacts to a message
- "user" leaves the guild in which the reaction happened
- "user" is removed from the cache by whichever means, e.g. restarting the bot will force this to happen
- A different user "admin" who has "manage message" permissions removes the reaction made by "user"

Discord will now fire a "ReactionRemoveEvent", with the "user_id" field containing the id of "user" and not "admin", resulting in an AssertionError

Bastian:
Thank you for reporting.
Our next but one release will add support for disabling the user cache completely, which will require a larger rewrite.
This will most likely fix this bug as a by-product.

PR #597, opened 2020-08-29
fix for user not having to be in cache in ReactionRemoveEvent

Joshix-1:
like said in #596


Bastian:
Thank you for the PR!

Unfortunately, I cannot merge this PR for the following reason: Our next but one release will include a complete rework of the user cache, which will fix this bug as a by-product. I could merge the PR to include it in the next release, but I don't want it to contain breaking changes. Because of this, the PR is in the unfortunate position that it cannot be included in the next release and will be obsolete in the next but one release. Sorry :(

PR #611, opened 2020-09-20
Switch to the "_new" fields

KILLEliteMaste:


javacord-core/src/main/java/org/javacord/core/entity/channel/ServerChannelImpl.java
```diff
@@ -85,9 +85,9 @@
     /**
      * Creates a new server channel object.
      *
-     * @param api The discord api instance.
+     * @param api    The discord api instance.
```
Joshix-1, line 88: There are 4 spaces after the api

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerChannelImpl.java
```diff
@@ -85,9 +85,9 @@
...
      * @param server The server of the channel.
-     * @param data The json data of the channel.
+     * @param data   The json data of the channel.
```
Joshix-1, line 90: There are 3 spaces after data.

javacord-core/src/main/java/org/javacord/core/entity/permission/PermissionsImpl.java
```diff
@@ -14,29 +16,29 @@
...
+    private final long allowed;
+    private final long denied;
 
     /**
      * Creates a new instance of this class.
      *
-     * @param allow An int containing all allowed permission types.
-     * @param deny An int containing all denied permission types.
+     * @param allow A long containing all allowed permission types.
+     * @param deny  A long containing all denied permission types.
```
Joshix-1, line 26: There are 2 spaces after the deny.

javacord-core/src/main/java/org/javacord/core/entity/permission/RoleImpl.java
```diff
@@ -98,9 +98,9 @@
     /**
      * Creates a new role object.
      *
-     * @param api The discord api instance.
+     * @param api    The discord api instance.
```
Joshix-1, line 101: too many spaces after api

javacord-core/src/main/java/org/javacord/core/entity/permission/RoleImpl.java
```diff
@@ -98,9 +98,9 @@
...
      * @param server The server of the role.
-     * @param data The json data of the role.
+     * @param data   The json data of the role.
```
Joshix-1, line 103: too many spaces after data

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -376,10 +376,10 @@ private void handleGroupChannel(JsonNode jsonChannel) {
     /**
      * Dispatches a ServerChannelChangeOverwrittenPermissionsEvent.
      *
-     * @param channel The channel of the event.
+     * @param channel        The channel of the event.
```
Joshix-1, line 379: why the spaces?

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -376,10 +376,10 @@ private void handleGroupChannel(JsonNode jsonChannel) {
...
      * @param newPermissions The new overwritten permissions.
      * @param oldPermissions The old overwritten permissions.
-     * @param entity The entity of the event.
+     * @param entity         The entity of the event.
```
Joshix-1, line 382: spaces

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
KILLEliteMaste: Auto format. It aligns all descriptions below each other rather than next to the parameter name. 
If this is not wanted I can revert it.

Joshix-1: I don't think this is wanted, it is aligned to the parameter name everywhere.

Bastian: Javacord is a bit inconsistent with this. Honestly, this is something I don't care about very much. 
Personally, I prefer to not align the description (i.e., a single space behind the parameter).

KILLEliteMaste: Checkstyle didn't yell at me so I took this as it was. Maybe it would be helpful if there was a check for that in checkstyle or even better if you provide your IDE format configuration, so everyone uses the same for this project.

Bastian:
The `_new` suffix does no longer exist in v8 afaik?

KILLEliteMaste:
> The `_new` suffix does no longer exist in v8 afaik?

This is correct. But I think we still have to convert the type to long of the permissions to respect the new size. But maybe https://github.com/Javacord/Javacord/pull/679 can implement it right away

Bastian:
Agreed, we should switch to `long`.
I think it's easier if you just remove the `_new` suffix from your PR and I merge it after #679.

Joshix-1:
This should get high priority as it is necessary for switching to v8.

KILLEliteMaste:
This is now ready to be merged if everything is alright

Joshix-1 (approved):
lgtm

PR #619, opened 2020-09-25
Allow custom gateway identify ratelimiters

coaxial1:
This PR introduces a new mechanic to control the 5 second gateway identify ratelimit.

Currently, Javacord has a hardcoded mechanic in place that only works within a single JVM.
This is a problem for larger bots that might need to split up their shards into multiple JVMs as they might run into identify ratelimit errors if they don't sync this ratelimit across all shards.

The old mechanic has been replaced by re-using the _Ratelimiter_ interface that is already used for the global 50/1 REST ratelimiter, which can already be replaced with a custom implementation. 

By default, Javacord will use one Ratelimiter per token and JVM (just like it has been done for the 50/1 default global ratelimiter) that allows 1 identify request per 5500ms. Users can supply some sort of network/IPC based implementation to the _DiscordApiBuilder_ to ensure that all shards respect the same ratelimit.

@Bastian: Not sure if this is the way to go. The Discord dev docs are inconsistent about wether you have to delay the IDENTIFY payload or the connection itself (both is mentioned in different locations: https://discord.com/developers/docs/topics/gateway#rate-limiting vs. https://discord.com/developers/docs/topics/gateway#identifying-example-gateway-identify). Not sure if it would be okay to connect to the gateway and then, in the worst case, wait 5 minutes (e.g. Yunite has 60 shards = 5 minutes) until sending the IDENTIFY packet. If that's allowed, the ratelimiter could work more accurately as we could request the quota right before sending the already prepared identify packet. Right now, it's waiting before connecting the websocket, without taking the delay between connecting and identifying into account - that's why I arbitrarily added 500ms just in case.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -115,6 +115,17 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
     Optional<Ratelimiter> getGlobalRatelimiter();
 
     /**
+     * Gets the current gateway identify ratelimiter.
+     *
+     * <p>If you did not provide a ratelimiter yourself, this method will return a {@link LocalRatelimiter}
+     * which is set to allow one gateway identify request per 5500ms and is shared with every bot with the same token
+     * in the same Java program.
+     *
+     * @return The current gateway identify ratelimiter.
+     */
+    Ratelimiter getGatewayIdentifyRatelimiter();
+
+    /**
      * Gets the latest gateway latency.
      *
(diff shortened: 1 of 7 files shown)


Bastian:
I'm pretty sure that you only need to delay the IDENTITY payload and not the whole connection. Adding the extra 500ms to make up for some network latency inconsistencies sounds like a good idea though.

Bastian (approved):
Looks good to me 👍 

Can you please rebase to the latest version and squash the PR into a single commit?

coaxial1:
Rebased and squashed :)

PR #630, opened 2020-09-29
Message#deleteAfter convenience methods

cap5lut:
This PR adds convenience methods to the `Message` interface, to delete message after a given delay.
By default it uses the `ScheduledExecutorService` returned from `ThreadPool#getScheduler()`

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -30,6 +30,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


Saladoc (approved):

Bastian:
Thanks for the PR!

I'm not sure if we want to go this route. If we add `Message#deleteAfter(...)`, someone might also want `Message#removeReactionAfter(...)`, `ServerChannel#deleteAfter(...)`, `MessageBuilder#sendAfter(...)` and so on.

I think this is out of scope for Javacord.

Saladoc:
I disagree with you on that Bastian. The other usages you mentioned are never (or seldom) asked for while deleting a Message after a given time is an evergreen in the help channel. Besides, you opened that can of worms with `ListenerManager#removeAfter` so your slippery slope argument is at worst invalid and at best hypocritical.

We could of course generify this by adding to the `ThreadPool` the following method:
```java
public <T> CompletableFuture<T> runAfter(Supplier<CompletableFuture<T>> task, long delay, TimeUnit unit) {
    CompletableFuture<T> future = new CompletableFuture<>();
    getScheduler().schedule( () -> task.get().whenComplete( (result, ex) -> {
                        if (ex != null) {
                            future.completeExceptionally(ex);
                        } else {
                            future.complete(result);
                        }
                )
            }, delay, unit);
    return future;
}
```

That way, anything could be delayed easily without cluttering the api with methods for everything and have people use 
```java
someChannel.sendMessage("Foo").thenAccept(message -> message.getApi().getThreadPool().runAfter(message::delete, 5, TimeUnit.MINUTES);
```

coaxial1:
To be honest, I agree with Saladoc. Apart from the fact that a lot of bots are deleting their message after a few seconds for several purposes, I'm using delayed message deletion a few times myself, and I ended up creating a utility method for it (basically a static ``MessageUtil.deleteMessageAfter(message, 3, TimeUnit.SECONDS)``) because writing something like Saladoc wrote in his last paragraph each and every time is not beautiful. A direct Javacord method on each message object would make this easier.

cap5lut:
from discord:
> **cap5lut**: so basically i rework PR  630 for the ThreadPool#runAfter method?
> **Bastian**: Given the feedback from most users, I think you should keep the deleteAfter(...) methods. We can additionally add the runAfter() methods.
> **cap5lut**: if yes, does it make sense to have 2 versions of the method, one that gets the ScheduledExecutorService passed and one that uses the thread pool's own scheduler?
>**Bastian**: I don't think so. If someone wants to user their own scheduler, they can just write their own runAfter(...) method.

I'm thinking about Bastian's argument about users asking for other entities having this method as well,
basically I could make it an interface like
```java
interface Deletable extends DiscordEntity {
    CompletableFuture<Void> delete();
    default CompletableFuture<Void> deleteAfter(...) {
        // implementation
    }
}```
i think that would be an easily maintainable way without doing too much on the entity interfaces itself,
but unlike now that would be a breaking change i think.

whats your opinion about it?

coaxial1:
To me, that looks like a good option. And I think there are some other PRs marked as breaking-change that could all be merged in 3.2.0, together with Bastians event & entity rewrite, API / Gateway v8 etc. There will be a lot of changes anyways so I wouldn't care too much about it being a breaking change tbh

Bastian:
> whats your opinion about it?
Sounds good.

Joshix-1:
What's the state of this?

cap5lut:
I did not find the time to further work on this yet.

KILLEliteMaste:
Closed due to a newer version in #1038

PR #633, opened 2020-09-30
Fix ServerVoiceChannelMemberEvent

Bastian:


javacord-core/src/main/java/org/javacord/core/event/channel/server/voice/ServerVoiceChannelMemberEventImpl.java
```diff
@@ -2,26 +2,27 @@
...
+     * The member of the event.
      */
-    private final Long userId;
+    private final Member member;
 
     /**
      * Creates a new voice channel member event.
      *
-     * @param userId The id of the user of the event.
+     * @param member The member of the event..
```
KILLEliteMaste, line 20: Additional dot

javacord-core/src/main/java/org/javacord/core/event/channel/server/voice/ServerVoiceChannelMemberLeaveEventImpl.java
```diff
@@ -20,13 +21,13 @@
     /**
      * Creates a new server voice channel member leave event.
      *
-     * @param userId The id of the user of the event.
+     * @param member Themember of the event.
```
KILLEliteMaste, line 24: Missing space

KILLEliteMaste (changes requested):

PR #637, opened 2020-10-01
Add user public flags

Mysterypotatoguy:
Adds the public user flags (aka badges) property to the User object

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
@@ -1,0 +1,50 @@
+package org.javacord.api.entity.user;
+
+public enum UserFlag {
+
+    DISCORD_EMPLOYEE(1 << 0, "Discord Employee"),
+    DISCORD_PARTNER(1 << 1, "Discord Partner"),
+    HYPESQUAD_EVENTS(1 << 2, "HypeSquad Events"),
+    BUG_HUNTER_LEVEL_1(1 << 3, "Bug Hunter Level 1"),
+    HOUSE_BRAVERY(1 << 6, "House Bravery"),
+    HOUSE_BRILLIANCE(1 << 7, "House Brilliance"),
+    HOUSE_BALANCE(1 << 8, "House Bravery"),
+    EARLY_SUPPORTER(1 << 9, "Early Supporter"),
+    TEAM_USER(1 << 10, "Team User"),
+    SYSTEM(1 << 12, "System"),
+    BUG_HUNTER_LEVEL_2(1 << 14, "Bug Hunter Level 2"),
+    VERIFIED_BOT(1 << 16, "Verified Bot"),
+    VERIFIED_BOT_DEVELOPER(1 << 17, "Verified Bot Developer");
+
+    private final int flag;
+    private final String description;
+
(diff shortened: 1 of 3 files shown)


javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
```diff
@@ -0,0 +1,50 @@
...
+    HOUSE_BALANCE(256, "House Bravery"),
+    EARLY_SUPPORTER(512, "Early Supporter"),
+    TEAM_USER(1024, "Team User"),
+    SYSTEM(4096, "System"),
+    BUG_HUNTER_LEVEL_2(16384, "Bug Hunter Level 2"),
+    VERIFIED_BOT(65536, "Verified Bot"),
+    VERIFIED_BOT_DEVELOPER(131072, "Verified Bot Developer");
+
+    int flag;
+    String description;
```
Bastian, line 20:
```suggestion
    private final int flag;
    private final String description;
```

javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
```diff
@@ -0,0 +1,50 @@
+package org.javacord.api.entity.user;
+
+public enum UserFlag {
+
+    DISCORD_EMPLOYEE(1, "Discord Employee"),
+    DISCORD_PARTNER(2, "Discord Partner"),
+    HYPESQUAD_EVENTS(4, "HypeSquad Events"),
+    BUG_HUNTER_LEVEL_1(8, "Bug Hunter Level 1"),
+    HOUSE_BRAVERY(64, "House Bravery"),
```
Bastian, line 9: I think it would be better to write it like in the Discord docs instead of using decimal notation:
```suggestion
    DISCORD_EMPLOYEE(1 << 0, "Discord Employee"),
    DISCORD_PARTNER(1 << 1, "Discord Partner"),
    HYPESQUAD_EVENTS(1 << 2, "HypeSquad Events"),
    BUG_HUNTER_LEVEL_1(1 << 3, "Bug Hunter Level 1"),
    HOUSE_BRAVERY(1 << 5, "House Bravery"),
```
and so on

Bastian (changes requested):

Bastian (approved):

javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
```diff
@@ -0,0 +1,50 @@
...
+
+public enum UserFlag {
+
+    DISCORD_EMPLOYEE(1 << 0, "Discord Employee"),
+    DISCORD_PARTNER(1 << 1, "Discord Partner"),
+    HYPESQUAD_EVENTS(1 << 2, "HypeSquad Events"),
+    BUG_HUNTER_LEVEL_1(1 << 3, "Bug Hunter Level 1"),
+    HOUSE_BRAVERY(1 << 6, "House Bravery"),
+    HOUSE_BRILLIANCE(1 << 7, "House Brilliance"),
+    HOUSE_BALANCE(1 << 8, "House Bravery"),
```
Zeitheron, line 11: Duplicate `House Bravery` ? I think it's supposed to be `House Balance`

Issue #574, opened 2020-06-07
Add missing canXyz() methods


Bastian:
There are some missing `canXyz` helper methods, like
* `ServerVoiceChannel#canSpeak(User)`
* `ServerVoiceChannel#canUseVideo(User)`
* ...

(The list is incomplete)

PR #638, opened 2020-10-01
Moved and added missing canXY() methods from VoiceChannel to ServerVoiceChannel

KILLEliteMaste:
Already existing canXY() methods now check only the required permission.
Removed canConnectToVoiceChannel and canMuteUsersInVoiceChannel from MessageAuthor because there was no use.

Solves #574

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
@@ -1,6 +1,7 @@
 package org.javacord.api.entity.channel;
 
 import org.javacord.api.audio.AudioConnection;
+import org.javacord.api.entity.permission.PermissionType;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.listener.channel.server.voice.ServerVoiceChannelAttachableListenerManager;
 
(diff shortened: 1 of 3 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/MessageAuthor.java
```diff
@@ -514,37 +514,6 @@ default boolean canMentionEveryoneInTextChannel() {
...
 
-    /**
-     * Checks if the author can connect to the voice channel where the message was sent.
-     * In private chats (private channel or group channel) this always returns {@code true} if the user is
-     * part of the chat.
-     * Always returns {@code false} if the author is not a user or if the channel is not a voice channel.
-     *
-     * @return Whether the author can connect to the voice channel or not.
-     */
-    default boolean canConnectToVoiceChannel() {
```
Bastian, line 525: Please deprecate the methods instead of just removing them. That way, we don't introduce a breaking change.

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
```diff
@@ -73,6 +74,165 @@ default boolean isConnected(User user) {
...
 
+    /**
+     * Checks if the given user is a priority speaker in this voice channel.
+     *
+     * @param user The user to check.
+     * @return Whether the given user is a priority speaker or not.
+     */
+    default boolean isPrioritySpeaker(User user) {
+        return asServerVoiceChannel().map(serverVoiceChannel -> serverVoiceChannel
+                .hasAnyPermission(user, PermissionType.ADMINISTRATOR, PermissionType.PRIORITY_SPEAKER))
```
Bastian, line 85: This does not take into account if the user is the server's owner. (also at other places)

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
```diff
@@ -73,6 +74,165 @@ default boolean isConnected(User user) {
...
     }
 
+    /**
+     * Checks if the given user is a priority speaker in this voice channel.
+     *
+     * @param user The user to check.
+     * @return Whether the given user is a priority speaker or not.
+     */
+    default boolean isPrioritySpeaker(User user) {
+        return asServerVoiceChannel().map(serverVoiceChannel -> serverVoiceChannel
```
Bastian, line 84: Unnecessary `asServerVoiceChannel()` (also at other places)

Bastian (changes requested):
Thank you! 

Just some small parts that need improvements.

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
KILLEliteMaste: It does because the server owner automatically has all permissions and so the permission check will return true

Saladoc (approved):

PR #640, opened 2020-10-01
Add a method to replace mentions with readable content

Joshix-1:
If you just stored the content and want to replace the mentions with updated names you would have to write an own method, so why not just add it to DiscordApi so everyone can access it?


Bastian:
Maybe better make it a static method in `Message` or something similar. The `DiscordApi` is already quite polluted with hundreds of methods.

Joshix-1:
If it was a static method in message you would need the DiscordApi as parameter. I don't think that it would be better than.

Bastian (approved):

Bastian:
Thank you!

Issue #552, opened 2020-05-04
Inconsistent / Unfitting download methods


Bastian:
Our download methods are currently no consistent.
In some places like `Icon` we have `#asByteArray()`, `#asInputStream()`, and `#asBufferedImage()` but in other places like `MessageAttachment` we have `#downloadAsByteArray()`, `#downloadAsInputStream()`, and `#downloadAsBufferedImage()`.

PR #641, opened 2020-10-01
Change name of download methods for consistency

KILLEliteMaste:
Solved #552

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/Attachment.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/Attachment.java
@@ -73,19 +73,54 @@ public interface Attachment extends DiscordEntity {
     Optional<Boolean> isEphemeral();
 
     /**
-     * Gets the attachment as an input stream.
+     * Downloads the attachment as an input stream.
+     *
+     * @return The attachment as an input stream.
+     * @throws IOException If an IO error occurs.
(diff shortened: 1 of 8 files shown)


Bastian:
The old methods should get deprecated instead of just being removed.

Saladoc:
It might be prudent to instead of duplicating the implementation code have the old methods be `default` methods on the interfaces and have them call the new method. That way we contain our backwards-compatibility change in the API and it will be easier to remove on the next breaking update.

Saladoc (approved):
Seeing as we include a huge set of breaking changes in the next release we might just as well forego the deprecation

Bastian:
> Seeing as we include a huge set of breaking changes in the next release we might just as well forego the deprecation

Deprecations are not only to not have breaking changes, but also to make it easier for the user to migrate. When they update, they get a warning and can look at the JavaDocs what the replacement method is. When we just remove the method without deprecation, they just have an error and have to guess the replacement method or read all the changelogs since their last update (not everyone updates regularly).

Saladoc (changes requested):
Following the discussion on Discord we should stick with the earlier plan and keep the older methods, but deprecate them, scheduled to remove in a future release.

KILLEliteMaste:
Added the old methods back and marked them as deprecated

Saladoc (approved):

PR #643, opened 2020-10-03
Utilize Jackson databinding

Bastian:
WIP


Bastian:
I will not finish this PR.
Switching the complete code-base to databind is much more work than expected and not worth the effort.

PR #650, opened 2020-10-11
Add Message#crossPost() and make Message#edit() return the Message

Joshix-1:


javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -153,8 +153,10 @@ public MessageImpl(DiscordApiImpl api, TextChannel channel, JsonNode data) {
         Long webhookId = data.has("webhook_id") ? data.get("webhook_id").asLong() : null;
         author = new MessageAuthorImpl(this, webhookId, data);
 
-        MessageCacheImpl cache = (MessageCacheImpl) channel.getMessageCache();
-        cache.addMessage(this);
+        if (channel != null) {
```
Bastian, line 156: In which case can the channel be `null`?
If that's the case, we would also have to make functions like `Message#getChannel()` return an Optional.

Joshix-1: I don't now why I did this.
I don't think it can be null.

Bastian: I've removed it

PR #655, opened 2020-10-24
Expose UnicodeEmoji Implementation.

Saladoc:
# What
 - provide equals and hashCode methods for all `Emoji` implementations
 - rename `UnicodeEmojiImpl` to `UnicodeEmoji`
 - expose `UnicodeEmoji` in the API project

# Why
Consider the following use case:
A developer wishes to have a bot react to reactions added to a certain message. The actions will be different per emoji and the developer opts to store them in a `Map<Emoji, Action>`. The developer wishes for both unicode and custom emojis to be possible keys for this map.

## The Issue
With the current implementation, the developer is not able to obtain a unicode emoji object without depending on the core artifact.

### Possible Solution 1: Two Maps
The developer could use two maps, `Map<String, Action>` for unicode emojis and `Map<CustomEmoji, Action>` for custom emojis. This defeats the purpose of having `Emoji` as a unifying interface and has a negative impact on readability. Also, it adds further complexity to concurrency concerns.

### Possible Solution 2: Round Trip
The developer could use `Message#addReactions()` with the unicode emojis as strings to obtain the same emojis as `Emoji` instances. This is needlessly complex and would require a context in which the bot could send (and, for cleanup, remove) reactions without bothering server users.

### Possible Solution 3: Custom Implementation
The developer could write their own `UnicodeEmoji` class implementing the `Emoji` interface. This is possible, but to work properly still requires the `equals` and `hashCode` methods to be implemented in an implementation-agnostic manner in the `CustomEmojiImpl` class. It would also require the developer to follow a hidden contract with regards to their own implementations.

### Possible Solution 4: Exposing the Implementation
This is the solution implemented by this pull request. The `UnicodeEmojiImpl` is self-contained and therefore can be safely moved to the api module. The only issue there might be is users trying to use `UnicodeEmoji#fromString` with the name of an emoji rather than its code points. This could be addressed by a simple check for typically present chars like `_` or `:` and a corresponding `IllegalArgumentException` being thrown.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/emoji/UnicodeEmoji.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/emoji/UnicodeEmoji.java
@@ -1,8 +1,4 @@
-package org.javacord.core.entity.emoji;
-
-import org.javacord.api.entity.emoji.CustomEmoji;
-import org.javacord.api.entity.emoji.Emoji;
-import org.javacord.api.entity.emoji.KnownCustomEmoji;
+package org.javacord.api.entity.emoji;
 
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -10,13 +6,13 @@ import java.util.concurrent.ConcurrentHashMap;
 /**
  * The representation of a unicode {@link Emoji}.
  */
-public class UnicodeEmojiImpl implements Emoji {
+public class UnicodeEmoji implements Emoji {
 
     /**
      * A static map which contains all unicode emojis.
      */
-    // Can be static, because unicode emojis are unique, even for different discord accounts/shards.
-    private static final ConcurrentHashMap<String, UnicodeEmojiImpl> unicodeEmojis = new ConcurrentHashMap<>();
+    // Can be static, because unicode emojis are really unique, even for different discord accounts/shards.
+    private static final ConcurrentHashMap<String, UnicodeEmoji> unicodeEmojis = new ConcurrentHashMap<>();
 
     /**
      * The emoji string.
@@ -28,7 +24,7 @@ public class UnicodeEmojiImpl implements Emoji {
      *
      * @param emoji The emoji string.
      */
-    private UnicodeEmojiImpl(String emoji) {
+    private UnicodeEmoji(String emoji) {
         this.emoji = emoji;
     }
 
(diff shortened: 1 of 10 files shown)


javacord-api/src/main/java/org/javacord/api/entity/emoji/UnicodeEmoji.java
```diff
@@ -71,4 +67,15 @@ public boolean isAnimated() {
     public String toString() {
         return String.format("UnicodeEmoji (emoji: %s)", emoji);
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        return  (this == obj)
+                || ((obj instanceof Emoji) && equalsEmoji((Emoji) obj));
```
Vampire, line 74: This could easily violate the `equals` contract, for example regarding symmetry unless you make the class `final`.
That's why usually `equals` methods check for the class being identical, not `instanceof`.
If you for example create a subclass and in its `equals` method check more properties, the subclass instance could return `false` while this instance returns `true`.

Why did you actually overwrite equals and hashCode?
The `fromString` and private constructor make sure that `emoji` property and `UnicodeEmoji` instance is a 1:1 relationship which makes the custom `equals` and `hashCode` superfluous as the default implementations are perfectly fine.

javacord-core/src/main/java/org/javacord/core/entity/emoji/CustomEmojiImpl.java
```diff
@@ -111,9 +110,7 @@ public boolean isAnimated() {
     @Override
     public boolean equals(Object o) {
         return (this == o)
-               || !((o == null)
-                    || (getClass() != o.getClass())
-                    || (getId() != ((DiscordEntity) o).getId()));
+               || ((o instanceof CustomEmoji) && equalsEmoji((CustomEmoji) o));
```
Vampire, line 113: Same as above, why did you change this code?

Vampire (changes requested):

Saladoc:
After approximately a million years I have revisited this PR - in hindsight, I have no clue why I ever thought this would be a good Idea, but here we are.

Vampire:
> When considering this I felt that an interface / impl separation would be in line with the other emojis, but we only ever do that for objects that are not meant to be created by the user.
Other classes that are specifically meant to be created and used by the user (like the gazillions of XYBuilder classes) do not follow this pattern either.

Actually the user is not creating instances of `UnicodeEmoji`, even with your change.
The constructor is private.
He is just requesting an instances by given criterion.
Just like he requests a server by giving its ID, he requests the unicode emoji by giving its string representation.

This change as it is also introduces a bit of an inconsistency, because there is for example `Emoji#asUnicodeEmoji()` that is returning an `Optional<String>`, not an `Optional<UnicodeEmoji>`.
Would it maybe be better to just have some means to request the emoji for the string like `Emoji DiscordApi#getEmojiByUnicode(String)`, or `Emoji Emoji.byUnicode(String)` or similar? Both would easily be implementable.

Besides that all solutions allow the user to get `Emoji` instances of arbitrary strings. :-(
Currently they have to go through or come from Discord and are checked by their servers whether it is a supported unicode emoji or not.

Saladoc:
This will have to be rethought and redone when I got the time.

Issue #660, opened 2020-11-07
ServerUpdater#removeRolesFromUser removing all roles


ThatBlockyPenguin:
I have a method that searches through all the roles a user currently has, and removes the roles that begin with a certain phrase. If I use `rolesToRemove.forEach(r -> user.removeRole(r))`, it runs a-ok. However, I know that a better way of doing it is to use a ServerUpdater. However, if I run `new ServerUpdater(server).removeRolesFromUser(user, rolesToRemove).update()`, it sure does remove the correct roles... along with every other role the user has, too. Not so great.

I'm using the latest Javacord (v3.1.1)
~Thanks!

Vampire:
@Bastian this is rather critical and hits multiple users.
Pepich also wanted to open an issue about it and another user also just complained.
For that other user it is `canKickUsersFromServer()` that is misbehaving.
For Pepich it is `User#getRoles(Server)`.
The latter is the real problem, as the others use that method and thus behave wrongly.
The problem is easily reproducible and the cause is, that after your changes for intents, the `GuildMemberUpdateHandler` simply does not propagate the changes in roles to the `RoleImp`l` objects anymore.
The result of this is, that the roles do not know their users properly anymore but only the state when the bot was last restarted.

While thinking about it, this here could of course also be due to missing intents.

Bastian:
> The problem is easily reproducible and the cause is, that after your changes for intents, the `GuildMemberUpdateHandler` simply does not propagate the changes in roles to the `RoleImp`l` objects anymore.
> The result of this is, that the roles do not know their users properly anymore but only the state when the bot was last restarted.

That's not fully true.
The `MemberImpl` object has a `roleIds` field that gets updated with the event. This is also what should be used for the `User#canXyz()` method.

The idea was to have the Member object as the SSOT and the `Role#getMembers()` method being a simple filter for `Server#getMembers()` with `User#hasRole(...)` but it seems like I forgot to update it.

Bastian:
Regarding intents: Without the `GUILD_MEMBERS` intent, the `Role#getMembers()` method will obviously stop working.
However, the `User#getRoles(Server)` method will still return the correct roles if the user was retrieved from an event on the same server (e.g. from `MessageCreateEvent#getUser()`).

Bastian:
[...]

PR #662, opened 2020-11-15
Use member object as SSOT for role membership

Bastian:
Fixes https://github.com/Javacord/Javacord/issues/660

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/permission/RoleImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/permission/RoleImpl.java
@@ -9,6 +9,8 @@ import org.javacord.api.entity.server.Server;
 import org.javacord.api.entity.user.User;
 import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.entity.server.ServerImpl;
+import org.javacord.core.entity.user.Member;
+import org.javacord.core.entity.user.UserImpl;
 import org.javacord.core.listener.server.role.InternalRoleAttachableListenerManager;
 import org.javacord.core.util.rest.RestEndpoint;
 import org.javacord.core.util.rest.RestMethod;
@@ -17,12 +19,9 @@ import org.javacord.core.util.rest.RestRequest;
 import java.awt.Color;
 import java.util.Collection;
 import java.util.Comparator;
-import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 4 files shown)


javacord-core/src/main/java/org/javacord/core/entity/permission/RoleImpl.java
```diff
@@ -264,7 +233,14 @@ public boolean isDisplayedSeparately() {
 
     @Override
     public boolean hasUser(User user) {
-        return userIds.contains(user.getId());
+        return ((UserImpl) user).getMember()
+                .filter(member -> member.getServer().getId() == server.getId())
```
Vampire, line 237: If you edit anyway, I'd like to suggest using the more object oriented `member.getServer().equals(server)`.

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1274,12 +1268,16 @@ public boolean isMember(User user) {
...
     @Override
     public List<Role> getRoles(User user) {
-        return getRealMemberById(user.getId())
-                .map(Member::getRoles).orElseGet(() -> {
-                    ArrayList<Role> list = new ArrayList<>();
-                    list.add(getEveryoneRole());
-                    return list;
-                });
+        return ((UserImpl) user).getMember()
+                .filter(member -> member.getServer().getId() == getId())
```
Vampire, line 1272: `member.getServer().equals(this)` like above

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1274,12 +1268,16 @@ public boolean isMember(User user) {
...
+                .map(Member::getRoles)
+                .orElseGet(() ->
+                        getRealMemberById(user.getId())
+                                .map(Member::getRoles).orElseGet(() -> {
+                                    ArrayList<Role> list = new ArrayList<>();
+                                    list.add(getEveryoneRole());
```
Vampire, line 1278: I don't think it is correct that a user has the everyone role if he is not a member of the server. I think it should instead be `.orElseGet(Collections::emptyList)`.

Vampire (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
Bastian: I'm unsure about whether it should contain the role or not, but considering that it has always been like this in Javacord, I won't change it (now). This would be a breaking change.

Vampire: That's not correct or at least it wasn't before the cache change.
`org.javacord.api.entity.server.Server#getRoles(org.javacord.api.entity.user.User)` was
```java
return Collections.unmodifiableList(
        getRoles().stream()
                .filter(role -> role.hasUser(user))
```
and `org.javacord.core.entity.permission.RoleImpl#hasUser` started with
```java
if (isEveryoneRole()) {
    return getServer().isMember(user);
}
```
So you correctly only got the everyone role if the user was a member of that server.

Vampire (changes requested):
lgtm except for the bug with the everyone role that was introduced with the cache/intent-changes.
The old behavior that Javacord always had makes much more sense imho, because if a user is not a member of a server it does not make sense if he has any role, but he should have no role on that server.

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
Bastian: Seems like my memory was wrong then :|

Vampire (approved):
lgtm

Bastian:
Thx for the review :)

PR #666, opened 2020-11-23
Add ServerNewsChannel and allow to follow them

Joshix-1:
See: https://discord.com/developers/docs/resources/channel#follow-news-channel

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -9,6 +9,7 @@ import org.javacord.api.entity.channel.ChannelCategory;
 import org.javacord.api.entity.channel.PrivateChannel;
 import org.javacord.api.entity.channel.RegularServerChannel;
 import org.javacord.api.entity.channel.ServerChannel;
+import org.javacord.api.entity.channel.ServerNewsChannel;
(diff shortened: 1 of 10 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -1518,6 +1526,58 @@ default String makeMentionsReadable(String content) {
...
+     *
+     * @param name The name of the server news channels.
+     * @return A collection with all server news channels with the given name.
+     */
+    default Collection<ServerNewsChannel> getServerNewsChannelsByNameIgnoreCase(String name) {
+        return Collections.unmodifiableList(
+                getServerNewsChannels().stream()
+                        .filter(channel -> channel.getName().equalsIgnoreCase(name))
+                        .collect(Collectors.toList()));
+    }
```
montlikadani, line 1579: I would rather remove this method and instead put a boolean in the previous method as a parameter called `ignoreCase`. This would make less code and more understandable.

Joshix-1: I just did it like in the other methods. And I think that it would be less understandable.

BoneyIsSpooky: As originally written, the IgnoreCase method is consistent with existing methods.

PR #667, opened 2020-11-24
Allow banning users by id

Joshix-1:
This can be useful if you want to ban users that aren't in the cache. (You can ban users that aren't on the server.)
With this you don't have to request the users from discord and can just ban them.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -1818,7 +1818,7 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      * @return A future to check if the ban was successful.
      */
     default CompletableFuture<Void> banUser(User user) {
-        return banUser(user, 0, null);
+        return banUser(user.getId(), 0, null);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -1841,6 +1841,70 @@ default ServerUpdater createUpdater() {
      */
     CompletableFuture<Void> banUser(User user, int deleteMessageDays, String reason);
```
KILLEliteMaste, line 1842: Change to 
```
CompletableFuture<Void> banUser(User user, int deleteMessageDays, String reason) {
    return banUser(Long.toUnsignedString(user.getIdAsString()), deleteMessageDays, reason);
}
```

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1362,8 +1362,13 @@ public void selfUndeafen() {
 
     @Override
     public CompletableFuture<Void> banUser(User user, int deleteMessageDays, String reason) {
+        return banUser(user.getIdAsString(), deleteMessageDays, reason);
+    }
```
KILLEliteMaste, line 1366: Not necessary anymore if you apply the previous change

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):
lgtm

PR #679, opened 2020-12-23
Replies and v8

### LINKED ISSUE #642: Upgrade to gateway version v8
https://discord.com/developers/docs/change-log#api-and-gateway-v8

Joshix-1:
fixes #642 and adds replies

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerChannelBuilderDelegateImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerChannelBuilderDelegateImpl.java
@@ -88,14 +88,14 @@ public class ServerChannelBuilderDelegateImpl implements ServerChannelBuilderDel
         for (Map.Entry<Long, Permissions> entry : overwrittenUserPermissions.entrySet()) {
             permissionOverwrites.addObject()
                     .put("id", Long.toUnsignedString(entry.getKey()))
-                    .put("type", "member")
+                    .put("type", 1)
                     .put("allow", entry.getValue().getAllowedBitmask())
                     .put("deny", entry.getValue().getDeniedBitmask());
         }
         for (Map.Entry<Long, Permissions> entry : overwrittenRolePermissions.entrySet()) {
             permissionOverwrites.addObject()
                     .put("id", Long.toUnsignedString(entry.getKey()))
-                    .put("type", "role")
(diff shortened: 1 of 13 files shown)


javacord-core/src/main/java/org/javacord/core/entity/channel/ServerChannelBuilderDelegateImpl.java
```diff
@@ -88,14 +88,14 @@ protected void prepareBody(ObjectNode body) {
         for (Map.Entry<Long, Permissions> entry : overwrittenUserPermissions.entrySet()) {
             permissionOverwrites.addObject()
                     .put("id", Long.toUnsignedString(entry.getKey()))
-                    .put("type", "member")
+                    .put("type", "1")
```
Bastian, line 91: Does this work with a string?
In the `ServerChannelUpdaterDelegateImpl` you set it as an int?

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -185,19 +194,31 @@ public MessageImpl(DiscordApiImpl api, TextChannel channel, JsonNode data) {
...
         }
 
-        if (data.has("activity") && !data.get("activity").isNull()) {
+        if (data.hasNonNull("activity")) {
             activity = new MessageActivityImpl(this, data.get("activity"));
         } else {
             activity = null;
         }
+
+        if (data.has("message_reference")) {
```
Bastian, line 211: Better always use `hasNonNull`. You cannot trust Discord or its docs.
```suggestion
        if (data.hasNonNull("message_reference")) {
```

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerChannelBuilderDelegateImpl.java
Joshix-1: It works with Strings too, but I changed it to be an int.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
Joshix-1: Changed it. :)

Bastian (approved):

PR #682, opened 2020-12-27
Add Message#getNonce

### LINKED ISSUE #680: Messages are missing nonce fields
MessageBuilders provide setNonce() methods but the returned Message is missing said nonce field.

Joshix-1:
Fixes #680

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
@@ -97,6 +97,11 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
     private final MessageActivityImpl activity;
 
     /**
+     * The nonce of the message.
+     */
+    private final String nonce;
+
+    /**
      * If the message should be cached forever or not.
      */
     private volatile boolean cacheForever = false;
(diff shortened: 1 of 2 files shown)


Bastian (approved):
Lgtm! 👍

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -185,19 +190,25 @@ public MessageImpl(DiscordApiImpl api, TextChannel channel, JsonNode data) {
...
         }
 
-        if (data.has("activity") && !data.get("activity").isNull()) {
+        if (data.hasNonNull("activity")) {
             activity = new MessageActivityImpl(this, data.get("activity"));
         } else {
             activity = null;
         }
+
+        if (data.has("nonce")) {
```
Bastian, line 207: Just one thing: Maybe better use 
```suggestion
        if (data.hasNonNull("nonce")) {
```
otherwise, a `null` nonce would appear at `"null"`. I'm not sure if Discord ever sends null-nounces, but it's Discord: You can never know.

Joshix-1: Changed it.

PR #684, opened 2020-12-31
Added Get Guild Member endpoint

KILLEliteMaste:
This adds the possibility to get a member of a server outside of an event. So it's no longer required to activate the GUILD_MEMBERS intent if you just want to get the users roles for instance.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
@@ -1406,6 +1406,13 @@ public class ServerImpl implements Server, Cleanupable, InternalServerAttachable
     }
 
     @Override
+    public CompletableFuture<User> requestMember(long userId) {
+        return new RestRequest<User>(getApi(), RestMethod.GET, RestEndpoint.SERVER_MEMBER)
+                .setUrlParameters(getIdAsString(), Long.toUnsignedString(userId))
+                .execute(result ->  new MemberImpl(api, this, result.getJsonBody(), null).getUser());
+    }
+
+    @Override
     public CompletableFuture<Void> kickUser(User user, String reason) {
         return new RestRequest<Void>(getApi(), RestMethod.DELETE, RestEndpoint.SERVER_MEMBER)
                 .setUrlParameters(getIdAsString(), user.getIdAsString())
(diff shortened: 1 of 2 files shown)


javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1355,6 +1355,14 @@ public void selfUndeafen() {
                 this, getConnectedVoiceChannel(api.getYourself()).orElse(null), null, false);
     }
 
+    @Override
+    public CompletableFuture<Member> requestMember(long userId) {
+        return new RestRequest<Member>(getApi(), RestMethod.GET, RestEndpoint.SERVER_MEMBER)
+                .setUrlParameters(getIdAsString(), Long.toUnsignedString(userId))
+                .execute(result ->  new MemberImpl(api, this, result.getJsonBody(), null));
+
+    }
```
Bastian, line 1364:
```suggestion
    @Override
    public CompletableFuture<User> requestMember(long userId) {
        return new RestRequest<User>(getApi(), RestMethod.GET, RestEndpoint.SERVER_MEMBER)
                .setUrlParameters(getIdAsString(), Long.toUnsignedString(userId))
                .execute(result -> new MemberImpl(api, this, result.getJsonBody(), null).getUser());

    }
```

This way, we don't have to expose the member object (yet).

AyuAi:
This is going to be quite a useful change for non-full caches after #816 👀

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1407,6 +1407,14 @@ public void selfUndeafen() {
                 this, getConnectedVoiceChannel(api.getYourself()).orElse(null), null, false);
     }
 
+    @Override
+    public CompletableFuture<User> requestMember(long userId) {
+        return new RestRequest<User>(getApi(), RestMethod.GET, RestEndpoint.SERVER_MEMBER)
+                .setUrlParameters(getIdAsString(), Long.toUnsignedString(userId))
+                .execute(result ->  new MemberImpl(api, this, result.getJsonBody(), null).getUser());
+
+    }
```
Bastian, line 1416:
```suggestion
    @Override
    public CompletableFuture<User> requestMember(long userId) {
        return new RestRequest<User>(getApi(), RestMethod.GET, RestEndpoint.SERVER_MEMBER)
                .setUrlParameters(getIdAsString(), Long.toUnsignedString(userId))
                .execute(result -> new MemberImpl(api, this, result.getJsonBody(), null).getUser());
    }
```
Just some nitpicky format changes :)

Bastian (approved):
Lgtm 👍

PR #685, opened 2021-01-02
Move user updates from PresenceUpdateHandler to GuildMemberUpdateHandler

coaxial1:
This PR moves the handling of base user updates (username, discriminator, avatar) from the PresenceUpdateHandler to the GuildMemberUpdateHandler.

https://github.com/discord/discord-api-docs/pull/1307#issuecomment-581561519
This comment from Discord suggests that they fire a ``GUILD_MEMBER_UPDATE``event if a user changes so bots can opt out of receiving presences just for this purpose. I tested it myself, and Discord is indeed firing one ``GUILD_MEMBER_UPDATE``packet per mutual guild once a user changes their name.

Moving the whole logic makes sense as handling presences is a huge computational burden, and users might want to disable it while still receiving user updates; additionally, ``GUILD_MEMBER_UPDATE``events are much less frequent and should be used for updating the user cache even if presence updates are enabled to improve performance. 

Currently, Javacord is not updating the base user based on the user object in ``GUILD_MEMBER_UPDATE``events, and only listens to changes to the base user data in presence updates. Therefore, this PR moves the whole handling of base user updates to the Guild Member Update Handler.


KILLEliteMaste:
lgtm

PR #686, opened 2021-01-03
Fixed non working IDE Setup links

KILLEliteMaste:


coaxial1:
I think you need to update ``.github/README_template.md`` instead. Did the same mistake a while ago

KILLEliteMaste:
> I think you need to update `.github/README_template.md` instead. Did the same mistake a while ago

You are right, thank you 🙂

Issue #562, opened 2020-05-09
Change return type of collections


Bastian:
Some methods like `DiscordApi#getChannels()` return a `Collection<>`. These should be changed to return a `Set<>` (or `List<>` if appropriate, but that's already the case most of the time).

Joshix-1:
java.util.Set?

Bastian:
Yes

PR #692, opened 2021-01-06
allow multiple activities

### LINKED ISSUE #691: Streaming not recognized properly with custom status
Purple streaming on profile is not recognized when a status update happens.

In the words of Vampire: 
"The PRESENCE_UPDATE has a game field and an activities field.
The game is what is also shown as status in the client and is what we return on getActivity and what the event is fired about.
The activities is an array where your custom status and your twitch streaming is contained, but Javacord does not currently evaluate that field, I guess it is something new." -- Vampire

He also explained that javacord only evaluates game and ignores the activities

Javacord version: 3.1.2

Joshix-1:
fixes #691

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
@@ -80,7 +80,7 @@ public interface User extends DiscordEntity, Messageable, Nameable, Mentionable,
      *
      * @return The activities of the user.
      */
-    Collection<Activity> getActivities();
+    Set<Activity> getActivities();
 
     /**
      * Checks if the user can manage the target role.
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
@@ -76,11 +76,11 @@ public interface User extends DiscordEntity, Messageable, Nameable, Mentionable,
     }
 
     /**
-     * Gets the activity of the user.
+     * Gets the activities of the user.
      *
-     * @return The activity of the user.
+     * @return The activities of the user.
      */
-    Optional<Activity> getActivity();
+    Collection<Activity> getActivities();
 
     /**
      * Checks if the user can manage the target role.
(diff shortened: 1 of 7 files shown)


javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -76,11 +76,11 @@ default boolean isBotOwner() {
...
 
     /**
-     * Gets the activity of the user.
+     * Gets the activities of the user.
      *
-     * @return The activity of the user.
+     * @return The activities of the user.
      */
-    Optional<Activity> getActivity();
+    Collection<Activity> getActivities();
```
Bastian, line 83: This should be a `Set<Activity>` (if the order does not matter) or a `List<Activity>` (if the order does matter) instead. 

See https://github.com/Javacord/Javacord/issues/562

javacord-api/src/main/java/org/javacord/api/event/user/UserChangeActivityEvent.java
```diff
@@ -2,25 +2,25 @@
...
 
     /**
-     * Gets the old activity of the user.
+     * Gets the old activities of the user.
      *
-     * @return The old activity of the user.
+     * @return The old activities of the user.
      */
-    Optional<Activity> getOldActivity();
+    Collection<Activity> getOldActivities();
```
Bastian, line 17:
```suggestion
    Set<Activity> getOldActivities();
```

javacord-api/src/main/java/org/javacord/api/event/user/UserChangeActivityEvent.java
```diff
@@ -2,25 +2,25 @@
...
 
     /**
-     * Gets the new activity of the user.
+     * Gets the new activities of the user.
      *
-     * @return The new activity of the user.
+     * @return The new activities of the user.
      */
-    Optional<Activity> getNewActivity();
+    Collection<Activity> getNewActivities();
```
Bastian, line 24:
```suggestion
    Set<Activity> getNewActivities();
```

Joshix-1: changed it :)

PR #698, opened 2021-01-11
Fix event dispatching (#697)

### LINKED ISSUE #697: NPE when listener was running too long
```
[ERROR] 2021-01-11 00:04:03 [Javacord - Central Scheduler - 1] EventDispatcherBase - Interrupted a a listener thread for SERVERNAME, because it was running over 120 seconds! This was most likely caused by a deadlock or very heavy computation/blocking operations in the listener thread. Make sure to not block listener threads!
[ERROR] 2021-01-11 00:04:03 [Javacord - Central ExecutorService - 991] EventDispatcherBase - Unhandled exception in a listener thread for SERVERNAME
java.lang.NullPointerException: null
	at org.javacord.core.util.event.EventDispatcherBase.lambda$checkRunningListenersAndStartIfPossible$17(EventDispatcherBase.java:265) ~[javacord-core-3.1.2.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) [?:?]
	at java.lang.Thread.run(Thread.java:832) [?:?]
```
Note: The application was absolutely broken and I couldn't find the cause for it so far meaning I can't reproduce this

Vampire:
Due to race conditions sometimes NullPointerExceptions happened
and multiple listeners for the same queue selector could run
concurrently.

fixes #697

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
@@ -6,15 +6,14 @@ import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.entity.server.ServerImpl;
 import org.javacord.core.util.logging.LoggerUtil;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


KILLEliteMaste (approved):
LGTM!

PR #704, opened 2021-01-31
Add Cloudflare Rate Limit Handling

LogalDeveloper:
Discord's API has an undocumented global rate limit imposed by [Cloudflare](https://support.cloudflare.com/hc/en-us/articles/115001635128-Configuring-Cloudflare-Rate-Limiting) when a single IPv4 address or IPv6 /64 IP range exceeds an unknown threshold. There is an important difference between Cloudflare's rate limit and Discord's rate limits: Cloudflare does not send the retry after value in the JSON response body. Instead, it is sent in the `Retry-After` HTTP response header. Javacord currently only checks the JSON response body for the retry after value. This pull request adds detection of a Cloudflare imposed rate limit so it can properly fetch the retry after value from the HTTP response header.


KILLEliteMaste:
Closing this PR as there is a more recent version of trying to fix the issue in #756

Issue #524, opened 2020-04-09
No ReactionEvents for uncached private channels


VolcanoCookies:
Version: 3.0.5
Version Commit ID: 917e4b351e5c8a1f53cb02be68d05f06b5477986

It appears that globally attached ReactionAddListener and ReactionRemoveListener do not receive reaction events from *uncached?* private channels.

The following code does not receive ReactionEvents from private channels, unless a private message is first sent to the bot after it has started, or the private channel is opened in any other way. *Message loads the channel into cache?*

```
api.addReactionAddListener(event -> {
    System.out.println("Event received");
});
````

coaxial1:
This is probably related to https://github.com/discord/discord-api-docs/issues/1316

Bastian:
@VolcanoCookies thank you for the bug report and @coaxial1 thanks for the hint.

I see two options:
1. We fix this bug now by making the channel in all message related events optional (including message edit and message delete) and only provide the channel id and utility methods to interact with the channel without having to fetch the channel object.
2. Wait for the v7 gateway version:
[![image](https://user-images.githubusercontent.com/5033001/79043119-80703100-7bfd-11ea-9265-aab8fea63ea2.png)](https://github.com/discord/discord-api-docs/issues/184#issuecomment-312960055)
Unfortunately jhgg wasn't clear if in v7 the full channel information will also be available for reactions, edits and deletions or only for message create events.

PR #708, opened 2021-02-03
compensate missing CHANNEL_CREATE event for text channels

### LINKED ISSUE #707: [Text chat] DM issue with 3.1.3-SNAPSHOT
When an user tries to DM the bot, if the bot, on this running session, did not DM this user, the `MessageCreateEvent` won't be called

image

Javacord version : `3.1.3-SNAPSHOT`

Joshix-1:
fixes #707 
I don't now if it is a good fix, but it works if you don't send messages from another shard than shard 0.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageCreateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageCreateHandler.java
@@ -1,15 +1,21 @@
 package org.javacord.core.util.handler.message;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.logging.log4j.Logger;
 import org.javacord.api.DiscordApi;
 import org.javacord.api.entity.channel.ServerChannel;
+import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.message.Message;
 import org.javacord.api.entity.message.MessageAuthor;
 import org.javacord.api.entity.server.Server;
 import org.javacord.api.event.message.MessageCreateEvent;
+import org.javacord.core.entity.channel.PrivateChannelImpl;
+import org.javacord.core.entity.user.MemberImpl;
+import org.javacord.core.entity.user.UserImpl;
 import org.javacord.core.event.message.MessageCreateEventImpl;
 import org.javacord.core.util.event.DispatchQueueSelector;
 import org.javacord.core.util.gateway.PacketHandler;
+import org.javacord.core.util.logging.LoggerUtil;
 
 import java.util.Optional;
 
@@ -17,6 +23,11 @@ import java.util.Optional;
  * Handles the message create packet.
  */
 public class MessageCreateHandler extends PacketHandler {
+
+    /**
+     * The logger of this class.
+     */
+    private static final Logger logger = LoggerUtil.getLogger(MessageCreateHandler.class);
 
     /**
      * Creates a new instance of this class.
@@ -29,20 +40,41 @@ public class MessageCreateHandler extends PacketHandler {
 
     @Override
     public void handle(JsonNode packet) {
-        api.getTextChannelById(packet.get("channel_id").asText()).ifPresent(channel -> {
-            Message message = api.getOrCreateMessage(channel, packet);
-            MessageCreateEvent event = new MessageCreateEventImpl(message);
+        long channelId = packet.get("channel_id").asLong();
 
-            Optional<Server> optionalServer = channel.asServerChannel().map(ServerChannel::getServer);
-            MessageAuthor author = message.getAuthor();
-            api.getEventDispatcher().dispatchMessageCreateEvent(
-                    optionalServer.map(DispatchQueueSelector.class::cast).orElse(api),
-                    optionalServer.orElse(null),
-                    channel,
-                    author.asUser().orElse(null),
-                    author.isWebhook() ? author.getId() : null,
-                    event);
-        });
+        // if the message isn't from a server
+        // See https://github.com/discord/discord-api-docs/issues/2248
+        if (!packet.hasNonNull("guild_id")) {
+            UserImpl author = new UserImpl(api, packet.get("author"), (MemberImpl) null, null);
+
+            PrivateChannelImpl privateChannel = PrivateChannelImpl
+                    .getOrCreatePrivateChannel(api, channelId, author.getId(), author);
+
+            handle(privateChannel, packet);
+            return;
+        }
+
+        Optional<TextChannel> optionalChannel = api.getTextChannelById(channelId);
+        if (optionalChannel.isPresent()) {
+            handle(optionalChannel.get(), packet);
+        } else {
+            LoggerUtil.logMissingChannel(logger, channelId);
+        }
+    }
+
+    private void handle(TextChannel channel, JsonNode packet) {
+        Message message = api.getOrCreateMessage(channel, packet);
+        MessageCreateEvent event = new MessageCreateEventImpl(message);
+
+        Optional<Server> optionalServer = channel.asServerChannel().map(ServerChannel::getServer);
+        MessageAuthor author = message.getAuthor();
+        api.getEventDispatcher().dispatchMessageCreateEvent(
+                optionalServer.map(DispatchQueueSelector.class::cast).orElse(api),
+                optionalServer.orElse(null),
+                channel,
+                author.asUser().orElse(null),
+                author.isWebhook() ? author.getId() : null,
+                event);
     }
 
 }
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionAddHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionAddHandler.java
@@ -1,12 +1,14 @@
 package org.javacord.core.util.handler.message.reaction;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.logging.log4j.Logger;
 import org.javacord.api.DiscordApi;
-import org.javacord.api.entity.channel.ServerChannel;
+import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.emoji.Emoji;
 import org.javacord.api.entity.message.Message;
 import org.javacord.api.entity.server.Server;
 import org.javacord.api.event.message.reaction.ReactionAddEvent;
+import org.javacord.core.entity.channel.PrivateChannelImpl;
 import org.javacord.core.entity.emoji.UnicodeEmojiImpl;
 import org.javacord.core.entity.message.MessageImpl;
 import org.javacord.core.entity.server.ServerImpl;
@@ -15,6 +17,7 @@ import org.javacord.core.entity.user.MemberImpl;
 import org.javacord.core.event.message.reaction.ReactionAddEventImpl;
 import org.javacord.core.util.event.DispatchQueueSelector;
 import org.javacord.core.util.gateway.PacketHandler;
+import org.javacord.core.util.logging.LoggerUtil;
 
 import java.util.Optional;
 
@@ -22,6 +25,11 @@ import java.util.Optional;
  * Handles the message reaction add packet.
  */
 public class MessageReactionAddHandler extends PacketHandler {
+
+    /**
+     * The logger of this class.
+     */
+    private static final Logger logger = LoggerUtil.getLogger(MessageReactionAddHandler.class);
 
     /**
      * Creates a new instance of this class.
@@ -34,37 +42,50 @@ public class MessageReactionAddHandler extends PacketHandler {
 
     @Override
     public void handle(JsonNode packet) {
-        api.getTextChannelById(packet.get("channel_id").asText()).ifPresent(channel -> {
-            long messageId = packet.get("message_id").asLong();
-            long userId = packet.get("user_id").asLong();
-            Server server = channel.asServerChannel().map(ServerChannel::getServer).orElse(null);
-            Member member = null;
-            if (packet.hasNonNull("member")) {
-                member = new MemberImpl(api, (ServerImpl) server, packet.get("member"), null);
-            }
-            Optional<Message> message = api.getCachedMessageById(messageId);
+        long channelId = packet.get("channel_id").asLong();
+        long messageId = packet.get("message_id").asLong();
+        long userId = packet.get("user_id").asLong();
+        String serverId = packet.hasNonNull("guild_id") ? packet.get("guild_id").asText() : null;
 
-            Emoji emoji;
-            JsonNode emojiJson = packet.get("emoji");
-            if (!emojiJson.has("id") || emojiJson.get("id").isNull()) {
-                emoji = UnicodeEmojiImpl.fromString(emojiJson.get("name").asText());
-            } else {
-                emoji = api.getKnownCustomEmojiOrCreateCustomEmoji(emojiJson);
-            }
+        TextChannel channel;
+        if (serverId == null) { // if private channel:
+            channel = PrivateChannelImpl.getOrCreatePrivateChannel(api, channelId, userId, null);
+        } else {
+            channel = api.getTextChannelById(channelId).orElse(null);
+        }
 
-            message.ifPresent(msg -> ((MessageImpl) msg).addReaction(emoji, userId == api.getYourself().getId()));
+        if (channel == null) {
+            LoggerUtil.logMissingChannel(logger, channelId);
+            return;
+        }
 
-            ReactionAddEvent event = new ReactionAddEventImpl(api, messageId, channel, emoji, userId, member);
+        Optional<Server> server = api.getServerById(serverId);
 
-            Optional<Server> optionalServer = channel.asServerChannel().map(ServerChannel::getServer);
-            api.getEventDispatcher().dispatchReactionAddEvent(
-                    optionalServer.map(DispatchQueueSelector.class::cast).orElse(api),
-                    messageId,
-                    optionalServer.orElse(null),
-                    channel,
-                    userId,
-                    event);
-        });
+        Member member = null;
+        if (packet.hasNonNull("member") && server.isPresent()) {
+            member = new MemberImpl(api, (ServerImpl) server.get(), packet.get("member"), null);
+        }
+        Optional<Message> message = api.getCachedMessageById(messageId);
+
+        Emoji emoji;
+        JsonNode emojiJson = packet.get("emoji");
+        if (!emojiJson.has("id") || emojiJson.get("id").isNull()) {
+            emoji = UnicodeEmojiImpl.fromString(emojiJson.get("name").asText());
+        } else {
+            emoji = api.getKnownCustomEmojiOrCreateCustomEmoji(emojiJson);
+        }
+
+        message.ifPresent(msg -> ((MessageImpl) msg).addReaction(emoji, userId == api.getYourself().getId()));
+
+        ReactionAddEvent event = new ReactionAddEventImpl(api, messageId, channel, emoji, userId, member);
+
+        api.getEventDispatcher().dispatchReactionAddEvent(
+                server.map(DispatchQueueSelector.class::cast).orElse(api),
+                messageId,
+                server.orElse(null),
+                channel,
+                userId,
+                event);
     }
 
 }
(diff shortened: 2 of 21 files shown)


Joshix-1:
With the second commit the text channel will not be created with the bot as receiver.

Joshix-1:
> as reaction to some channel command
That will only happen on shard 0. You won't receive commands on other shards.

Vampire:
> That will only happen on shard 0. You won't receive commands on other shards.

Why not?
I'm talking about commands in server channels

Vampire:
> With the second commit the text channel will not be created with the bot as receiver.

But now you miss the message create event for the bots message.
Maybe the channel needs to be optional, or in this edge case you need to request the channel by id from Discord.

Joshix-1:
I don't like any option. The channel being optional would be a big change. And requesting the id would create an extra unnecessary request.

Vampire:
It's imho not unnecessary if you otherwise loose the event. And it is just one request and only in case the first incoming message on a private channel is by the bot.

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageCreateHandler.java
```diff
@@ -29,20 +39,48 @@ public MessageCreateHandler(DiscordApi api) {
...
-                    author.asUser().orElse(null),
-                    author.isWebhook() ? author.getId() : null,
-                    event);
-        });
+        String channelId = packet.get("channel_id").asText();
+
+        // if the message isn't from a server and the private channel isn't cached
+        // See https://github.com/discord/discord-api-docs/issues/2248
+        if (!packet.hasNonNull("guild_id") && !api.getTextChannelById(channelId).isPresent()) {
+            UserImpl recipient = new UserImpl(api, packet.get("author"), (MemberImpl) null, null);
```
Vampire, line 47: Definitely a nit, but I'd like to suggest calling this `author`. :-)

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageCreateHandler.java
```diff
@@ -29,20 +39,48 @@ public MessageCreateHandler(DiscordApi api) {
...
+
+            if (recipient.isYourself()) {
+                // request private channel from discord
+                new RestRequest<PrivateChannel>(api, RestMethod.GET, RestEndpoint.CHANNEL)
+                        .setUrlParameters(channelId)
+                        .execute(result ->
+                                new PrivateChannelImpl(api, result.getJsonBody()))
+                        .thenAccept(channel -> handle(channel, packet));
```
Vampire, line 55: I think in this case we should actually wait for the result of the request synchronously, or the order of events is not correct anymore.
As this is an edge case anyway, I think it should be ok to wait here without introducing a performance problem.

Vampire, line 55: Also in this case currently no `PrivateChannelCreateEvent` is dispatched.
I think it should uniformly dispatch the event in both cases.
Otherwise you can for example not easily attach a listener to all newly create channels as this one would be missing.

Vampire (changes requested):

Vampire (approved):
lgtm

thanks :-)

coaxial1:
I don't think requesting a channel synchronously is a good idea because I think the case you consider as "edge case" is the normal case. Almost all DM interactions with bots are initiated by bots themselves. Let it be welcome messages, DM replies to commands, DMs sent as notification, DMs as a response to a reaction, etc. There are barely use cases for the user to open a DM channel to a bot on their own behalf, except for special use cases like single-server Ticketing systems or similar things; this is mostly due to the fact that without any open DM session to a user, as a bot, you wouldn't know to which server a request or DM belongs. Therfore, DM interactions / sessions are often created based on server interactions. At least I have only seen one bot yet that allowed DM interactions without a preceding server interaction, and I've seen a lot.

If this is blocking the complete packet handler thread (and thus all event processing on that shard) for 100 milliseconds to wait for the channel request response, and you only send 5 DMs per second or need to dispatch a large number of DMs at once, you'd block shard 0 for a relatively long time from processing anything else - and you reach that number very easily on large bots as you're sending most DMs from other shards. Additionally, you'd consume twice as many requests from the already very low global ratelimits than actually needed.

Instead, we should think about not dispatching a channel create event for DMs at all. That way, we can dispatch the message create event asynchronously, and prevent blocking the handler thread. We'd need to make the channel optional, but can add a new method like "getChannelId()" to the Message interface.

We're wrapping Discords API, and they stopped dispatching channel create events for DM, and so should we; they have done it in v6 to support legacy clients only (that worked with the discontinued bot DM list). It would be a breaking change but we have a very breaking change release ahead anyways as we're switching to v8.

Vampire:
Switching to v8 per-se is not a breaking change, but an implementation detail unless this causes changes in the Javacord API.
Making the channel `Optional` was my other suggestion how to deal with it, but Joshix didn't like it.

Not dispatching the channel create event would be a breaking change behaviour-wise, but it could certainly be done.
But that then also needs to be clearly communicated in the release notes and the channel made `Optional`, maybe adding a `requestChannel` method like we have it for other places.

I'd second that move actually, I just tried to get the code here correct behaviour-wise as Joshix didn't like the `Optional` idea.

I guess @Bastian should share a thought about it.
But probably the `Optional` road and no channel create event is indeed the way to go.

Vampire:
@Joshix-1 what do you think about the last two comments and changing to the `Optional` way?

Joshix-1:
I agree.
How would you make Message#isPrivateMessage() work then? I don't think it is a good idea to just assume that it is private if the channel is missing because of webhooks. I think there can be some changes now that the channel will be Optional.
https://github.com/Javacord/Javacord/blob/19bf2f31110ebd098b3a4ab315a533757a4be772/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java#L893-L895

coaxial1:
I guess we would need to work with the `guild_id` sent by Discord in the message payload to determine if it's private or not. iirc, group chats are only available for user bots and we should remove that feature entirely anyways ( #576 )

Joshix-1:
I think making the channel Optional is the way to go. (I spend way too much time making the channel optional.)

It's currently in a "it build without errors" state and needs testing.

coaxial1:
I understand that, and it looks good. Nevertheless, that alone shouldn't be the most important factor when choosing a solution. :)

I just think this: 
- Making the channel optional everywhere is a breaking change that will impact more code than an optional recipient. You probably call ``Message#getChannel`` more frequently than ``PrivateChannel#getRecipient`` (tbh, I wrote 70k lines of code with Javacord and didn't even know it existed 😬 ). For a lot of work done with Javacord it'd probably just be a minimal change, if any at all
- If we'd just make the recipient optional, we'd have a more justifiable, straight-forward solution for the missing channel create event on the gateway.
- It's hard to tell a user why a channel can be optional / why a message doesn't have a channel in some cases, which doesn't make much sense from a users perspective - it might cause confusion when a channel is clearly there in all cases (on guilds), but still comes as an Optional, although the only case where it's empty may be in DMs
- In contrary, it's easy to explain why a recipient may be null (e.g. by just showing what we got from the gateway, as we simply don't have it).
- With an optional recipient, we could even still deliver a channel create event with an empty recipient field, which is not the case in the current solution. In the rare case that someone is using it, we could still provide it, and provide a method to fetch the recipient if required.
- With an optional recipient, we would already have the message we received in the cache, while with the current solution, the message would not be cached.

I apologize that I come up with this "idea" after you worked on the "optional channel" solution - just trying to find the most practical one for everyone.

What do you guys think? :)

Joshix-1:
Making the channel optional has the advantage that we can always return the message when sending messages with webhooks. Otherwise it wouldn't be possible to get the object when the bot isn't on the server with the message. 
This is a rare problem, but it could get more important with slash commands.

But you are right; making the recipient optional, would be the better way for this pr.

coaxial1:
We should also check if we can resolve #524 with this PR as it's a similar problem... 🤔

coaxial1:
> This is a rare problem, but it could get more important with slash commands.

I don't see an issue there (yet). I do not see a reason to start an interaction from anywhere else than a server the bot is in; and I don't think Javacord should even try to support "interactions only", that's not what it's for, imho. We should see slash commands as an guild based add-on, so we probably always have the server text channel for the interaction cached. Do I oversee something obvious? :)

Joshix-1:
> We should also check if we can resolve #524 with this PR as it's a similar ekzem... thinking 
Would #524 get fixed with the recipient being optional?

> so we probably always have the server text channel for the interaction cached.
Maybe, I don't know. Do the interaction create events respect shards?

coaxial1:
> Would #524 get fixed with the recipient being optional?
Well, no, but since we're trying to compensate the missing channel create event, we could fix that reaction issue as well. It's caused by the same thing: There's no channel create event preceding the reaction add event.

> Maybe, I don't know. Do the interaction create events respect shards?
Yes, if you opt to receive them through the gateway, they come in on the shard of the server the interaction has been initiated on. Only the other possibility (receiving interactions through a HTTP call to your own server / VM) is shard agnostic, but is intended to be used by interaction only applications that do not have a bot account and active gateway sessions

Joshix-1:
I made the recipient optional now. (not tested yet, just a draft)
This pr now tries to fix the problems with reactions (#524) aswell.

The version with the optional channel is at: https://github.com/Joshix-1/Javacord/tree/optional_channel , but I think it is a better idea to don't make the channel optional.

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionRemoveAllHandler.java
```diff
@@ -29,22 +31,31 @@ public MessageReactionRemoveAllHandler(DiscordApi api) {
...
-        });
+        long messageId = packet.get("message_id").asLong();
+        Optional<Message> message = api.getCachedMessageById(messageId);
+
+        message.ifPresent(msg -> ((MessageImpl) msg).removeAllReactionsFromCache());
+
+        long channelId = packet.get("channel_id").asLong();
+        TextChannel channel = api.getTextChannelById(channelId).orElse(null);
+        if (channel == null) {
+            if (!packet.hasNonNull("guild_id")) { // if channel null and private channel:
```
coaxial1, line 42: I'd suggest to return early if `packet.hasNonNull("guild_id")` is `true`, so we could get rid of `else`

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionRemoveHandler.java
```diff
@@ -31,32 +33,42 @@ public MessageReactionRemoveHandler(DiscordApi api) {
...
-                    userId,
-                    event);
-        });
+        long messageId = packet.get("message_id").asLong();
+        long userId = packet.get("user_id").asLong();
+        Optional<Message> message = api.getCachedMessageById(messageId);
+
+        long channelId = packet.get("channel_id").asLong();
+        TextChannel channel;
+        if (!packet.hasNonNull("guild_id")) { // if private channel:
```
coaxial1, line 42: I'd swap the `if` and `else` block here, and remove the `!` from the condition

coaxial1 (changes requested):
Wow, this looks good and very clean.

I added a few very minor comments and suggestions, if you don't mind. Please feel free to accept/reject them as you see fit.

I'll do a few manual tests tomorrow.

Joshix-1:
I updated it and it seems to work.

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionAddHandler.java
```diff
@@ -34,37 +35,49 @@ public MessageReactionAddHandler(DiscordApi api) {
...
+        if (serverId == null) { // if private channel:
+            channel = PrivateChannelImpl.getOrCreatePrivateChannel(api, channelId, userId, null);
+        } else {
+            channel = api.getTextChannelById(channelId).orElse(null);
+        }
 
-            message.ifPresent(msg -> ((MessageImpl) msg).addReaction(emoji, userId == api.getYourself().getId()));
+        if (channel == null) {
+            return;
+        }
```
KILLEliteMaste, line 52: I assume this should never happen since the event will get lost. Please add a logger message

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionRemoveAllHandler.java
```diff
@@ -29,22 +31,31 @@ public MessageReactionRemoveAllHandler(DiscordApi api) {
...
+        Optional<Message> message = api.getCachedMessageById(messageId);
+
+        message.ifPresent(msg -> ((MessageImpl) msg).removeAllReactionsFromCache());
+
+        long channelId = packet.get("channel_id").asLong();
+        TextChannel channel = api.getTextChannelById(channelId).orElse(null);
+        if (channel == null) {
+            if (packet.hasNonNull("guild_id")) {
+                return; // we don't know anything about the channel as it is part of a server and not cached
+            }
```
KILLEliteMaste, line 44: Same here. Add a logger message since all server channels should be cached

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionRemoveHandler.java
```diff
@@ -31,32 +33,42 @@ public MessageReactionRemoveHandler(DiscordApi api) {
...
+        TextChannel channel;
+        if (!packet.hasNonNull("guild_id")) {
+            channel = api.getTextChannelById(channelId).orElse(null);
+        } else { // if private channel:
+            channel = PrivateChannelImpl.getOrCreatePrivateChannel(api, channelId, userId, null);
+        }
+
+        if (channel == null) {
+            return;
+        }
```
KILLEliteMaste, line 50: Same here. Add a logger message

Joshix-1:
I added logging messages and removed the blank lines.

lost-illusi0n:
Just tried this fix (on commit 80285c8) and it seemed to work... until it didn't. 
Not sure what exactly caused it but after a few minutes (after responding and handling messages fine, both in DMs and servers), it had stopped handling messages from servers correctly. I could still DM the bot, and the bot could DM me, but it was not able to do so with messages from servers. Here's the assertion error I was getting spammed with after a few minutes.
```
23:26:23.316 [Javacord - Handlers Processor] WARN  org.javacord.core.util.gateway.PacketHandler - Couldn't handle packet of type MESSAGE_CREATE. Please contact the developer! (packet: {"type":19,"tts":false,"timestamp":"2021-03-27T23:26:23.258000+00:00","referenced_message":{"type":0,"tts":false,"timestamp":"2021-03-27T23:25:45.658000+00:00","pinned":false,"mentions":[],"mention_roles":[],"mention_everyone":false,"id":"825510941301932044","flags":0,"embeds":[],"edited_timestamp":null,"content":"alright boys and girls whos gonna sign up for the neuralink","channel_id":"110373943822540800","author":{"username":"GoatGeek","public_flags":0,"id":"96979199411113984","discriminator":"0001","avatar":"a_4c5080f0287d6f8340b4206a056ac479"},"attachments":[]},"pinned":false,"nonce":"825511098684801024","message_reference":{"message_id":"825510941301932044","guild_id":"110373943822540800","channel_id":"110373943822540800"},"mentions":[],"mention_roles":[],"mention_everyone":false,"member":{"roles":["110375768374136832","113379036524212224","231130892926582786","234237413675892736","407326634819977217","499303630567112704","520136799134416896","803324278987685888","344727499864932354","818750459350482954","530106314752524299","755499785020113036","585535347753222157","588728844345671696","540025225597812762","598574793712992286","801105321456173086","801550928108191777","746739847145717800","753295484973482055"],"premium_since":"2020-08-16T18:01:15.677000+00:00","pending":false,"nick":"amy ✨","mute":false,"joined_at":"2017-01-24T01:45:38.514000+00:00","is_pending":false,"hoisted_role":"234237413675892736","deaf":false},"id":"825511099008024598","flags":0,"embeds":[],"edited_timestamp":null,"content":"Sorry I don't want the government stealing my ABSOLUTELY PROPRIETARY thoughts","channel_id":"110373943822540800","author":{"username":"amy","public_flags":131584,"id":"128316294742147072","discriminator":"0001","avatar":"a_1d97b4b1fa012fb4438b96256c4d10da"},"attachments":[],"guild_id":"110373943822540800"})
java.lang.AssertionError: null
	at java.util.Optional.orElseThrow(Optional.java:403) ~[?:?] "
	at org.javacord.core.entity.message.MessageAuthorImpl.<init>(MessageAuthorImpl.java:55) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at org.javacord.core.entity.message.MessageImpl.<init>(MessageImpl.java:168) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at org.javacord.core.DiscordApiImpl.lambda$getOrCreateMessage$29(DiscordApiImpl.java:1092) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at java.util.Optional.orElseGet(Optional.java:364) ~[?:?] "
	at org.javacord.core.DiscordApiImpl.getOrCreateMessage(DiscordApiImpl.java:1092) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at org.javacord.core.util.handler.message.MessageCreateHandler.handle(MessageCreateHandler.java:56) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at org.javacord.core.util.handler.message.MessageCreateHandler.lambda$handle$0(MessageCreateHandler.java:51) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at java.util.Optional.ifPresent(Optional.java:178) ~[?:?] "
	at org.javacord.core.util.handler.message.MessageCreateHandler.handle(MessageCreateHandler.java:50) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at org.javacord.core.util.gateway.PacketHandler.lambda$handlePacket$0(PacketHandler.java:51) ~[javacord-core-3.1.3-SNAPSHOT.jar:?] "
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?] "
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?] "
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) [?:?] "
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) [?:?] "
	at java.lang.Thread.run(Thread.java:831) [?:?] "
```
Reverting this PR seemed to just go back to its original state.

Joshix-1:
I don't see how this PR should cause this. Does the development branch work for you? It seems weird that this PR should cause this, as I didn't change anything with the MessageAuthor.

lost-illusi0n:
Yes, I'm using the development branch along with a 2 line change within AudioSocket, and then I pulled this PR. It seemed to be fine on my dev bot, and when I then pushed the update to my production version, it had started spamming the above exception after a few minutes. Maybe it could be some freak one-time Discord thing?

lost-illusi0n:
Seemed to resolve it self *hopefully*, and I'm leaning on that this was some freak discord bug in combination with bad timing. It worked locally overnight, and is now running in production 👍

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionRemoveHandler.java
```diff
@@ -31,32 +40,43 @@ public MessageReactionRemoveHandler(DiscordApi api) {
...
+        long userId = packet.get("user_id").asLong();
+        Optional<Message> message = api.getCachedMessageById(messageId);
+
+        long channelId = packet.get("channel_id").asLong();
+        TextChannel channel;
+        if (!packet.hasNonNull("guild_id")) {
+            channel = api.getTextChannelById(channelId).orElse(null);
+        } else { // if private channel:
+            channel = PrivateChannelImpl.getOrCreatePrivateChannel(api, channelId, userId, null);
+        }
```
KILLEliteMaste, line 53: This causes the error @lost-illusi0n commented. Remove the `!` and it should work again

Joshix-1: I removed the `!`.

KILLEliteMaste:
Can you please squash the PR into a single commit?

KILLEliteMaste:
lgtm! Can you please rebase to the latest version?

PR #711, opened 2021-02-11
Fix SelfMuted and SelfDeafened status and events

### LINKED ISSUE #696: User#isSelfMuted/isSelfDeafened using server mute/deafen state instead of self mute/deafen state
Member creation is currently using the `mute` and `deafen` fields from the member object for self mute/deafen status which actually refer to server mute/deafen status
The correct initial states should be gotten from the `voice_states` array (which also depends on the `GUILD_VOICE_STATES` intent being enabled)
These states are also not being updated by `VOICE_STATE_UPDATE` events, though the event is dispatched

Aninoss:
This should fix #696

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
@@ -74,6 +74,70 @@ public interface ServerVoiceChannel extends ServerChannel, VoiceChannel, Categor
     }
 
     /**
+     * Gets the ids of the users that are self muted in this server voice channel.
+     *
+     * @return The ids of the users that are self muted in this server voice channel.
+     */
+    Collection<Long> getSelfMutedUserIds();
+
+    /**
+     * Gets the users that are self muted in this server voice channel.
+     *
+     * @return The users that are self muted in this server voice channel.
+     */
+    Collection<User> getSelfMutedUsers();
+
+    /**
+     * Checks whether the user with the given id is self muted in this channel.
+     *
+     * @param userId The id of the user to check.
+     * @return Whether the user with the given id is self muted in this channel or not.
+     */
+    boolean isSelfMuted(long userId);
+
+    /**
+     * Checks whether the given user is self muted in this channel.
+     *
+     * @param user The user to check.
+     * @return Whether the given user is self muted in this channel or not.
+     */
+    default boolean isSelfMuted(User user) {
+        return isSelfMuted(user.getId());
+    }
+
+    /**
+     * Gets the ids of the users that are self deafened in this server voice channel.
+     *
+     * @return The ids of the users that are self deafened in this server voice channel.
+     */
+    Collection<Long> getSelfDeafenedUserIds();
+
+    /**
+     * Gets the users that are self deafened in this server voice channel.
+     *
+     * @return The users that are self deafened in this server voice channel.
+     */
+    Collection<User> getSelfDeafenedUsers();
+
+    /**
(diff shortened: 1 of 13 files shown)


Vampire:
I just gave it a very quick skim, but why so many changes, especially structurally? It seems you make it a state of the channel, while it should be a state of the server, shouldn't it? Just the same as deafend and muted state. It worked fine when I implemented it, it was either broken by the audio changes or more likely by the intents / user cache changes. I would have expected maybe a 3 or 4 lines change to fix that.

Joshix-1:
I think the fix is fine. The old logic seemed flawed.
https://discord.com/developers/docs/resources/voice

Aninoss:
Yes, all these changes are necessarily unfortunately. But at least it shouldn't break any code, the api is mostly the same with just some additional methods in the ServerVoiceChannel class

Vampire:
But why is it now a state of the channel instead of the server like before?
And if really necessary, why were deafened and muted state not also necessary on the channel instead of on the server?

@Joshix-1 that link in no way explains why you think the old logic was flawed or why the fix is fine, can you elaborate a bit?
It worked perfectly fine when I implemented it.

I still think this change is wrong.
You can pretty well self-deafen and self-mute or be deafened and muted without being in any voice channel.
And the states also carry over if you change channels or disconnect and reconnect from / to audio.
Those properties are properties of the server, **not** the voice channel you are connected to.
A `VOICE_STATE_UPDATE` does not necessarily refer to any channel, the `channel` field can be `null` if you disconnected from audio. The `channel` field is just the currently connected to channel but does not directly relate to the other properties of that event necessarily.

Aninoss:
But the current method doesn't work at all, neither in my test bot nor in production (all intents are active), so I think relying on the server state for self mute / self deaf would give inaccurate results. Regular mute / deaf seems to be working fine though

Vampire:
Yes, as I said in the issue and also here, it got broken.
It worked perfectly fine when I implemented it.
And most probably the changes for user cache and intents broke it.
But that doesn't mean that the general approach was wrong, just that Bastian probably changed something in a wrong way so that it doesn't work anymore.
The fix is probably to change 2 or 3 lines that he wrongly changed in his commit and it should work again just fine.

Joshix-1:
> And most probably the changes for user cache and intents broke it.
I don't think that is the case. There are no documented changes i know of that could cause your code to not work.

Vampire:
Then it maybe were the changes for audio.
The point is, that these states worked perfectly fine when I implemented them and Discord-wise nothing changed.

Joshix-1:
> relying on the server state for self mute / self deaf would give inaccurate result

That's the problem with your implementation.
We need to use the Voice State Object https://discord.com/developers/docs/resources/voice

Vampire:
But I still think it was the user cache / intent changes.
For example the `Member` class takes the `selfDeaf` and `selfMute` state from the fields `deaf` and `mute`.
And the `Member` class did not exist before those changes.

I don't talk about changes on Discord side which you probably assume as you say "documented changes".
I talk about the changes in Javcord code that Bastian did.

Vampire:
> That's the problem with your implementation.
> We need to use the Voice State Object https://discord.com/developers/docs/resources/voice

Which is exactly what my implementation did

Vampire:
If you look at the code in 3.0.7 you can see how it works perfectly fine.
Before Bastian broke it by mixing deaf/mute with selfdeaf/selfmute.
The correct fix is probably to use the proper fields from the json in the member constructor and add fields for deaf/mute besides selfdeaf/selfmute.

Vampire:
Properly solved in #736

PR #731, opened 2021-04-04
Fixed getSlowmodeDelayInSeconds returning the reversed result

### LINKED ISSUE #715: ServerTextChannel.hasSlowmode is reversed
The condition at https://github.com/Javacord/Javacord/blob/8608eae1d5b33a9c493f98052ee23b453233721c/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerTextChannel.java#L118 should be `!= 0` or `> 0`

KILLEliteMaste:
Fixed #715

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerTextChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerTextChannel.java
@@ -118,7 +118,7 @@ public interface ServerTextChannel extends RegularServerChannel, TextChannel, Me
      * @return Whether this channel enforces a slowmode.
      */
     default boolean hasSlowmode() {
-        return getSlowmodeDelayInSeconds() == 0;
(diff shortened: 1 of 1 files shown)


Joshix-1 (approved):
👍️

Bastian:
Is this really a breaking change?
Yes, it changes the behaviour of the method but after all, it is a bug fix and does not change the API definition.

KILLEliteMaste:
I would say yes. As it returns something else it breaks your code and you need to action and change something

Saladoc:
Not a breaking change. Api does not change, return values (finally) match the documentation. This is a bugfix.

KILLEliteMaste:
> Not a breaking change. Api does not change, return values (finally) match the documentation. This is a bugfix.

Even if we don't change the API, it is still a breaking change. Once this gets merged existing code like
```java
if(serverTextChannel.hasSlowMode()){
} else {}
```
doesn't work anymore like it's meant to be and the user needs to change the code.

PR #736, opened 2021-04-09
Fix self_mute and self_deaf handling

### LINKED ISSUE #696: User#isSelfMuted/isSelfDeafened using server mute/deafen state instead of self mute/deafen state

AyuAi:
Fixes #696 with as minimal changes as possible.

The `mute` and `deaf` packets are handled appropriately in the `VoiceStateUpdateHandler` already by passing off the values to the `ServerImpl`, so the `MemberImpl` just had to be updated to handle the `self_mute` and `self_deaf` packets (since those are properties of the object).

The new `Member` in `VoiceStateUpdateHandler#handleServerVoiceChannel()` also had to be put into / update the `MemberCache`.


AyuAi:
> With these changes I was not able to get any other state out of the methods besides `false`, so something is not yet right.
> 
> As already said in the other PR that tries to fix this (#711),
> if you look at the code in 3.0.7 you can see how it works perfectly fine and from which events you need to update the states when.
> Before Bastian broke it by mixing deaf/mute with selfdeaf/selfmute in the intent / user cache changes.
> 
> Btw. @Bastian can you shed some light why you have selfDeafend and selfMuted in `Member`, but `deafend` and `muted` in `Server`? Both settings are server-wide so I would have expected both at the same place and not spread.

Do you think doing so would fix the issue? It sure seems like it might, as Server#isMuted()/isDeafened() work perfectly fine.

Vampire:
Maybe it's the opposite and `muted` and `deafened` should instead be movd to the `Member`, as `Member` in my understanding is things specific for a user on a server and all four voice states match this criterion.

But I didn't really look at the changes Bastian made for the user-cache and intents, so I don't know how to properly fix it or I would have already done it.
All I can say is, that the code in 3.0.7 was working perfectly fine in all situations, because I wrote and tested it. :-)

javacord-core/src/main/java/org/javacord/core/entity/user/MemberImpl.java
```diff
@@ -72,13 +72,13 @@ public MemberImpl(DiscordApiImpl api, ServerImpl server, JsonNode data, UserImpl
...
-            selfDeafened = data.get("deaf").asBoolean();
+        if (data.hasNonNull("self_deaf")) {
+            selfDeafened = data.get("self_deaf").asBoolean();
         } else {
             selfDeafened = false;
         }
-        if (data.hasNonNull("mute")) {
-            selfMuted = data.get("mute").asBoolean();
+        if (data.hasNonNull("self_mute")) {
+            selfMuted = data.get("self_mute").asBoolean();
```
Mysterypotatoguy, line 81: The field name was previously correct on these but the data pulled from them was not, these fields refer to the _server_ muted/deafened state.
The `self_mute` and `self_deaf` fields exist in the `voice_states` fields in the `GUILD_CREATE` packet

Bastian:
> Btw. @Bastian can you shed some light why you have selfDeafend and selfMuted in Member, but deafend and muted in Server? Both settings are server-wide so I would have expected both at the same place and not spread.

I don't remember having a particular reason for it. The member PR was a bit rushed because of the deadline for mandarory intents by Discord. Having everything in the member object definitely sounds like the way to go.

AyuAi:
I would like to note here for book-keeping that I'm not sure I can solve this issue more elegantly than I have done so far with adding extra constructors to MemberImpl and UserImpl to take Member/User objects respectively. It makes it pretty difficult to use the data from the VOICE_STATE_UPDATE (which may or may not have a given Member field) to update the member cache if 

1. I cannot easily use the Member in the member cache to create a new MemberImpl
2. I cannot easily make a UserImpl from a User object (I think this is necessary because of the same reason as point 1, where I need to be able to make a MemberImpl from the cache but MemberImpl requires a UserImpl to construct).

I'm looking for some opinions on the current state of these changes, especially since it's been getting a bit more work-aroundy than it was in the beginning.

I think this problem could be solved if the Member/UserImpls were mutable and if the reading of the data from Discord was simply done in the handlers; This would make Javacord more ready for if Discord has more events in the future that may not give you everything you need to construct a Member but might give very useful information about updating the user.

AyuAi:
I tested self_mute, self_deaf, mute, and deaf; As far as my testing went, this commit doesn't have any bugs, so I think I'm ready for review again.

AyuAi:
Ok I'm dumb and didn't understand the impl setup. Fixing commit

AyuAi:
As an aside, this bugfix implementation was built with the immutability of MemberImpl and UserImpl in mind, but I would like to note that I think it would be better if they were mutable internally (and stay immutable when a Javacord method returns a User to retain the immutability forward-facing) because I think it would be easiest to update the cache this way in the future, especially if Discord adds more events that update User/Member values but don’t pass along the original User/Member being updated (and only supply the ID). For ex: The `VOICE_STATE_UPDATE`

Maybe @Bastian has opinions on changing UserImpl/MemberImpl to being mutable and updating the cache via setters (instead of reinserting new objects)? I’m more than willing to work on an implementation later if that’s a foreseeably better direction to take Javacord’s cache handling, but I am not sure since I believe that the cache is being reworked by you guys so I don’t think I’m needed to do anything there.

javacord-core/src/main/java/org/javacord/core/util/handler/guild/VoiceStateUpdateHandler.java
```diff
@@ -105,7 +105,43 @@ private void handleSelf(JsonNode packet) {
...
+                        // If there is no member in the cache and we don't receive a member field,
+                        // we will log it and return. I'm actually not sure if this is normal behavior;
+                        // Maybe the logging should be removed if it is, but the data being lost (non-cacheable)
+                        // is still notable.
+                        logger.warn("Received VOICE_STATE_UPDATE packet without "
+                                + "member field or member in the cache: {}", packet);
+                        return;
+                    }
+
+                    // For the update, prioritizes using the Member in the cache;
```
AyuAi, line 132: I'm not sure if this is the right move, but it feels the most right; I think this because `VOICE_STATE_UPDATE`s are meant to convey new voice state data, not Member data, so using them to update the Member cache may be unreliable. When I was testing the code, I'm pretty sure that I found that the Member object provided through the `VOICE_STATE_UPDATE` had the old voice states; To me, it feels like it is basically worthless to even have the Member object here for anything but to try and populate the Member cache if it is empty, because we aren't even guaranteed the Member object .

With this in mind, I took the approach of simply updating the Member object that already exists in the cache, with the option of adding the `VOICE_STATE_UPDATE` Member to the cache only if the Member is not present in the cache. No matter which way it was done, the Member object had to be overridden with the new packet's voice states (self mute, self deafen, server mute, server deafen) to work properly.

AyuAi:
Repost/reword from Javacord Discord for PR clarity:

@Vampire I went ahead and rebased the PR over main; It doesn't seem like many changes happened in between to the code that matters here. Up to your discretion on if the unit tests that exist test the functionality of `MemberImpl` well enough (I am not sure, haven't checked), but I'm open to live retesting the bugfix if not. 

I do ask that this one doesn't go stale for 10 months again though 😆

Vampire (approved):
LGTM, thanks for the contribution and sorry for the big delay

PR #743, opened 2021-04-20
Fix failing channel update for SERVER_NEWS_CHANNEL

KILLEliteMaste:


javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -323,7 +323,9 @@ private void handleServerTextChannel(JsonNode jsonChannel) {
         }
 
         int oldSlowmodeDelay = channel.getSlowmodeDelayInSeconds();
-        int newSlowmodeDelay = jsonChannel.get("rate_limit_per_user").asInt(0);
+        //Check if "rate_limit_per_user" exists as a temporary fix until SERVER_NEWS_CHANNEL is handled separately.
+        int newSlowmodeDelay = jsonChannel.has("rate_limit_per_user")
```
Joshix-1, line 327: I would use hasNonNull here as discord can't be trusted and that's what Bastian always suggests using.

KILLEliteMaste: This code if effectively the same as `hasNonNull` as it will get the default value provided in `asInt` if it's value is null

PR #745, opened 2021-04-21
Fixed an error when following an announcement channel

### LINKED ISSUE #744: Following announcement channels causes javacord to not handle Message_Create properly
https://github.com/Javacord/Javacord/blob/f825905377ee30f8614190aef55da0e6b629da7f/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java#L223 (referring to line according to helper on the discord)

Confirmed this by testing it in a more controlled environment on my discord server with a new bot, using Javacord 3.3.0.

How to reproduce:
- Go on a server with an announcement channel
- Follow that servers announcement channel on a server where a bot is on that uses Javacord 3.3.0 (also uses the MessageCreateListener), newest version.

 Error Message upon reproduction:
 [Pastebin containing said error message](https://pastebin.com/1zAnibDp)
 
 Resolution: 
 According to the helper on the discord, the announcement following message
> **Username** has added **servername #news** to this channel. Its most important updates will show up here.

Does not contain a `message_id`, thus causing that error.

KILLEliteMaste:
Fixes #744 
This also adds the message reference object, which is available when receiving cross-posted / announcement /pin / reply messages which contains some information about the message's origin

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -43,9 +43,8 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Returns a {@code MessageBuilder} according to this {@code Message}.
      *
-     * @see MessageBuilder#fromMessage(Message)
-     *
      * @return The {@code MessageBuilder}.
+     * @see MessageBuilder#fromMessage(Message)
      */
     default MessageBuilder toMessageBuilder() {
         return MessageBuilder.fromMessage(this);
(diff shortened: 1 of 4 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -767,12 +765,13 @@ default URL getLink() throws AssertionError {
...
-     * Only present if this message is type {@code MessageType.REPLY}.
+     * Gets the message reference.
+     * The reference to another message for instance a cross-posted / channel follow / pin / reply message.
      *
-     * @return The id of the referenced message.
+     * @return The message reference.
+     * @see <a href="https://discord.com/developers/docs/resources/channel#message-reference-object">Discord docs for message reference</a>
      */
-    Optional<Long> getReferencedMessageId();
+    Optional<MessageReference> getMessageReference();
```
Joshix-1, line 774: I would suggest re-adding getReferencedMessageId() as utility-method to make the pr not breaking.

KILLEliteMaste: On the one side I agree with you but on the other side I think re-adding this method just adds a useless method and doesn't provide any benefit except just having this not being a breaking change. But if you or others insist to re-add it, I can do that.

PR #749, opened 2021-04-29
Fix global ratelimiter math and correct / update javadocs

coaxial1:
I've fixed the bad math of the default global ratelimiter as discussed on Discord.

I've replaced the recommended formula to calculate the bucket duration, because to be honest, the old one was just wrong or I did not understand it correctly. I'm open for discussion, but the one I suggested will work (see examples in this table).
grafik

Additionally, I updated and corrected a few documentation lines around the global ratelimiter. For example, a documentation in the builder said that Javacord will not use a global ratelimiter if none is set by the users, which is wrong (we have default global ratelimiters).

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
+++ b/javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
@@ -18,19 +18,19 @@ import java.time.Duration;
  * </ul>
  * For the 50 requests / 1 second ratelimit, these rules allow the following ratelimits (only a subset):
  * <ul>
- * <li>About {@code 1} request per {@code 20.5 ms}
- *     ({@code amount = 1} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(1D / 49D))})
- * <li>About {@code 5} request per {@code 111.1 ms}
- *     ({@code amount = 5} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(5D / 45D))}).
+ * <li>About {@code 1} request per {@code 21 ms}
+ *     ({@code amount = 1} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(1000D / 49D))})
+ * <li>About {@code 5} request per {@code 112 ms}
+ *     ({@code amount = 5} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(5000D / 45D))}).
  * <li>{@code 10} request per {@code 250 ms}
- *     ({@code amount = 10} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10D / 40D))}).
+ *     ({@code amount = 10} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10000D / 40D))}).
  * <li>{@code 25} request per {@code 1 sec}
- *     ({@code amount = 25} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10D / 40D))}).
+ *     ({@code amount = 25} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(25000D / 25D))}).
  * </ul>
  * Choosing a lower {@code amount} increases the maximum throughput but can limits your ability to perform actions
  * in bulk.
  *
- * @see <a href="https://javacord.org/r/local-ratelimiter">Related wiki article</a>
+ * @see <a href="https://javacord.org/wiki/advanced-topics/ratelimits.html">Related wiki article</a>
  * @see DiscordApiBuilder#setGlobalRatelimiter(Ratelimiter)
  */
 public class LocalRatelimiter implements Ratelimiter {
@@ -125,7 +125,7 @@ public class LocalRatelimiter implements Ratelimiter {
         }
 
         // Reset the limit when the last reset timestamp is past
-        if (System.nanoTime() > nextResetNanos) {
+        if (System.nanoTime() >= nextResetNanos) {
             remainingQuota = amount;
             try {
                 nextResetNanos = System.nanoTime() + bucketDuration.toNanos();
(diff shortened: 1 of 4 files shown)


javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
```diff
@@ -14,21 +14,23 @@
  * <ul>
  * <li>Choose an {@code amount} that is between {@code 1} and half of the real amount
  *     (in this example {@code 1} - {@code 25}).
- * <li>Calculate the {@code bucketDuration} using {@code (amount * realDuration) / (realAmount - amount)}.
+ * <li>Calculate the {@code bucketDuration} using {@code realDuration / (realAmount / amount)}.
  * </ul>
  * For the 50 requests / 1 second ratelimit, these rules allow the following ratelimits (only a subset):
  * <ul>
  * <li>About {@code 1} request per {@code 20.5 ms}
```
Vampire, line 21: The values do not match the new results

javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
```diff
@@ -14,21 +14,23 @@
...
  * <li>{@code 10} request per {@code 250 ms}
- *     ({@code amount = 10} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10D / 40D))}).
+ *     ({@code amount = 10} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(1000D / (50D / 10D)))}).
  * <li>{@code 25} request per {@code 1 sec}
- *     ({@code amount = 25} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10D / 40D))}).
+ *     ({@code amount = 25} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(1000D / (50D / 25D)))}).
  * </ul>
  * Choosing a lower {@code amount} increases the maximum throughput but can limits your ability to perform actions
  * in bulk.
+ * In general, it's recommended to set the ratelimit a bit lower than the actual maximum to make up for network jitter.
```
Vampire, line 32: If that is recommended, why not incorporate it in the suggested formula for example by using `0.9 * realAmount`?

Vampire (changes requested):

javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
coaxial1: I think this would cause confusion. Imho, the given formula should yield exact results for the given inputs. If I say I want "exactly 50 requests per 1 second", the formula should exactly do that.
If a user wants to follow the advice, they can just lower `realAmount` to whatever the see fit. I'd generally say that someone who sets a custom ratelimiter should be advanced enough to understand and implement that advice.

There may even be use cases where that recommendation is exactly what you don't want since our `Ratelimiter` interface and this implementation are very "generic" and could be reused in more scenarios than just global ratelimiting.
We could either just remove it, or replace `In general` with `For global ratelimiting`.

What do you think?

Bastian:
> because to be honest, the old one was just wrong or I did not understand it correctly

The formula exists to prevent misalignment with Discord global ratelimit bucket. Here's a simple example:
image

Setting the `LocalRatelimiter` to allow 25 req / 0,5 sec would make it possible to execute up to 150 requests in less than 1 sec (because the "first" ones could be at the "end" of the LocalRatelimiter bucket and then the "last" ones at the "start" of the LocalRatelimiter bucket).

Vampire:
Ah, so we need the old formula, but correctly implemented and maybe explained as comment why.

coaxial1:
Got it, thanks for the explanation.

coaxial1:
Should be good now.

coaxial1:
We should add the explanation why we need this to a wiki article.
The linked page https://javacord.org/r/local-ratelimiter does currently resolve to 404.

coaxial1:
I'd suggest to change the link to https://javacord.org/wiki/advanced-topics/ratelimits.html and create another PR to talk about global ratelimits on that wiki page. Opinions?

javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
Vampire: Actually the values are still not correct, as a `ceil` is done there should be no digits after the point

javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
```diff
@@ -19,18 +19,18 @@
...
+ *     ({@code amount = 1} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(1000D / 49D))})
  * <li>About {@code 5} request per {@code 111.1 ms}
- *     ({@code amount = 5} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(5D / 45D))}).
+ *     ({@code amount = 5} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(5000D / 45D))}).
  * <li>{@code 10} request per {@code 250 ms}
- *     ({@code amount = 10} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10D / 40D))}).
+ *     ({@code amount = 10} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10000D / 40D))}).
  * <li>{@code 25} request per {@code 1 sec}
- *     ({@code amount = 25} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(10D / 40D))}).
+ *     ({@code amount = 25} and {@code bucketDuration = Duration.ofMillis((long) Math.ceil(25000D / 40D))}).
```
Vampire, line 28: This should be `/ 25D`, shouldn't it?

Vampire (changes requested):

javacord-api/src/main/java/org/javacord/api/util/ratelimit/LocalRatelimiter.java
coaxial1: Yup.

Vampire (approved):

PR #750, opened 2021-05-13
Interaction response

KILLEliteMaste:
This implements the ability to respond to interactions.
Also fixed a hard coded limit for webhook embeds in the message builder.

Internal PR implementations:
* [x] Responding to an interaction
* [x]  Followup messages

Issue implementations:
* [x] Respond to application commands

Usage example:
The most basic response to receiving an interaction is by simply sending your initial response.
```java
api.addInteractionCreateListener(event -> {
    //Both are the same
    event.respond().setContent("Initial Response").sendInitialResponse(event.getInteraction());
    new InteractionMessageBuilder().setContent("Initial Response").sendInitialResponse(event.getInteraction());
    //Delete your initial response
    event.deleteInitialResponse();
});
```
You can also hide your response from other user with setting the EPHEMERAL flag for the message
```java
api.addInteractionCreateListener(event -> {
    event.respond().setContent("Initial Response").setFlag(Flag.EPHEMERAL).sendInitialResponse(event.getInteraction());
});
```
It's also possible to respond to an interaction later but for a maximum of 15 minutes after the first time you have answered to an interaction with your first response or through #respondLater()
```java
api.addInteractionCreateListener(event -> {
    event.respondLater().thenAccept(unused -> {
        //Edit your original response. In this case you edit the "loading state" response
        new InteractionMessageBuilder().setContent("Deferred Response").editOriginalResponse(event.getInteraction());
        //
        //... 5minutes later
        //
        new InteractionMessageBuilder().setContent("Followup Message").sendFollowupMessage(event.getInteraction()).thenAccept(message -> {
            new InteractionMessageBuilder().setContent("Edited followup message").editFollowupMessage(event.getInteraction(), message.getId());
        });
    });
});
```

PR #758, opened 2021-05-27
Add Message Components

dylhack:
## TODO
- [x] Component Builders
- [x] Sending Components
- [x] Emoji Buttons
- [x] Handling websocket packets


dylhack:
I've decided to create this a draft now to receive input from other contributors what best practices to follow if incorrectly followed at any point of time.

dylhack:
Looking good so far... (working)
```kotlin
val button = ButtonBuilder()
    .setLabel("Hello world")
    .setCustomId("hello_world")
    .setStyle(ButtonStyle.Primary)

val actionRow = ActionRowBuilder()
    .addComponent(button);

val message = MessageBuilder()
    .setContent("Hello world")
    .addActionRow(actionRow)
    .send(channel)
    .join()
```

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ButtonBuilderDelegateImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        return toJsonNode(object);
+    }
+
+    /**
+     * Gets the button as a {@link ObjectNode}. This is what is sent to Discord.
+     *
+     * @param object The object, the data should be added to.
+     * @return The button as a ObjectNode.
+     */
+    public ObjectNode toJsonNode(ObjectNode object) {
```
coaxial1, line 115: I'd like to suggest to add a few sanity checks to this method and throw a IllegalStateExceptio  or something similar if they are violated. There are a few rules that a user must obey, and it'd be rejected by Discord anyways if they do not.

1. Style is not optional; Buttons without a style are not accepted
2. All non-link buttons must have a "custom_id"
3. All URL-style buttons must have a "url" and must not have a "custom_id"

javacord-core/src/main/java/org/javacord/core/entity/message/component/ActionRowImpl.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param data The json data of the component.
+     */
+    public ActionRowImpl(JsonNode data) {
+        super(ComponentType.ActionRow);
+
+        this.components = new ArrayList<>();
+        if (data.has("components")) {
+            for (JsonNode componentJson : data.get("components")) {
+                // When more components get added this must be changed to check
```
coaxial1, line 26: I'd suggest to do this in this PR already.
You can easily get the type through the JsonNode, and then just decode it into a Type since you already have a Enum for it. :)

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ActionRowBuilderDelegateImpl.java
```diff
@@ -0,0 +1,70 @@
...
+import org.javacord.api.entity.message.component.ComponentType;
+import org.javacord.api.entity.message.component.internal.ActionRowBuilderDelegate;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ActionRowBuilderDelegateImpl implements ActionRowBuilderDelegate {
+    private final ComponentType type = ComponentType.ActionRow;
+
+    private final List<ButtonBuilder> components = new ArrayList<>();
```
coaxial1, line 16: Shouldn't this collection consist of other "Components" rather than "Buttons"? There may be other components that fit into ActionRows in the future, and the docs explicitly say that ActionsRows may consist of any other component types.

You could use your "ComponentBuilder" interface to generalize this.

Of course, all other interfaces this implementation relates to should be adapted similarly.

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ActionRowBuilderDelegateImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public void addComponent(ButtonBuilder component) {
```
coaxial1, line 19: Here again, we should be able to add Components in general, not only Buttons. :)

coaxial1 (changes requested):
Hey!
Thank you very much for looking into this that quickly - awesome!
I've added a few comments you could look into - as always, all of it is open for discussion.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
```diff
@@ -398,6 +407,27 @@ public StringBuilder getStringBuilder() {
             body.put("avatar_url", avatarUrl.toExternalForm());
         }
 
+        if (embeds.size() != 0) {
```
coaxial1, line 410: This code must also be present in the regular textchannel send method above this method. Buttons will work with all kinds of messages, not only Webhooks.

Additionally, we need make sure that components work with Interaction Responses as well. Those have their own Message Builder implementation, see #750

coaxial1:
Another thing that is missing at the moment:
We need to enable users to reply to interactions spawned by users clicking a button. While this should already work if you want to replay with a message, Discord has introduced two response types that work with component interactions only (6 and 7) that allows us to immediately or later update the message the components are attached to:
https://discord.com/developers/docs/interactions/slash-commands#interaction-response-interactioncallbacktype

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonStyle.java
```diff
@@ -0,0 +1,76 @@
...
+            case 3:
+                return Success;
+            case 4:
+                return Danger;
+            case 5:
+                return Link;
+            default:
+                return null;
+        }
+    }
```
KILLEliteMaste, line 75: Please change the method name to either `fromId` or `fromValue` to keep more consistent with other enums styles.
Javacord **never** returns `null` in the api. So like in any other enum add another value `UNKNOWN` and return this instead.

javacord-api/src/main/java/org/javacord/api/entity/message/component/ComponentType.java
```diff
@@ -0,0 +1,42 @@
+package org.javacord.api.entity.message.component;
+
+public enum ComponentType {
```
KILLEliteMaste, line 3: Same as in the ButtonStyle enum

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+     * Set the button's style.
+     *
+     * @param styleValue The style of the button based on Discord's style values.
+     * @return The current instance in order to chain call methods.
+     * @link https://discord.com/developers/docs/interactions/message-components#buttons-button-styles
+     */
+    public ButtonBuilder setStyle(int styleValue) {
+        delegate.setStyle(styleValue);
+        return this;
+    }
```
KILLEliteMaste, line 51: Instead of having another method in the delegate, parse the value directly in the builder and pass it to `setStyle(ButtonStyle)`

KILLEliteMaste (changes requested):
Just had a quick look at it as coaxial1 already mentioned a few things

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonStyle.java
dylhack: Okay I've committed changes, see: 46a7a1f2a80a5947097e42bbe1e18a98b583b2a1

javacord-api/src/main/java/org/javacord/api/entity/message/component/ComponentType.java
dylhack: Okay I've committed changes, see: 46a7a1f

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ButtonBuilderDelegateImpl.java
```diff
@@ -0,0 +1,140 @@
...
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.javacord.api.entity.message.component.ButtonStyle;
+import org.javacord.api.entity.message.component.ComponentType;
+import org.javacord.api.entity.message.component.internal.ButtonBuilderDelegate;
+
+public class ButtonBuilderDelegateImpl implements ButtonBuilderDelegate {
+    private final ComponentType type = ComponentType.Button;
+
+    private ButtonStyle style = null;
```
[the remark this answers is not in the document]
dylhack: I've committed changes, see: 0d2d88bb61f26271be4c43cd6ae09e754dbdea7f

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonBuilder.java
dylhack: I've committed changes, see: 29140fc902cbe6811bb6d4e3cd1a811a5393470a

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ButtonBuilderDelegateImpl.java
dylhack: Awesome thanks I didn't quite know what to use if an attribute was populated properly.

dylhack: > All URL-style buttons must have a "url" and must not have a "custom_id"

Should we still throw an IllegalStateException if custom_id exists?

dylhack: I've committed changes, see: fbb9a0fa25eac42a8d960afd05020faa315b9fd5

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonStyle.java
KILLEliteMaste: 1. By convention enum values a written in full uppercase
2. Remove the trailling `,` at the end of the value definition
3. Assign the value `-1` to the unknown value as this is going to be least likely to be a value discord chooses.
The same applies of course for the ComponentType

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ActionRowBuilderDelegateImpl.java
dylhack: I've committed changes, see: e9bc4320a89c73aa606a04587d53a1cd26cf5060

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ActionRowBuilderDelegateImpl.java
dylhack: I've committed changes, see: e9bc4320a89c73aa606a04587d53a1cd26cf5060

javacord-core/src/main/java/org/javacord/core/entity/message/component/ActionRowImpl.java
dylhack: I've committed changes, see: 606bbaa024b7f57b8b562f9803e25eb4f19196e4

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonStyle.java
dylhack: review: 98f8fd68932c086ec764e932b6cf4c0067244a54

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ActionRowBuilderDelegateImpl.java
coaxial1: Thank you. From a first sight, I do not get why we need a generic type in the ComponentBuilder, tho. 
That seems kinda complex. Is there a way to avoid it? You should be fine working with using "getType", casting and inheritance only if I'm not completely wrong right now.

dylhack: I'll look again later today, but I believe I was having a casting issue

dylhack: Okay I've fixed the issue, ready for review: 408743e1455ea52eb694a61ac57992b636d9ec51

dylhack:
This approach for implementing the emoji property needs reviewed: a4f2afd124f5e672d49ed7e200bf45d2e916b50e

javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/ActionRowBuilderDelegateImpl.java
coaxial1: Perfect!

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
dylhack: Here is my approach to solve this.

A MessageBuilderDelegateImpl#prepareComponents method
https://github.com/dylhack/Javacord/blob/73bc3e3aec08437218a47152fb064e9a38421752/javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java#L549-L558
```java
private void prepareComponents(ObjectNode body) {
    if (components.size() != 0) {
        ArrayNode componentsNode = JsonNodeFactory.instance.objectNode().arrayNode();
        for (int i = 0; i < components.size() && i < 5; i++) {
            ActionRowBuilderDelegateImpl component = (ActionRowBuilderDelegateImpl) components.get(i).getDelegate();
            componentsNode.add(component.toJsonNode());
        }
        body.set("components", componentsNode);
    }
}
```

which is implemented here (send text channel):
https://github.com/dylhack/Javacord/blob/73bc3e3aec08437218a47152fb064e9a38421752/javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java#L338

and here (send webhook):
https://github.com/dylhack/Javacord/blob/73bc3e3aec08437218a47152fb064e9a38421752/javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java#L421

dylhack:
In my latest commit (bc5f11d41d93aef6b7dc8050ae405c9049b2d319) I added support for parsing the two new attributes of [ApplicationCommandInteractionData](https://discord.com/developers/docs/interactions/message-components#responding-to-a-component-interaction) (custom_id and component_type). I created a new class [InteractionComponentData](https://github.com/Javacord/Javacord/blob/bc5f11d41d93aef6b7dc8050ae405c9049b2d319/javacord-core/src/main/java/org/javacord/core/command/InteractionComponentDataImpl.java), but I would like a contributor to review this approach.

https://github.com/Javacord/Javacord/blob/bc5f11d41d93aef6b7dc8050ae405c9049b2d319/javacord-core/src/main/java/org/javacord/core/command/InteractionImpl.java#L52-L58

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
dylhack: After you approve of this, I'll work on responding

coaxial1: Yup, that's a solid and DRY approach 👍

coaxial1:
> In my latest commit ([bc5f11d](https://github.com/Javacord/Javacord/commit/bc5f11d41d93aef6b7dc8050ae405c9049b2d319)) I added support for parsing the two new attributes of [ApplicationCommandInteractionData](https://discord.com/developers/docs/interactions/message-components#responding-to-a-component-interaction) (custom_id and component_type). I created a new class [InteractionComponentData](https://github.com/Javacord/Javacord/blob/bc5f11d41d93aef6b7dc8050ae405c9049b2d319/javacord-core/src/main/java/org/javacord/core/command/InteractionComponentDataImpl.java), but I would like a contributor to review this approach.
> 
> https://github.com/Javacord/Javacord/blob/bc5f11d41d93aef6b7dc8050ae405c9049b2d319/javacord-core/src/main/java/org/javacord/core/command/InteractionImpl.java#L52-L58

That looks fine to me. The downside here is that a user would have to manually cast the "InteractionData", depending on the type, right? That is not generally a problem, but a very "Javacordy" thing to do would be to offer two methods here:

``InteractionData#asApplicationCommandInteractionData()`` that returns an ``Optional<ApplicationCommandInteractionData()`` if and only if the underlying Object is actually of this type; it should return an empty Optional otherwise.

The same thing could be added for components: ``InteractionData#asInteractionComponentData()``

You could just define these methods in your InteractionData interface, and implement them in your two data classes accordingly. That would save our users from a cast. This pattern is present in a lot of places in Javacord, for example in Emojis, in Channels, Users etc.

coaxial1:
I have a strong feeling that this PR will be great. In the end, we should definitely ask Bastian to review the final state, though, as it's pretty big and we might oversee something

dylhack:
> > In my latest commit ([bc5f11d](https://github.com/Javacord/Javacord/commit/bc5f11d41d93aef6b7dc8050ae405c9049b2d319)) I added support for parsing the two new attributes of [ApplicationCommandInteractionData](https://discord.com/developers/docs/interactions/message-components#responding-to-a-component-interaction) (custom_id and component_type). I created a new class [InteractionComponentData](https://github.com/Javacord/Javacord/blob/bc5f11d41d93aef6b7dc8050ae405c9049b2d319/javacord-core/src/main/java/org/javacord/core/command/InteractionComponentDataImpl.java), but I would like a contributor to review this approach.
> > 
> > https://github.com/Javacord/Javacord/blob/bc5f11d41d93aef6b7dc8050ae405c9049b2d319/javacord-core/src/main/java/org/javacord/core/command/InteractionImpl.java#L52-L58
> 
> That looks fine to me. The downside here is that a user would have to manually cast the "InteractionData", depending on the type, right? That is not generally a problem, but a very "Javacordy" thing to do would be to offer two methods here:
> 
> ``InteractionData#asApplicationCommandInteractionData()`` that returns an ``Optional<ApplicationCommandInteractionData>`` if and only if the underlying Object is actually of this type; it should return an empty Optional otherwise.
> 
> The same thing could be added for components: ``InteractionData#asInteractionComponentData()``
> 
> You could just define these methods in your InteractionData interface, and implement them in your two data classes accordingly. That would save our users from a cast. This pattern is present in a lot of places in Javacord, for example in Emojis, in Channels, Users etc.
> 
> Edit: Another approach would be to offer these methods as getters, e.g ``getInteractionComponentData()``directly on the Interaction itself. I do not have any strong preferences on any of these; you could even do both, but at least one of them :D

Yeah absolutely, I was thinking of that while looking at the Emoji interface. I think the getters approach will be better since the user can check the type of interaction and get based on that.

dylhack:
Looking good so far
```kotlin
override fun onInteractionCreate(event: InteractionCreateEvent?) {
    if (event == null) {
        return
    }
    val interaction = event.interaction
    val button = interaction.componentData.get()
    val id = button.customId

    when (id) {
        "play_previous" -> {
            event.respond()
                .setContent("Playing previous song.")
                .sendInitialResponse(interaction)
                .join()
        }
        "play_next" -> {
            event.respond()
                .setContent("Playing next song.")
                .sendInitialResponse(interaction)
                .join()
        }
        else -> {
            event.respond()
                .setContent("An error occurred.")
                .sendInitialResponse(interaction)
                .join()
        }
    }
}
```

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
dylhack: I was able to get this working for #750 (see [here](https://github.com/Javacord/Javacord/pull/758#issuecomment-850774223))

KILLEliteMaste (changes requested):
It is not possible to respond to a component interaction with another component message.
1. The InteractionMessageBuilder and WebhookMessageBuilder are missing the component methods. 
2. The responding endpoints with value 6/7 which are only available for components are also missing. Refer to https://discord.com/developers/docs/interactions/slash-commands#interaction-response-interactioncallbacktype

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -48,11 +49,22 @@ public static MessageBuilder fromMessage(Message message) {
         return builder;
     }
 
+    /**
+     * Add an ActionRowBuilder to the components.
+     *
+     * @param builder The ActionRowBuilder.
+     * @return The current instance in order to chain call methods.
+     */
+    public MessageBuilder addActionRow(ActionRowBuilder builder) {
```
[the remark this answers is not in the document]
dylhack: > The top-level components field is an array of ActionRow components.

Yes. But I was thinking of sugar coating it by doing it for the user if they attempt to add a different component.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -177,6 +183,13 @@ public MessageImpl(DiscordApiImpl api, TextChannel channel, JsonNode data) {
             }
         }
 
+        if (data.has("components")) {
+            for (JsonNode componentJson : data.get("components")) {
+                ActionRowImpl actionRow = new ActionRowImpl(componentJson);
+                components.add(actionRow);
+            }
+        }
```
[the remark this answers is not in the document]
dylhack: > The top-level components field is an array of ActionRow components.

Yes

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -48,11 +49,22 @@ public static MessageBuilder fromMessage(Message message) {
         return builder;
     }
 
+    /**
+     * Add an ActionRowBuilder to the components.
```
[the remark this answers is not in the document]
dylhack: resolved: 48974659e62fe1a9ba5f504035e4bdc9ee688f99

javacord-api/src/main/java/org/javacord/api/command/Interaction.java
```diff
@@ -28,7 +29,21 @@
...
      */
-    Optional<ApplicationCommandInteractionData> getData();
+    Optional<InteractionData> getData();
+
+    /**
+     * Get application data.
+     *
+     * @return The application data.
+     */
+    Optional<ApplicationCommandInteractionData> getApplicationData();
```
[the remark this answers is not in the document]
dylhack: resolved: 6fa6b756e5b8506495ceede2a743c38763576810

javacord-core/src/main/java/org/javacord/core/entity/message/component/ActionRowImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+                switch (type) {
+                    case ACTION_ROW:
+                        throw new IllegalStateException("An action row is inside of an action row.");
+                    case BUTTON:
+                        Button button = new ButtonImpl(componentJson);
+                        components.add(button);
+                        break;
+                    default:
+                        throw new IllegalStateException("Unknown component parsed, please contact developer!");
```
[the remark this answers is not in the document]
dylhack: Would you want the full object or just the type?
https://github.com/Javacord/Javacord/blob/6fa6b756e5b8506495ceede2a743c38763576810/javacord-core/src/main/java/org/javacord/core/entity/message/component/ActionRowImpl.java#L36-L39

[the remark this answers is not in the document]
KILLEliteMaste: I think the type is sufficient. Something like
`"Couldn't parse the component of type 'ID'. Please contact the developer!"`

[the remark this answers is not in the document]
dylhack: resolved: 9924ea5cc31992435607d6573e95607e69dfba98

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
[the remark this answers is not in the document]
KILLEliteMaste: I agree. A method like `addComponents(Component... components)` which under the hood automatically adds the components to an ActionRow would be a nice addition. But I see a problem with the current approach of extending `Components`.
From Discord:
> The top-level components field is an array of ActionRow components.
> An ActionRow is a non-interactive container component for other types of components. It has a type: 1 and a sub-array of components of other types.
> An ActionRow cannot contain another ActionRow

Currently it is possible to add an ActionRowBuilder to an ActionRowBuilder as it expects a `Component`. I think there needs to be something like a TopLevelComponent for components like ActionRow to fix this issue. Then it shouldn't be a problem to add the above mentioned convenience method.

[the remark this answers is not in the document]
dylhack: Well I mean an addComponent to the MessageBuilder:
1. if an action row doesn't exist then create one and append it
2. if the last action row has met max capacity then append a new one
3. if the message has met max action row capacity then throw an exception
4. finally append the component to the next available action row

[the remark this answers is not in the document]
dylhack: and if the component added is an ActionRow well that's self-explanatory.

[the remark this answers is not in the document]
KILLEliteMaste: The exact behaviour is discussable(and btw. we don't have any checks if there are too many of a thing as Discord can change these values any time). But the issue I mentioned still exist. 
This is currently valid code:
```
ActionRowBuilder actionRow = new ActionRowBuilder().addComponent(new ActionRowBuilder());
```
But as discord states:
> An ActionRow cannot contain another ActionRow

Therefore the addComponent method in the MessageBuilder is also capable of adding another ActionRow to an ActionRow.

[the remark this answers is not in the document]
coaxial1: Sounds like a good suggestion, but this needs to be documented precisely.

About the issue mentioned above (by @KILLEliteMaste ), we could either:
- Add a new top level component above things that can be added to actions rows (sth like "ToActionRowsAddableComponent" (please pick a better name) that extends "Component"), and let ActionRows only accept these
- or keep it simple and just throw an exception if someone is dumb enough to add an ActionRow to an ActionRow

I'm not sure, but I think I'd prefer the second approach to keep things simple ... what do you think?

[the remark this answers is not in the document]
dylhack: Throwing an exception seems better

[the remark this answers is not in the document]
KILLEliteMaste: I can image of Discord also adding at some point things like a GridPane. So imo the current best approach is as I said to add a TopLevelComponent which things like the ActionRow can extend and Button and so on can extend from Component which can be added to the ActionRow. I would say already adding a "ToActionRowsAddableComponent" would be a bit too early as we don't know in which direction Discord will be going to and this jsut adds more complexity for now.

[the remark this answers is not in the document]
dylhack: But one thing is for certain
```java
if (component.getType() == ComponentType.ACTION_ROW) {
   throw new Exception();
}
```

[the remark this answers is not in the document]
KILLEliteMaste: But this is not necessary if you can only pass a `LowLevelComponent` as the ActionRow will be a `TopLevelComponent`. So both extends from `Component` and then you set `LowLevelComponent` as the argument for the methods.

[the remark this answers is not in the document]
dylhack: Yeah HighLevelComponent and LowLevelComponent seems like a good approach

dylhack:
> It is not possible to respond to a component interaction with another component message.
> 
>     1. The InteractionMessageBuilder and WebhookMessageBuilder are missing the component methods.
> 
>     2. The responding endpoints with value 6/7 which are only available for components are also missing. Refer to https://discord.com/developers/docs/interactions/slash-commands#interaction-response-interactioncallbacktype
> 
>     3. There's also a new field for InteractionApplicationCommandCallbackData see https://github.com/discord/discord-api-docs/pull/3012/files

1. 8b4d6a68104b81c0a728cc6a54c9757bf7ad0268 - Oops, I thought it would've inherited naturally from the MessageBuilder, but that's only true to the MessageBuilderDelegate.
2. 8f10d2c3f54535551474960cec8925a0f9f30457
3. 8f10d2c3f54535551474960cec8925a0f9f30457 - [applied here](https://github.com/Javacord/Javacord/pull/758/commits/8f10d2c3f54535551474960cec8925a0f9f30457#diff-03a8b082d6927937bee65123d6928793fc87e77bcc743ad904483bd600026bd8R145)

dylhack:
Should we rename the commands package to interactions?

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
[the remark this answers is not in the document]
dylhack: ready for review: 68d4f0ea637e7c3c7f7d29b0e9ea6056b179f69d

javacord-api/src/main/java/org/javacord/api/entity/message/component/ActionRowBuilder.java
```diff
@@ -5,7 +5,7 @@
 
 import java.util.List;
 
-public class ActionRowBuilder implements ComponentBuilder {
+public class ActionRowBuilder implements ComponentBuilder, HighLevelComponentBuilder {
```
KILLEliteMaste, line 8: ComponentBuilder is not needed as HighLevelComponentBuilder already extends it

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonBuilder.java
```diff
@@ -5,7 +5,7 @@
 import org.javacord.api.entity.message.component.internal.ButtonBuilderDelegate;
 import org.javacord.api.util.internal.DelegateFactory;
 
-public class ButtonBuilder implements ComponentBuilder {
+public class ButtonBuilder implements ComponentBuilder, LowLevelComponentBuilder {
```
KILLEliteMaste, line 8: ComponentBuilder is not needed as LowLevelComponentBuilder already extends it

javacord-api/src/main/java/org/javacord/api/entity/message/InteractionMessageBuilder.java
```diff
@@ -118,11 +118,11 @@ public InteractionMessageBuilder addEmbeds(EmbedBuilder... embeds) {
     /**
      * Adds the component to the message.
```
KILLEliteMaste, line 119: You forgot to update the Javadoc accrodingly to the High/Low level component as it says in all places just "component"

dylhack: Well I explicitly changed button & action row to "component" because I didn't see a problem with just "component"

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonBuilder.java
dylhack: resolved: ed1c787d638a35dd3f0b9f391a94ae345e51d6a8

KILLEliteMaste:
> Should we rename the command package to interaction?

What package do you mean exactly?

dylhack:
> > Should we rename the command package to interaction?
> 
> What package do you mean exactly?

`org.javacord.api.command` -> `org.javacord.api.interaction`

KILLEliteMaste:
> > > Should we rename the command package to interaction?
> > 
> > 
> > What package do you mean exactly?
> 
> `org.javacord.api.command` -> `org.javacord.api.interaction`

I would say so, yes. Same of course for `org.javacord.core`

javacord-api/src/main/java/org/javacord/api/entity/message/InteractionMessageBuilder.java
dylhack: resolved: fc69d32ea74de6f6c18ce1c44ab3429157ef5c6d

KILLEliteMaste (commented):
At this point this PR looks pretty decent to me!

dylhack:
@Bastian could I get a review?

Bastian:
I'll try to look into it asap.
But I probably won't get to it until monday.

coaxial1:
Hey, I have a few things to add, I don't think we're finished yet. :)

IMO, it doesn't make sense that you have the ``updateMessage()`` and ``updateMessageLater()`` methods in the ``InteractionMessageBuilder``. These methods must be in the same place as "respond()" and "respondLater()". Those 4 methods are 4 different ways to react to an incoming interaction (types 4, 5, 6 and 7) and should (must, tbh) be in the same location.

Also, it doesn't make sense that you're calling a shared internal method from both "updateMessage[Later]" methods; the ``updateMessageLater()`` request should not have anything in the ``data`` field, because we're not updating anything yet. We can even use this response type to simply acknowledge that someone clicked a button, but not do anything. It's similar to ``respondLater`` which also doesn't have anything in the ``data`` field, and even has a hardcoded request body. We should do the same for ``updateMessageLater``.

After we moved the new methods into ``InteractionCreateEvent``(that's where ``respond``and ``respondLater``are in) we also should make sure that we get a prefilled message builder if we call ``update`` or ``updateLater``, since we're updating an already existing message, and do not only create a new message like in ``respond`` and  ``respondLater``.

The call chain for ``update()`` should be really similar to the one of ``respond()``.
For ``respond()``, we can do:
```java
api.addInteractionCreateListener(event -> {
    event.respond().setContent("Initial Response").sendInitialResponse(event.getInteraction());
});
```
_(btw I think that the need to add ``event.getInteraction()`` as parameter in the end is also bad, but I'll address that with a PR once we merged this one)_

The same should be possible for ``update``, with the exception that it would look like

```java
api.addInteractionCreateListener(event -> {
    if(event.getInteraction().getType() == MESSAGE_COMPONENT)
        event.updateMessage().setContent("Updated Message Content").update();
});
```

...and that the MessageBuilder (whatever kind of message builder, maybe we need a new one?) is already prefilled with the values of the actual message we're updating so we can do changes to some message fields (e.g. content) without dropping everything else (e.g. embeds) that was previously present in the message.

Lastly, I think we should make sure that we throw an exception if someone uses one of the new ``update / updateLater`` methods if the interaction is not of type ``MESSAGE_COMPONENT``.

Sorry for that long wall of text! :D

dylhack:
Thanks, I'll look into this later today when I get back on

dylhack:
> (btw I think that the need to add event.getInteraction() as parameter in the end is also bad, but I'll address that with a PR once we merged this one)

I was thinking about this the whole time, I didn't know if there was a specific reason for doing it this way

coaxial1:
> I was thinking about this the whole time, I didn't know if there was a specific reason for doing it this way

I think the reason was that you can use an InteractionMessageBuilder to send a follow up message, and you need to tell the message builder to which this follow up message belongs.

There's surely a better way to solve it - but it's not a matter of this PR I guess ;D I feel like we should do a "Polish interaction API" PR after this one to get everything aligned.

dylhack:
> IMO, it doesn't make sense that you have the updateMessage() and updateMessageLater() methods in the InteractionMessageBuilder. These methods must be in the same place as "respond()" and "respondLater()". Those 4 methods are 4 different ways to react to an incoming interaction (types 4, 5, 6 and 7) and should (must, tbh) be in the same location.

> Also, it doesn't make sense that you're calling a shared internal method from both "updateMessage[Later]" methods; the updateMessageLater() request should not have anything in the data field, because we're not updating anything yet. We can even use this response type to simply acknowledge that someone clicked a button, but not do anything. It's similar to respondLater which also doesn't have anything in the data field, and even has a hardcoded request body. We should do the same for updateMessageLater.

> (btw I think that the need to add event.getInteraction() as parameter in the end is also bad, but I'll possibly address that with a PR once we merged this one)

Ready for review: 260ea497ed59f97d560b6254afd3333a47363e0c


> After we moved the new methods into InteractionCreateEvent(that's where respondand respondLaterare in) we also should make sure that we get a prefilled message builder if we call update or updateLater, since we're updating an already existing message, and do not only create a new message like in respond and respondLater.

Instead of copying the data without the user knowing I created a copy method which is split up into multiple commits (see below). If the user wants to they can copy the message attached to the Interaction instance.
 * ActionRowBuilder#copy: 96dc9591f8edc3f7c1a60170bbe991029e55da61, bd9861cb9cd58e1ec42002656a0b684c83b65458
 * ButtonBuilder#copy: 8aa95de6f068e1a7c34eb163bab67d65d3242662, 6e10eda26a407e6321551ea56e0ef36573520066
 * MessageBuilder#copy: d9ba555d7551242a6f69b613ee83f423dba27656, 90ef2525f12720b87ae2f672510ef3fb56b31958

KILLEliteMaste:
Changing the reponding flow really shouldn't be part of this PR but my intentions doing it this way was to follow the existing MessageBuilders. There has already been a discussion in the Discord channel #design-discussion to change how the MessageBuilder works but there werent a lot of responses at the end of it so I have kept it the same as we already do it. And imo changing it just for the interaction responses would be confuising to the already confusion multiple MessageBuilders.

coaxial1:
About the copy methods: Since you have a new "interaction" field in the InteractionMessageBuilder, is there a method like "copyFromInteraction()" which just takes the message from the interaction that is saved in the interaction field and calls "copy(Message m)" with it? Otherwise, we'd have to pass in ``event.getInteraction().getMessage()`` again.

dylhack:
> About the copy methods: Since you have a new "interaction" field in the InteractionMessageBuilder, is there a method like "copyFromInteraction()" or "copyOriginal()" or "copyComponentMessage()" which just takes the message from the interaction that is saved in the interaction field and calls "copy(Message m)" with it? Otherwise, we'd have to pass in ``event.getInteraction().getMessage()`` again.

sounds good, I'll add that later today

dylhack:
> Changing the reponding flow really shouldn't be part of this PR but my intentions doing it this way was to follow the existing MessageBuilders. There has already been a discussion in the Discord channel #design-discussion to change how the MessageBuilder works but there werent a lot of responses at the end of it so I have kept it the same as we already do it. And imo changing it just for the interaction responses would be confuising to the already confusion multiple MessageBuilders.

We can meet in the middle by adding functions that intake interaction objects and keeping the ones that don't and making the interaction property nullable

dylhack:
> About the copy methods: Since you have a new "interaction" field in the InteractionMessageBuilder, is there a method like "copyFromInteraction()" or "copyOriginal()" or "copyComponentMessage()" which just takes the message from the interaction that is saved in the interaction field and calls "copy(Message m)" with it? Otherwise, we'd have to pass in `event.getInteraction().getMessage()` again.

Done: b82d6538bc641037f765d829a6b825bc0ddb87a8

dylhack:
> Changing the reponding flow really shouldn't be part of this PR but my intentions doing it this way was to follow the existing MessageBuilders. There has already been a discussion in the Discord channel #design-discussion to change how the MessageBuilder works but there werent a lot of responses at the end of it so I have kept it the same as we already do it. And imo changing it just for the interaction responses would be confuising to the already confusion multiple MessageBuilders.

What do you think? 6c103deeb930461be80d54f0d27f6cb0eeae7035

dylhack:
Is there a type 3 component? I sometimes get a PacketHandler error with this body:
```json
{
  "version": 1,
  "type": 3,
  "message": {
    "id": "849056836371480576",
    "flags": 64
  },
  "member": {},
  "id": "849056843589615657",
  "guild_id": "",
  "data": {
    "custom_id": "click_me",
    "component_type": 2
  },
  "channel_id": "803685985346256956",
  "application_id": "753666554637451425"
}
```

coaxial1:
Please do not use the assert keyword in that manner :) Just throw any appropriate exception to indicate what went wrong.
Otherwise, I think it looks fine. I'll try to do some testing with the PR soon.

KILLEliteMaste:
> > Changing the reponding flow really shouldn't be part of this PR but my intentions doing it this way was to follow the existing MessageBuilders. There has already been a discussion in the Discord channel #design-discussion to change how the MessageBuilder works but there werent a lot of responses at the end of it so I have kept it the same as we already do it. And imo changing it just for the interaction responses would be confuising to the already confusion multiple MessageBuilders.
> 
> What do you think? [6c103de](https://github.com/Javacord/Javacord/commit/6c103deeb930461be80d54f0d27f6cb0eeae7035)

Imo this makes it even worse as it allows it to be not funtional at all if it's being used wrong. And again changing this api design shouldn't be part of this PR as it's quiet a bigger topic and may affect more than just responding to interactions

dylhack:
> Please do not use the assert keyword in that manner :) Just throw any appropriate exception to indicate what went wrong.
> Otherwise, I think it looks fine. I'll try to do some testing with the PR soon.

removed as of: 7cf8ecff10022cb0b76efc51cd313221f295c843

dylhack:
> > > Changing the reponding flow really shouldn't be part of this PR but my intentions doing it this way was to follow the existing MessageBuilders. There has already been a discussion in the Discord channel #design-discussion to change how the MessageBuilder works but there werent a lot of responses at the end of it so I have kept it the same as we already do it. And imo changing it just for the interaction responses would be confuising to the already confusion multiple MessageBuilders.
> > 
> > 
> > What do you think? [6c103de](https://github.com/Javacord/Javacord/commit/6c103deeb930461be80d54f0d27f6cb0eeae7035)
> 
> Imo this makes it even worse as it allows it to be not funtional at all if it's being used wrong. And again changing this api design shouldn't be part of this PR as it's quiet a bigger topic and may affect more than just responding to interactions

okay reverted: 76f5ba2abbafa0de450133675fb342bce41ed995

javacord-api/src/main/java/org/javacord/api/entity/message/component/ButtonStyle.java
```diff
@@ -0,0 +1,76 @@
...
+    ButtonStyle(int i) {
+        this.data = i;
+    }
+
+    /**
+     * The button style internal value.
+     *
+     * @return The button style's internal value
+     */
+    public int value() {
```
Bastian, line 22:
```suggestion
    public int getValue() {
```

javacord-api/src/main/java/org/javacord/api/entity/message/component/ActionRowBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+public class ActionRowBuilder implements HighLevelComponentBuilder {
+    private final ActionRowBuilderDelegate delegate = DelegateFactory.createActionRowBuilderDelegate();
+
+    /**
+     * Add a low-level component builder.
+     *
+     * @param component The low-level component builder.
+     * @return The builder instance to chain methods.
+     */
+    public ActionRowBuilder addComponent(LowLevelComponentBuilder component) {
```
Bastian, line 17: The `MessageBuilder` has a `addComponents(...)` method to add multiple components at once.
To be consistent, we should also add a similar method here.

Bastian (approved):
I didn't dive super deep into the code, but I've played around with the API a bit and took a quick look at the source code.

Overall it looks good to me and I think we can merge this PR 👍 

(_Receiving events for buttons (and interactions) in general is still a bit of a PITA (since it's super low-level at the moment) but that's something that can be improved in the future and is out of scope for this PR._)

Bastian:
Ohh, and however merges this: Please squash it into a single commit and not merge all 69 commits into Javacord.

javacord-api/src/main/java/org/javacord/api/event/interaction/InteractionCreateEvent.java
```diff
@@ -33,6 +33,20 @@
...
+     * @return An interaction message builder.
+     */
+    InteractionMessageBuilder updateMessage();
+
+    /**
+     * Edit the message the component was attached to later.
+     *
+     * @return A completable future which lets Discord know the message will be updated later.
+     */
+    CompletableFuture<Void> updateMessageLater();
```
KILLEliteMaste, line 48: As these are only valid for component messages what do you think about naming them something like `updateComponentMessage`?

javacord-api/src/main/java/org/javacord/api/entity/message/InteractionMessageBuilder.java
```diff
@@ -541,12 +618,21 @@ public StringBuilder getStringBuilder() {
...
         return delegate.editFollowupMessage(interaction, messageId);
     }
 
+    /**
+     * Update the message the components were attached to.
+     *
+     * @param interaction The original interaction.
+     * @return The completable future to determine if the message was updated.
+     */
+    public CompletableFuture<Void> update(Interaction interaction) {
```
KILLEliteMaste, line 635: Shouldn't this be called `updateMessage`? Or according to the name if you decide to change it

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
```diff
@@ -19,34 +19,41 @@
...
      */
     CompletableFuture<Message> sendFollowupMessage(Interaction interaction);
 
+    /**
+     * Edit the message the component was attached to.
+     *
+     * @param interaction The interaction.
+     * @return The completable future to determine if the message was updated.
+     */
+    CompletableFuture<Void> update(Interaction interaction);
```
KILLEliteMaste, line 49: Same here. Shouldn't this be called `updateMessage`? Or according to the name if you decide to change it

javacord-api/src/main/java/org/javacord/api/entity/message/InteractionMessageBuilder.java
dylhack: I think it should be update since it's obvious they're updating a message after calling updateMessage. Like if these builders actually ended with "build"

javacord-api/src/main/java/org/javacord/api/event/interaction/InteractionCreateEvent.java
dylhack: Yeah that would be better

dylhack: resolved: 7b4b0b7f53def7c10a6ecf721622280ef471d9d1

javacord-api/src/main/java/org/javacord/api/entity/message/InteractionMessageBuilder.java
KILLEliteMaste: Don't forget you can still use only the InteractionMessageBuilder which will then look like this:
`new InteractionMessageBuilder().update()`. And we actually don't even need the `updateComponentMessage` in the event as there is already `respond` which is the same except for the name. Furthermore if there's at any time a new interaction which also updates a message we would have to use another name anyways and then it may not be that nice to have `update and updateXY` as only one says what it actually updates

coaxial1: Well, but imo it's more confusing to call ``respond()`` if I don't want to respond at all, so using it for both is not a very good solution.

dylhack: resolved: 2530bb6a99cfbab8ca532859a588e86535805081

KILLEliteMaste: But you only call it if you want to respond to an interaction instantly. If not you use on of these two `respondLater` `updateComponentMessageLater`

coaxial1: And what do I call if I want to update the comopment message right now? ``respond()`` would be semantically incorrect.

KILLEliteMaste: Well, you respond to the invoked interaction which may be a slash command or a pressed button. I also open for to rename it to `createInteractionMessageBuilder` which describes what it actually returns

dylhack:
Could you all review: 3e793c3d2e15dd626c8f2ef3b2e3d4db1f01dd4b. This will allow us to do stuff like this
```kotlin
InteractionMessageBuilder()
    .setContent("Sent.")
    .sendInitialResponse(interaction)
    .join()
    .deleteInitialResponse(interaction)
```

dylhack:
Could you all review: 3e793c3d2e15dd626c8f2ef3b2e3d4db1f01dd4b. This will allow us to do stuff like this
```kotlin
InteractionMessageBuilder()
    .setContent("Sent.")
    .sendInitialResponse(interaction)
    .join()
    .deleteInitialResponse(interaction)
```

KILLEliteMaste:
@coaxial1 is something left that needs to be addressed or do you think we can merge this PR?

coaxial1:
I think it's fine. We can get onto optimizing the call flow for responding in a new PR.

coaxial1 (approved):

KILLEliteMaste:
@dylhack Can you please rebase to the latest version and squash the PR into a single commit?

KILLEliteMaste (approved):

PR #762, opened 2021-06-02
Adjusted user flags to the current existing ones

KILLEliteMaste:
#### Breaking changes: 
- Removed `SYSTEM`
- Renamed various other user flags to match the documentation
- Added `CERTIFIED_MODERATOR`, `BOT_HTTP_INTERACTIONS`

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
@@ -2,19 +2,21 @@ package org.javacord.api.entity.user;
 
 public enum UserFlag {
 
-    DISCORD_EMPLOYEE(1 << 0, "Discord Employee"),
-    DISCORD_PARTNER(1 << 1, "Discord Partner"),
-    HYPESQUAD_EVENTS(1 << 2, "HypeSquad Events"),
+    NONE(0, "None"),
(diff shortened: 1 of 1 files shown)


javacord-api/src/main/java/org/javacord/api/entity/user/UserFlag.java
```diff
@@ -11,10 +11,10 @@
     HOUSE_BALANCE(1 << 8, "House Bravery"),
     EARLY_SUPPORTER(1 << 9, "Early Supporter"),
     TEAM_USER(1 << 10, "Team User"),
-    SYSTEM(1 << 12, "System"),
     BUG_HUNTER_LEVEL_2(1 << 14, "Bug Hunter Level 2"),
     VERIFIED_BOT(1 << 16, "Verified Bot"),
-    VERIFIED_BOT_DEVELOPER(1 << 17, "Verified Bot Developer");
+    VERIFIED_BOT_DEVELOPER(1 << 17, "Verified Bot Developer"),
```
coaxial1, line 16: We should update this to "Early Verified Bot Developer".

coaxial1 (changes requested):
We can update the dev flag in this PR as well. :)

coaxial1 (approved):
LGTM

PR #764, opened 2021-06-06
Add GitHub Action for codestyle checks

Bastian:
The action checks for code style violations and displays a warning in the UI:
image
image

You can view it in action here: https://github.com/Bastian/Javacord/pull/5

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -1,0 +1,29 @@
+name: Codestyle
+on: pull_request
+
+jobs:
+  checkstyle_job:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Bastian:
I would appreciate a second opinion on the proposed solutions.
At the moment, I have no idea how to tell the gradle checkstyle task to create a single report, so the only option I see is to merge the reports manually (which is something I would like to avoid if possible).

Bastian:
Okay, nevermind.
I've found a solution: Just give them two different names.

Bastian:
Unfortunately, the annotations are only shown when looking at a single commit but not when looking at all changes combined (the "Files changed" button).

PR #768, opened 2021-06-09
Rework of interaction responses

coaxial1:
First attempt on reworking how responding to interactions works. 

I'd appreciate a suggestion on how to avoid the massive code duplication in the new response builders; due to the fact that they have to return their own type for call chaining, I haven't found a way to do it, but it's late.
Even Stackoverflow suggests that this doesn't work differently: https://stackoverflow.com/a/21188247

Here's a working snippet if you want to play around and see the API. It covers not all possible ways to react to an interaction

```java
import org.javacord.api.DiscordApiBuilder;
import org.javacord.api.entity.message.MessageBuilder;
import org.javacord.api.entity.message.component.ActionRowBuilder;
import org.javacord.api.entity.message.component.ButtonBuilder;
import org.javacord.api.entity.message.component.ButtonStyle;
import org.javacord.api.entity.message.embed.EmbedBuilder;
import org.javacord.api.interaction.ApplicationCommand;
import org.javacord.api.interaction.ApplicationCommandBuilder;
import org.javacord.api.interaction.ApplicationCommandOptionBuilder;
import org.javacord.api.interaction.ApplicationCommandOptionType;
import org.javacord.api.util.logging.ExceptionLogger;

public class Test {
        var api = new DiscordApiBuilder()
                .setToken("hidden")
                .login().join();

        // Create a say command if not present
        ApplicationCommand sayCommand = api.getGlobalApplicationCommands()
                .join()
                .stream()
                .filter(cmd -> cmd.getName().equals("say"))
                .findAny()
                .orElseGet(() -> new ApplicationCommandBuilder()
                        .setName("say")
                        .setDescription("The bot will say what you tell it.")
                        .addOption(new ApplicationCommandOptionBuilder().setName("text").setDescription("The text I should say.")
                                .setRequired(true).setType(ApplicationCommandOptionType.STRING).build())
                        .createGlobal(api)
                        .join());

        api.addInteractionCreateListener(e -> {

            // On slash command, reply with the requested phrase, and add buttons to
            // 1. Delete the message
            // 2. Say it again
            e.getInteraction().asApplicationCommandInteractionWithCommandId(sayCommand.getId())
                    .ifPresent(i -> i.createImmediateResponder()
                            .setContent(i.getOptions().get(0).getStringValue().orElse("ERROR!"))
                            .addComponents(getActionRow())
                            .respond()
                            .exceptionally(ExceptionLogger.get()));

            // If delete button is pressed, first ack, then delete.
            e.getInteraction().asMessageComponentInteractionWithCustomId("del")
                    .ifPresent(i -> {
                        i.acknowledge().thenAccept(v -> i.getMessage().delete());
                    });

            // If we should say it again, send as follow up, just as demo instead of immediate response
            e.getInteraction().asMessageComponentInteractionWithCustomId("again")
                    .ifPresent(i -> {
                        i.acknowledge().thenAccept(v ->
                                i.createFollowupMessageBuilder()
                                        .setContent(i.getMessage().getContent())
                                        .addComponents(getActionRow())
                                        .send()
                                        .exceptionally(ExceptionLogger.get()));
                    });
        });
    }

    private static ActionRowBuilder getActionRow() {
        return new ActionRowBuilder()
                .addComponents(new ButtonBuilder()
                                .setCustomId("del")
                                .setLabel("Delete")
                                .setStyle(ButtonStyle.DANGER),
                        new ButtonBuilder()
                                .setCustomId("again")
                                .setLabel("Say it again!")
                                .setStyle(ButtonStyle.PRIMARY)
                );
    }
   ```

### CODE REVIEW

**Comment by KILLEliteMaste** [comment]
I have played around a bit with the syntax and it's going to a good way. The first thing I came across was the implementation of the `Interaction` in `MessageComponentInteraction`. This causes a duplication of methods and result in these thing being the same
```java
e.getInteraction().getVersion()
i.getVersion()
```
which I don't really like as these information are already available in the interaction object.
Furthermore this allows the user to infinite call the asXY methods `e.getInteraction().asMessageComponentInteraction().get().asMessageComponentInteraction().get()`.

And for the components abstraction I think we should have another layer (asButton / asDropdown) as someone said there will be a dropdown in the future which will send additional data. Doing it that way is also future proof as once theres even more data we don't have useless methods in the `MessageComponentInteraction` object.

javacord-core/src/main/java/org/javacord/core/interaction/InteractionImpl.java
```diff
@@ -158,4 +136,16 @@ public String getToken() {
...
     }
+
+    /**
+     * Checks if a response method to this interaction has already been called.
+     * Throws an IllegalStateException if true; sets the related flag to <code>true</code> and returns otherwise.
+     */
+    public void checkResponseMethodCallable() {
+        if (responseSent) {
+            throw new IllegalStateException("You already replied to this interaction; you can not reply twice.");
+        }
```
KILLEliteMaste, line 147: I am not sure if that's how we should handle this as it quiet breaks the code flow as you have to catch it while the CF otherwise would just execute exceptionally. If we want to avoid a request to Discord, we could also throw a CF that executes exceptionally. But it would be a good thing if we could expose the response sent to the api, just in case someone wants to check it by himself.

javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
```diff
@@ -29,8 +37,33 @@ public void handle(JsonNode packet) {
...
+                break;
+            case MESSAGE_COMPONENT:
+                int componentType = packet.get("data").get("component_type").asInt();
+                switch (ComponentType.fromId(componentType)) {
+                    case BUTTON:
+                        interaction = new ButtonInteractionImpl(api, channel, packet);
+                        break;
+                    case ACTION_ROW:
+                        logger.error("Received a message component interaction of type ACTION_ROW. This should not"
+                                + " be possible.");
```
KILLEliteMaste, line 54: Should be a warning

javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
```diff
@@ -29,8 +37,33 @@ public void handle(JsonNode packet) {
...
+                        return;
+                    default:
+                        logger.error(String.format("Received message component interaction of unknown type %s. "
+                                + "Please contact the developer!", componentType));
```
KILLEliteMaste, line 58: Should be a warning

javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
```diff
@@ -29,8 +37,33 @@ public void handle(JsonNode packet) {
...
+                        return;
+                }
+                break;
+            default:
+                logger.error(String.format("Received interaction of unknown type %s. "
+                        + "Please contact the developer!", type));
```
KILLEliteMaste, line 64: Should be a warning

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
```diff
@@ -22,56 +22,64 @@
...
     /**
      * Edits the message.
      *
      * @param interaction The interaction.
-     * @param messageId The message id of the followup message which should be edited.
+     * @param messageId   The message id of the followup message which should be edited.
      * @return The sent message.
      */
-    CompletableFuture<Message> editFollowupMessage(Interaction interaction, String messageId);
+    CompletableFuture<Message> editFollowupMessage(InteractionBase interaction, String messageId);
```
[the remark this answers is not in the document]
coaxial1: Actually, we do in the public InteractionMessageBuilder, this is just the internal delegate. Do you think it's important to offer it here as well?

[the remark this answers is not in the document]
coaxial1: We do this in the public InteractionMessageBuilder and InteractionFollowupMessageBuilder. Do you think it's important to offer that in the internal delegate class as well even though it's not to be accessed by any library user?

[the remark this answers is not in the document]
KILLEliteMaste: Ah my bad. It's fine like it is  :)

javacord-api/src/main/java/org/javacord/api/event/interaction/InteractionCreateEvent.java
```diff
@@ -19,39 +21,59 @@
...
+     * For regular users, we recommend to use the different {@code getXXX()} methods to walk down the interaction
+     * inheritance tree.
+     *
+     * <p>If the interaction is not castable to the specified type, the {@code Optional} will be empty.
      *
-     * @return The completable future when the message has been deleted.
+     * @param type The type as which to obtain this interaction.
+     * @param <T>  The desired type.
+     * @return Returns an {@code Optional} of this interaction if it could be cast, otherwise an empty result.
+     * @throws NullPointerException If the type is {@code null}.
```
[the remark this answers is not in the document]
coaxial1: To keep it consistent, I'll also add a modified form of ``Specializable``, otherwise this method and ``Interaction#as(Class t)`` would be have differently.

javacord-core/src/main/java/org/javacord/core/interaction/ComponentInteractionOriginalMessageUpdaterImpl.java
```diff
@@ -0,0 +1,28 @@
...
+     * @param interaction The interaction to use.
+     */
+    public ComponentInteractionOriginalMessageUpdaterImpl(MessageComponentInteraction interaction) {
+        super(ComponentInteractionOriginalMessageUpdater.class);
+        this.interaction = (MessageComponentInteractionImpl) interaction;
+    }
+
+    @Override
+    public CompletableFuture<Void> update() {
+        interaction.checkResponseMethodCallable();
```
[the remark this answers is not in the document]
coaxial1: No; this is response type 7 and must be executed within 3 seconds after receiving the interaction. You can only send one direct response to an interaction, thus, the check is included here.

To be honest, I don't quite understand why they call type 6 "DeferredUpdateMessage" as I don't see how we could utilize any of the interaction or webhook endpoints to update the message later.
Also, since we're probably talking about bot messages, we can always update them in the future using the regular message update endpoint, so there wouldn't be any point in doing so either.
That's why I called type 6 "acknowledge()".

javacord-core/src/main/java/org/javacord/core/interaction/InteractionImpl.java
coaxial1: I would see it this way: Any code that ever happens to call an interaction response twice or more times is wrong in itself, and thus it's a good thing to break the code flow. The user is 100% doing something wrong here.

On the other side, we don't generally check for such things in Javacord (for example we don't do permission checking before sending a reqeust), so I don't know if we should do it here and let Discord send back the exception... Hm.

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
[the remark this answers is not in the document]
coaxial1: We actually do in user facing classes. I don't think it's necessary to offer it in the internal delegate?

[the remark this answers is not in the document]
KILLEliteMaste: > We actually do in user facing classes

Yeah, I must have missed it and it was my only concern that we do not offer it in the user facing classes.

javacord-core/src/main/java/org/javacord/core/interaction/ComponentInteractionOriginalMessageUpdaterImpl.java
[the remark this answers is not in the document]
KILLEliteMaste: That makes sense. Who knows why that added response type 6 🤷

javacord-core/src/main/java/org/javacord/core/interaction/InteractionImpl.java
KILLEliteMaste: But on the other side your are also right:
> Any code that ever happens to call an interaction response twice or more times is wrong in itself

As such code would **never** work, it's should be told to the user that there's something *horribly* wrong in the current code. And while the missing permission check in the other parts of the code can possibly work at some point so it makes quiet some sense to just try executing the request. 
Idk, the thing that this is just different from any other Javacord behaviour is a bit weird.

coaxial1: Yeah we can just drop the check entirely imo. Discord will reject it anyways so the user will be told as soon as they test it. Also, the current implementation prevents retryting the request in case of network errors, for example, since we set the flag even if the request doesn't work.
What do you think?

KILLEliteMaste: I'd vote too for removing it. This is the easiest as we don't need to have a more complex implementation to get it working right and we don't change the behaviour in comparison to the other currently existing methods.

KILLEliteMaste:
As stated in the Interactions issue, we should add some convenience methods for extracting simple options without having to use the streaming api.

coaxial1:
> As stated in the Interactions issue, we should add some convenience methods for extracting simple options without having to use the streaming api.

Let's do another PR? Seems like a separate issue (extracting option from incoming interactions vs. sending out a response)

javacord-api/src/main/java/org/javacord/api/event/interaction/InteractionCreateEvent.java
```diff
@@ -19,39 +21,61 @@
...
+     * @param type The type as which to obtain this interaction.
+     * @param <T>  The desired type.
+     * @return Returns an {@code Optional} of this interaction if it could be cast, otherwise an empty result.
      */
-    CompletableFuture<Void> deleteInitialResponse();
-
+    default <T extends InteractionBase> Optional<T> getInteractionAs(Class<T> type) {
+        if (type == null) {
+            return Optional.empty();
+        }
```
KILLEliteMaste, line 78: Not needed as you are using SafeSpecializable

KILLEliteMaste (commented):
LGTM!

PR #774, opened 2021-06-10
Add methods to update application commands

KILLEliteMaste:
# Example usage
Given is an application command `/say [text]` with a String argument called text.
To update the argument name from an application command on a sever:
```java
ApplicationCommandOption option = new ApplicationCommandOptionBuilder().setName("new_text")
                .setRequired(true).setType(ApplicationCommandOptionType.STRING).build();

new ApplicationCommandUpdater(APPLICATION_COMMAND_ID)
                .setApplicationCommandOptions(Collections.singletonList(option)).updateForServer(server);
```
Other ways to create an ApplicationCommandUpdater:
```java
ApplicationCommand.createApplicationCommandUpdater(APPLICATION_COMMAND).setName("xy").update(DISCORD_API);
ApplicationCommand.createApplicationCommandUpdater(APPLICATION_COMMAND_ID).setName("xy").update(DISCORD_API);
```

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
@@ -48,4 +48,14 @@ public interface ApplicationCommand extends DiscordEntity {
      * @return A future to check if the deletion was successful.
      */
     CompletableFuture<Void> delete();
+
+    /**
+     * Creates an application command updater from this ApplicationCommand instance.
+     *
+     * @return The application command updater for this ApplicationCommand instance.
+     */
+    default ApplicationCommandUpdater createApplicationCommandUpdater() {
+        return new ApplicationCommandUpdater(this.getId());
+    }
+
 }
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
@@ -1,0 +1,94 @@
+package org.javacord.api.interaction;
+
+import org.javacord.api.DiscordApi;
+import org.javacord.api.entity.server.Server;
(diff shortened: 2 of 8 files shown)


coaxial1:
I don't see any issues with your current code, but I would like to suggest to have a method ``ApplicationCommand#createUpdater()`` that creates a pre-filled updater to easily change e.g. only the name of an existing command, but keeping the rest. Generally, there should be a way to copy an existing command into the updater class.

Also, ``createApplicationCommandUpdater(long commandId)`` could have another variant accepting an ``ApplicationCommand``object instead of the ID, and return a pre-filled updater as well.

coaxial1:
Forget my last comment; most of it was rubbish.

Still, I'd suggest to offer a way to create an updater right from a method in ``ApplicationCommand``.

coaxial1:
Did you remove ``ApplicationCommand#createUpdater()`` again? :D

KILLEliteMaste:
> Did you remove `ApplicationCommand#createUpdater()` again? :D

Seems like I didn't even added it ^^

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
```diff
@@ -48,4 +48,24 @@
...
      */
     CompletableFuture<Void> delete();
+
+    /**
+     * Creates an application command updater.
+     *
+     * @param commandId The id of the application command which should be updated.
+     * @return The application command updater.
+     */
+    static ApplicationCommandUpdater createApplicationCommandUpdater(long commandId) {
```
coaxial1, line 58: I think you midunderstood me; I suggested to have non-static method that immediately creates an updater for ``this`` ApplicationCommand, i.e.
```java
return new ApplicationCommandUpdater(this.getId());
```

KILLEliteMaste: Got you. There you go ^^

coaxial1 (approved):
LGTM!

PR #777, opened 2021-06-12
Add interaction convenience methods

coaxial1:
Added lots of convenience methods. Doing this took forever. Really.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandInteractionOptionsProvider.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandInteractionOptionsProvider.java
@@ -1,0 +1,525 @@
+package org.javacord.api.interaction;
+
+import org.javacord.api.entity.Mentionable;
+import org.javacord.api.entity.channel.ServerChannel;
+import org.javacord.api.entity.permission.Role;
+import org.javacord.api.entity.user.User;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Internal interface that offers a few option related convenience methods.
+ * Outsourced to prevent code duplication.
+ */
+public interface ApplicationCommandInteractionOptionsProvider {
+    /**
+     * Gets a list with the params and values from the user, or, if this command contains subcommands and/or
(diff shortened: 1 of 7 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandInteractionOptionsProvider.java
```diff
@@ -0,0 +1,335 @@
...
+     *     use {@code requestFirstOptionMentionableValue()}.
+     *
+     * @return An Optional with the mentionable value of such an option if it exists; an empty Optional otherwise
+     */
+    default Optional<Mentionable> getFirstOptionMentionableValue() {
+        return getFirstOption().flatMap(ApplicationCommandInteractionOption::getMentionableValue);
+    }
+
+    /**
+     * Gets the mentionable value of the first option (if present).
```
KILLEliteMaste, line 112: Gets the mentionable value of the first option (if present) `which may be a user, channel or role`

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandInteractionOptionsProvider.java
```diff
@@ -0,0 +1,335 @@
...
+     * @return A list with the params and values from the user.
+     */
+    List<ApplicationCommandInteractionOption> getOptions();
+
+    /**
+     * Get the first option, if present. Useful if you're working with a command that has only one option.
+     *
+     * @return The option at index 0, if present; an empty Optional otherwise
+     */
+    default Optional<ApplicationCommandInteractionOption> getFirstOption() {
```
KILLEliteMaste, line 30: It would be nice to have these methods for the 2nd and 3rd option too as I think they are quiet common.

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandInteractionImpl.java
```diff
@@ -19,7 +19,7 @@
 
     /**
      * Class constructor.
-     *
+     *x
```
KILLEliteMaste, line 22: ?

KILLEliteMaste (approved):
LGTM

PR #778, opened 2021-06-12
Incoming webhooks token fix

Joshix-1:
Not every incoming webhook comes with a token. IncomingWebhooks without a token are useless. So I think we should just ignore them. https://github.com/discord/discord-api-docs/issues/3056

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/TextChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/TextChannel.java
@@ -841,7 +841,15 @@ public interface TextChannel extends Channel, Messageable, TextChannelAttachable
     CompletableFuture<List<Webhook>> getWebhooks();
 
     /**
-     * Gets a list of all incoming webhooks in this channel.
+     * Gets a list of all incoming webhooks in this channel, they are not guaranteed to have an accessible token.
+     *
+     * @return A list of all incoming webhooks in this channel.
+     */
+    CompletableFuture<List<Webhook>> getAllIncomingWebhooks();
+
+    /**
+     * Gets a list of incoming webhooks in this channel.
+     * This method only returns webhooks with a token, that the bot can access.
      *
      * @return A list of all incoming webhooks in this channel.
      */
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -2013,7 +2013,15 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
     CompletableFuture<List<Webhook>> getWebhooks();
 
     /**
-     * Gets a list of all incoming webhooks in this server.
+     * Gets a list of all incoming webhooks in this server, they are not guaranteed to have an accessible token.
+     *
+     * @return A list of all incoming webhooks in this server.
+     */
+    CompletableFuture<List<Webhook>> getAllIncomingWebhooks();
+
+    /**
+     * Gets a list of incoming webhooks in this server.
+     * This method only returns webhooks with a token, that the bot can access.
      *
      * @return A list of all incoming webhooks in this server.
      */
--- a/javacord-core/src/main/java/org/javacord/core/entity/channel/InternalTextChannel.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/channel/InternalTextChannel.java
@@ -187,17 +187,17 @@ public interface InternalTextChannel extends TextChannel, InternalTextChannelAtt
     }
 
     @Override
+    default CompletableFuture<List<Webhook>> getAllIncomingWebhooks() {
+        return new RestRequest<List<Webhook>>(getApi(), RestMethod.GET, RestEndpoint.CHANNEL_WEBHOOK)
+                .setUrlParameters(getIdAsString())
+                .execute(result -> WebhookImpl.createAllIncomingWebhooksFromJsonArray(getApi(), result.getJsonBody()));
+    }
+
+    @Override
     default CompletableFuture<List<IncomingWebhook>> getIncomingWebhooks() {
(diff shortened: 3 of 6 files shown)


coaxial1:
Hm. Why are they useless? What if I just want to get a list of incoming webhooks in a channel, without posting to them? :)

Joshix-1:
The only real point of incoming webhooks is that they have a token and that you are allowed to post to them.
Having the token Optional would defeat the entire purpose of the IncomingWebhook interface. 
You can still get all Webhooks and then filter out all the webhooks with type Incoming, I know that's stupid, but this pr is the quickest non-breaking change I could make.

We could maybe rename the interface to WebhookWithToken or something.

Joshix-1:
I changed it a bit, to allow getting all real incoming webhooks.
This is the best solution without breaking changes, I can think of right now. 

Things to consider:
- rename functions as the naming is confusing
- document things different
- make things more clear 
- rename IncomingWebhook to somthing with Token or Messageable

coaxial1 (approved):

javacord-api/src/main/java/org/javacord/api/entity/channel/TextChannel.java
```diff
@@ -841,7 +841,15 @@ default NonThrowingAutoCloseable typeContinuouslyAfter(
     CompletableFuture<List<Webhook>> getWebhooks();
 
     /**
-     * Gets a list of all incoming webhooks in this channel.
+     * Gets a list of all incoming webhooks in this channel, they are not guaranteed to have an accessible token.
+     *
+     * @return A list of all incoming webhooks in this channel.
+     */
+    CompletableFuture<List<Webhook>> getRealIncomingWebhooks();
```
KILLEliteMaste, line 848: If I see this method I would have no clue what a "Real" incoming webhook would be. If it's just a different type of webhook which may not have a token it should rather be called `getAllIncomingWebhooks`

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -1992,7 +1992,15 @@ default ServerUpdater createUpdater() {
     CompletableFuture<List<Webhook>> getWebhooks();
 
     /**
-     * Gets a list of all incoming webhooks in this server.
+     * Gets a list of all incoming webhooks in this server, they are not guaranteed to have an accessible token.
+     *
+     * @return A list of all incoming webhooks in this server.
+     */
+    CompletableFuture<List<Webhook>> getRealIncomingWebhooks();
```
KILLEliteMaste, line 1999: Same here as above

javacord-core/src/main/java/org/javacord/core/entity/channel/InternalTextChannel.java
```diff
@@ -186,14 +187,30 @@
...
         return new RestRequest<List<IncomingWebhook>>(getApi(), RestMethod.GET, RestEndpoint.CHANNEL_WEBHOOK)
                 .setUrlParameters(getIdAsString())
                 .execute(result -> {
                     List<IncomingWebhook> webhooks = new ArrayList<>();
                     for (JsonNode webhookJson : result.getJsonBody()) {
-                        if (webhookJson.get("type").asText().equals("1")) {
+                        // check if it has a token, as only incoming webhooks have one
+                        if (webhookJson.hasNonNull("token")) {
                             webhooks.add(new IncomingWebhookImpl(getApi(), webhookJson));
                         }
```
KILLEliteMaste, line 215: Why don't you check here if it's a webhhok of type `INCOMING` like in `getRealIncomingWebhooks`?

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1433,14 +1433,30 @@ public void selfUndeafen() {
...
+    }
+
     @Override
     public CompletableFuture<List<IncomingWebhook>> getIncomingWebhooks() {
         return new RestRequest<List<IncomingWebhook>>(getApi(), RestMethod.GET, RestEndpoint.SERVER_WEBHOOK)
                 .setUrlParameters(getIdAsString())
                 .execute(result -> {
                     List<IncomingWebhook> webhooks = new ArrayList<>();
                     for (JsonNode webhookJson : result.getJsonBody()) {
-                        if (WebhookType.fromValue(webhookJson.get("type").asInt()) == WebhookType.INCOMING) {
```
KILLEliteMaste, line 1443: Same here. Why did you remove the check?

javacord-core/src/main/java/org/javacord/core/entity/channel/InternalTextChannel.java
Joshix-1: I didn't think the check was necessary.

KILLEliteMaste: But then why did you leave it in the newly added `getRealIncomingWebhooks` method?

Joshix-1: ups, that was a mistake sorry

KILLEliteMaste: Also I would leave the check. Because once discord decides to return a different type of webhook, this may no longer work

RedsTom: Maybe combining the two conditions should be fine :
```java
/* 213 */ if (webhookJson.hasNonNull("token") && webhookJson.get("type").asText().equals("1")) {
/* 214 */     webhooks.add(new IncomingWebhookImpl(getApi(), webhookJson));
/* 215 */ }
```

Joshix-1:
I think I fixed all the addressed issues. I rebased some methods to remove duplicated code.

KILLEliteMaste (approved):
LGTM!

PR #779, opened 2021-06-13
Add the ability to modify the application command permissions

KILLEliteMaste:
# Example usage:

```java
//Update a single comand for a user
new ApplicationCommandPermissionsUpdater(server)
                .addPermission(ApplicationCommandPermissions.createApplicationCommandPermissions(USER_ID, ApplicationCommandPermissionType.USER, false))
                .update(APPLICATION_COMMAND_ID);

//Update multiple permissions at once
api.batchUpdateApplicationCommandPermissions(
    server,
    Arrays.asList(new ServerApplicationCommandPermissionsBuilder(APPLICATION_COMMAND_ID, Collections.singletonList(ApplicationCommandPermissions.createApplicationCommandPermissions(USER_ID, ApplicationCommandPermissionType.USER, false))),
                  new ServerApplicationCommandPermissionsBuilder(APPLICATION_COMMAND_ID, Collections.singletonList(ApplicationCommandPermissions.createApplicationCommandPermissions(ROLE_ID, ApplicationCommandPermissionType.ROLE, false)))));
```

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandPermissionsBuilderDelegateImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandPermissionsBuilderDelegateImpl.java
@@ -1,0 +1,48 @@
+package org.javacord.core.interaction;
+
+import org.javacord.api.interaction.ApplicationCommandPermissionType;
+import org.javacord.api.interaction.ApplicationCommandPermissions;
+import org.javacord.api.interaction.internal.ApplicationCommandPermissionsBuilderDelegate;
+
+import java.util.Objects;
+
+public class ApplicationCommandPermissionsBuilderDelegateImpl implements ApplicationCommandPermissionsBuilderDelegate {
+
+    private Long id;
+    private ApplicationCommandPermissionType type;
+    private Boolean permission;
+
+    /**
+     * Class constructor.
+     */
+    public ApplicationCommandPermissionsBuilderDelegateImpl() {
+    }
+
+    @Override
+    public void setId(long id) {
+        this.id = id;
+    }
+
+    @Override
+    public void setType(ApplicationCommandPermissionType type) {
+        Objects.requireNonNull(type,"The type can not be null!");
+        this.type = type;
+    }
+
+    @Override
+    public void setPermission(boolean permission) {
+        this.permission = permission;
+    }
+
+    @Override
+    public ApplicationCommandPermissions build() {
+        if (null == type) {
+            throw new IllegalStateException("Type can not be null!");
+        } else if (null == permission) {
+            throw new IllegalStateException("Permission can not be null!");
+        } else if (null == id) {
+            throw new IllegalStateException("ID can not be null!");
+        }
+        return new ApplicationCommandPermissionsImpl(id, type, permission);
+    }
+}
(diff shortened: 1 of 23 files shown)


KILLEliteMaste:
One thing that is a bit different that anything else in Javacord is the ServerApplicationCommandPermissionsBuilder. This is not really a builder rather a wrapper for 2 values which are needing for a batchUpdateApplicationCommandPermissions update. The wrapper is a partial ServerApplicationCommanPermissions. Otherwise I would have to make some values optional for the ServerApplicationCommanPermissions and therefore I decided to do a little wrapper class.

coaxial1:
I'm just thinking about wether i'd be a good idea to shorten calls like ``ApplicationCommandPermissions.createApplicationCommandPermissions(...)`` to ``ApplicationCommandPermissions.with(...)`` or ``ApplicationCommandPermissions.create(...)`` since the first one basically duplicates the name and is very long 🤔

coaxial1:
Or maybe even a shorthand for the application command permission updater, like
```
new ApplicationCommandPermissionsUpdater(server)
                .addPermission(USER_ID, ApplicationCommandPermissionType.USER, false)
                .update(APPLICATION_COMMAND_ID);
 ```
 instead of your first example, to make it more convenient to use

**Comment by KILLEliteMaste** [comment]
> I'm just thinking about wether i'd be a good idea to shorten calls like `ApplicationCommandPermissions.createApplicationCommandPermissions(...)` to `ApplicationCommandPermissions.with(...)` or `ApplicationCommandPermissions.create(...)` since the first one basically duplicates the name and is very long 🤔

Yeah makes sense. I think this would also follow the style we are currently doing with most other classes.

> Or maybe even a shorthand for the application command permission updater

I could just add it a another method so we have both variants

File: javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandPermissionsBuilderDelegateImpl.java [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void setPermission(boolean permission) {
+        this.permission = permission;
+    }
+
+    @Override
+    public ApplicationCommandPermissions build() {
+        if (null == type) {
+            throw new IllegalStateException("Type can not be null!");
```
coaxial1, line 37: I'd suggest to just perform these 3 null checks with ``Objects.requireNonNull(Object, String)``.

coaxial1 (changes requested):
Just one very small thing I'd change. Afterwards, LGTM :)

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandPermissionsBuilderDelegateImpl.java
KILLEliteMaste: I don't think so. If it should throw a NPE we should move the checks to the setter permission but if one of these properties is not set, the ApplicationCommandPermissions can not be built correctly so I would think it's rather an ***illegal state*** of the Builder.  What do you think about that?

coaxial1: Hmh, yeah, I guess you're right about the semantics, so let's keep it.
Let's add the requireNonNull checks to the setters as well (fail-fast principle), though?

KILLEliteMaste: Sure

coaxial1: I'd still vote for adding it 👍 
Let's not allow users to set type to null actively :D
The primitive setter can't be set to null, this would be prevented / indicated earlier (either by a JVM NPE when unboxing a primitive wrapper class or by the IDE when actually typing ``null``), so let's have equal protection for the type setter. :) You are right that we should still keep the IllegaleStateCheck for cases where the user doesn't call the setter at all
:)

coaxial1 (approved):

PR #782, opened 2021-06-15
Add Message reference object to fix issues with message reference

Joshix-1:
See: https://discord.com/developers/docs/resources/channel#message-reference-object-message-reference-structure

As there now can be message references without a message id, we need to implement it properly.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -768,12 +768,22 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     MessageAuthor getAuthor();
 
     /**
-     * Gets the id of the message referenced with a reply.
-     * Only present if this message is type {@code MessageType.REPLY}.
+     * Gets the message reference of the message.
+     *
+     * @return The message reference.
+     */
+    Optional<MessageReference> getMessageReference();
+
+    /**
+     * Gets the id of the referenced message.
      *
      * @return The id of the referenced message.
+     * @deprecated Use {@link #getMessageReference()} instead.
      */
-    Optional<Long> getReferencedMessageId();
+    @Deprecated
+    default Optional<Long> getReferencedMessageId() {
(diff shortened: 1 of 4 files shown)


KILLEliteMaste:
This is a duplicate of #745 but as this is a bit more complete/feature rich I am going to close my PR

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -768,12 +768,20 @@ default URL getLink() throws AssertionError {
...
+     */
+    Optional<MessageReference> getMessageReference();
+
+    /**
+     * Gets the id of the referenced message.
      *
      * @return The id of the referenced message.
      */
-    Optional<Long> getReferencedMessageId();
+    default Optional<Long> getReferencedMessageId() {
```
KILLEliteMaste, line 782: We had this discussion in the other PR. I would still vote for removing this(even if it's going to be a breaking change) as I dont think there are that many people that uses this even at all and this just clutters the Message interface.

Joshix-1: I kept this, because we wanted to keep the next release non-breaking. I would suggest we deprecate it.

KILLEliteMaste: Okay, I can agree with that. Do you mind already doing a followup PR which removes `getReferencedMessageId` once this is merged?

Joshix-1: Wouldn't it be smarter to do it after the next release to then increase the snapshot version with it?

KILLEliteMaste: That doesn't really matter. There are already quiet a few breaking PR's that would be merged into 3.4.0 so any of these could increase the version too. You could also of course do it directly in the PR where you remove the method.

KILLEliteMaste (approved):
LGTM!
Can you please squash the PR into a single commit?

PR #786, opened 2021-06-17
General component fixes

KILLEliteMaste:
This PR changes some method names and i.e. Button.primary now builds the object right away instead of returning the builder.
Changed the ButtonBuilder and ActionRowBuilder to build the actual object which are then passed as the arguments for i.e. the messagebuilder.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
@@ -51,432 +51,354 @@ public interface Button extends LowLevelComponent {
     Optional<Emoji> getEmoji();
 
     /**
-     * Create a new button builder with the red danger style and the given properties.
+     * Create a new button with the red danger style and the given properties.
      * This is a convenience shorthand method.
      *
      * @param customId The custom id for this button.
-     * @param label    The label for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder danger(String customId, String label) {
-        return with(customId, ButtonStyle.DANGER, label, (Emoji) null);
-    }
-
-    /**
-     * Create a new button builder with the red danger style and the given properties.
-     * This is a convenience shorthand method.
-     *
-     * @param customId The custom id for this button.
-     * @param label    The label for this button.
-     * @param emoji    The emoji for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder danger(String customId, String label, Emoji emoji) {
-        return with(customId, ButtonStyle.DANGER, label, emoji);
-    }
-
-    /**
-     * Create a new button builder with the red danger style and the given properties.
-     * This is a convenience shorthand method.
-     *
-     * @param customId     The custom id for this button.
-     * @param label        The label for this button.
-     * @param unicodeEmoji The emoji for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder danger(String customId, String label, String unicodeEmoji) {
-        return with(customId, ButtonStyle.DANGER, label, unicodeEmoji);
-    }
-
-    /**
-     * Create a new button builder with the red danger style and the given emoji.
-     * Due to signature collisions, there is no such shorthand for unicode emojis; please use
-     * the {@code with(...)} method instead.
-     * This is a convenience shorthand method.
-     *
-     * @param customId The custom id for this button.
-     * @param emoji    The emoji for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder danger(String customId, Emoji emoji) {
-        return with(customId, ButtonStyle.DANGER, null, emoji);
-    }
-
-    /**
-     * Create a new button builder with the red danger style.
-     * Note: You MUST call at least one of {@code setEmoji()} or {@code setLabel()} on the returned builder,
-     * as the button will be invalid otherwise.
-     * This is a convenience shorthand method.
-     *
-     * @param customId The custom id for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder danger(String customId) {
-        return with(customId, ButtonStyle.DANGER);
-    }
-
-    /**
-     * Create a new button builder with the blurple primary style and the given properties.
-     * This is a convenience shorthand method.
-     *
-     * @param customId The custom id for this button.
-     * @param label    The label for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder primary(String customId, String label) {
-        return with(customId, ButtonStyle.PRIMARY, label, (Emoji) null);
-    }
-
-    /**
-     * Create a new button builder with the blurple primary style and the given properties.
-     * This is a convenience shorthand method.
-     *
-     * @param customId The custom id for this button.
-     * @param label    The label for this button.
-     * @param emoji    The emoji for this button.
-     * @return the new button builder to be used with an ActionRow or a message
-     */
-    static ButtonBuilder primary(String customId, String label, Emoji emoji) {
-        return with(customId, ButtonStyle.PRIMARY, label, emoji);
-    }
-
-    /**
-     * Create a new button builder with the blurple primary style and the given properties.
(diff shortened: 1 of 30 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandOption.java
```diff
@@ -57,115 +57,125 @@
...
      * This is a convenience method.
      *
-     * @param type        The type of the option.
-     * @param name        The name of the option.
+     * @param type The type of the option.
+     * @param name The name of the option.
      * @param description The description of the option.
      * @return The new application command option builder.
      */
     static ApplicationCommandOptionBuilder with(ApplicationCommandOptionType type, String name, String description) {
```
coaxial1, line 87: Did you forget to change this one? It should also return the option itself instead of the builder I guess, and be renamed to "create" like the other methods.

javacord-core/src/main/java/org/javacord/core/entity/message/component/ActionRowImpl.java
```diff
@@ -35,16 +38,59 @@ public ActionRowImpl(JsonNode data) {
...
+
+        if (components.size() == 0) {
+            object.putArray("components");
+            return object;
+        }
+
+        ArrayNode componentsJson = JsonNodeFactory.instance.objectNode().arrayNode();
+        for (LowLevelComponent component : this.components) {
+            switch (component.getType()) {
+                case ACTION_ROW:
```
coaxial1, line 79: I don't think you need to add this case since your iterating over components of type ``LowLevelComponent`` which ActionRow doesn't implement, so this case can't happen

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
```diff
@@ -51,356 +51,361 @@
...
      * as the button will be invalid otherwise.
      * This is a convenience shorthand method.
      *
      * @param customId The custom id for this button.
-     * @return the new button builder to be used with an ActionRow or a message
+     * @return the new button to be used with an ActionRow or a message
      */
-    static ButtonBuilder success(String customId) {
-        return with(customId, ButtonStyle.SUCCESS);
+    static Button success(String customId) {
```
coaxial1, line 311: This method MUST return a builder instead of a button (please read the javadoc comment above it). If you immediately build, this will not yield a valid button as it does not have a label nor an emoji (empty buttons are not possible).

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
```diff
@@ -51,356 +51,361 @@
...
      * as the button will be invalid otherwise.
      * This is a convenience shorthand method.
      *
      * @param customId The custom id for this button.
-     * @return the new button builder to be used with an ActionRow or a message
+     * @return the new button to be used with an ActionRow or a message
      */
-    static ButtonBuilder secondary(String customId) {
-        return with(customId, ButtonStyle.SECONDARY);
+    static Button secondary(String customId) {
```
coaxial1, line 245: This method MUST return a builder instead of a button (please read the javadoc comment above it). If you immediately build, this will not yield a valid button as it does not have a label nor an emoji (empty buttons are not possible).

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
```diff
@@ -51,356 +51,361 @@
...
      * as the button will be invalid otherwise.
      * This is a convenience shorthand method.
      *
      * @param customId The custom id for this button.
-     * @return the new button builder to be used with an ActionRow or a message
+     * @return the new button to be used with an ActionRow or a message
      */
-    static ButtonBuilder primary(String customId) {
-        return with(customId, ButtonStyle.PRIMARY);
+    static Button primary(String customId) {
```
coaxial1, line 179: This method MUST return a builder instead of a button (please read the javadoc comment above it). If you immediately build, this will not yield a valid button as it does not have a label nor an emoji (empty buttons are not possible).

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
```diff
@@ -51,356 +51,361 @@
...
      * as the button will be invalid otherwise.
      * This is a convenience shorthand method.
      *
      * @param customId The custom id for this button.
-     * @return the new button builder to be used with an ActionRow or a message
+     * @return the new button to be used with an ActionRow or a message
      */
-    static ButtonBuilder danger(String customId) {
-        return with(customId, ButtonStyle.DANGER);
+    static Button danger(String customId) {
```
coaxial1, line 114: This method MUST return a builder instead of a button (please read the javadoc comment above it). If you immediately build, this will not yield a valid button as it does not have a label nor an emoji (empty buttons are not possible).

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
```diff
@@ -51,356 +51,361 @@
...
      * Note: You MUST call at least one of {@code setEmoji()} or {@code setLabel()} on the returned builder,
      * as the button will be invalid otherwise.
      * This is a convenience shorthand method.
      *
      * @param url The url for this link button.
-     * @return the new button builder to be used with an ActionRow or a message
+     * @return the new button to be used with an ActionRow or a message
      */
-    static ButtonBuilder link(String url) {
+    static Button link(String url) {
```
coaxial1, line 395: This method MUST return a builder instead of a button (please read the javadoc comment above it). If you immediately build, this will not yield a valid button as it does not have a label nor an emoji (empty buttons are not possible).

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
```diff
@@ -51,356 +51,361 @@
...
      * This is a convenience shorthand method.
      *
      * @param customId The custom id for this button.
-     * @param style    The button style for this button.
-     * @param label    The label for this button.
+     * @param style The button style for this button.
+     * @param label The label for this button.
      * @return the new button builder to be used with an ActionRow or a message
      */
     static ButtonBuilder with(String customId, ButtonStyle style, String label) {
```
coaxial1, line 411: Why didn't you change the ``with(...)`` methods to return a full button as well? A button with a style, customid and label would be perfectly valid.
Same for the methods below.

coaxial1 (changes requested):
Please see the comments below :)

javacord-core/src/main/java/org/javacord/core/entity/message/component/ActionRowImpl.java
KILLEliteMaste: This is copied from the ActionRowBuilder from a previous PR so I didn't had a look at the code anymore and just adjusted it where necessary.

javacord-api/src/main/java/org/javacord/api/entity/message/component/Button.java
KILLEliteMaste: Doing this clashes with the button methods returning a builder. `static ButtonBuilder dangerWith(String customId)`. I would rather vote for removing these methods as we already have methods to create any kind of valid button. What do you think?

coaxial1: The ``with()`` methods have the advantage that they accept a style as parameter, so it can be set programmatically if a lib user needs to accept a dynamic style instruction from elsewhere. If we remove them, they'd need to use ``new ButtonBuilder()`` or a switch statement (to choose the correct danger, primary, ... shorthand) which can both be avoided by simply letting these methods survive. :)

KILLEliteMaste: I mean removing the methods that only have the custom ID like `danger(String customId)` as they are have to return the builder and need an additional parameter. And we actually have for any case already a working method which returns a valid button if I've seen correctly.

coaxial1: Yeah, you're absolutely right. They don't make munch sense in this context anymore. I'd as well vote for removing shorthands that produce an invalid button in all cases. Sorry for the misunderstanding

coaxial1 (approved):

PR #787, opened 2021-06-18
Add slash commands example to the README

KILLEliteMaste:
This is just a minimal basic example of how you can respond to slash commands. I am open to suggestions/code examples to extend the README

### DIFF
--- a/.github/README_template.md
+++ b/.github/README_template.md
@@ -39,6 +39,48 @@ public class MyFirstBot {
 More sophisticated examples can be found at the [end of the README](#-more-examples). 
 You can also check out the [example bot](https://github.com/Javacord/Example-Bot) for a fully functional bot.
 
+### */* Slash commands
+First, you need to create the ping pong slash command. Run this code **one single time** to create the command:
+
+```java
+public class MyFirstBot {
+
+    public static void main(String[] args) {
+        // Insert your bot's token here
+        String token = "your token";
+
+        DiscordApi api = new DiscordApiBuilder().setToken(token).login().join();
+
+        ApplicationCommand.with("ping", "A simple ping pong command!").createGlobal(api).join();
+    }
+}
+```
+
+Discord now knows about your command and will offer it in your text channels if you type ``/``.
+*Note: Creating a global command may take up to 1 hour to become usable* 
+
+Next, let's see how we can let the bot send answers to this simple slash command:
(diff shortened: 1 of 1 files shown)


coaxial1:
Hmm, if someone copy pastes this code, it won't work. We need to add some code to create the command first if it does not exist yet, right? 🤔

KILLEliteMaste:
> Hmm, if someone copy pastes this code, it won't work. We need to add some code to create the command first if it does not exist yet, right? 🤔

Yes. I am not sure how in depth we should go in the readme as the wiki should explain everything.

coaxial1:
I think we should add a full example. We could do it like this (text below could be pasted right below the slash commands headline).

-----

First, you need to create the ping pong slash command. Run this code **one single time** to create the command:

```java
public class MyFirstBot {

        // Insert your bot's token here
        String token = "your token";

        DiscordApi api = new DiscordApiBuilder().setToken(token).login().join();

        ApplicationCommand.with("ping", "A simple ping pong command!").createGlobal(api).join();
    }
}
```

Discord now knows about your command and will offer it in your text channels if you type ``/``.

Next, let's see how we can let the bot send answers to this simple slash command:

[... the code you already wrote ...]

coaxial1:
Alternatively, we could link to something like this: https://slash-commands-gui.netlify.app
But imo we should present the Javaacord API to do it. So let's just have 2 snippets.

KILLEliteMaste:
Yeah, we shouldn't link to a 3rd party tool in the README. If we wanted to we could mention it in the wiki

.github/README_template.md
```diff
@@ -39,6 +39,50 @@ public class MyFirstBot {
...
+    }
+}
+```
+
+Discord now knows about your command and will offer it in your text channels if you type ``/``.
+*Note: Creating a global command may take up to 1 hour to become usable* 
+
+Next, let's see how we can let the bot send answers to this simple slash command:
+
+The following example shows the basic slash command aka. interaction command response for the previously created `ping` command:
```
coaxial1, line 64: I think we should either have only this line or line 62 (?) :)

KILLEliteMaste: Yeah, you are right ^^

coaxial1 (approved):
LGTM! I think that's more than enough for a simple README example.

PR #789, opened 2021-06-19
Add Multi-embed Support for Discord Bots

### LINKED ISSUE #772: Add multi embed support for bot messages
Discord now allows bots to send up to 10 embeds per message and therefore gets closer to the webhook behaviour.
https://github.com/discord/discord-api-docs/pull/3105

ShindouMihou:
This will add multi-embed support for the bot side. I don't know what to do with `.setEmbed()` but I kept it there for non-breaking change, though I also don't see any wrong with keeping it there since it works nonetheless.

Actually, I wonder if we should keep the behavior of `setEmbed()` which is to make it the only embed on the message?

Fixes #772


KILLEliteMaste (commented):
Thank you for the PR!
As far as I've seen you have just added the ability to add multiple embeds to the MessageBuilder and haven't touched the WebhookMessageBuilder. Atm I think they are 1:1 except for the WebhookBuilder being able to set an avatar and display name. Therefore you can move the embed methods from the WebhookMessageBuilderDelegate to the MessageBuilderDelegate and be able reuse them as we currently have duplicated code.

ShindouMihou:
> Thank you for the PR!
> As far as I've seen you have just added the ability to add multiple embeds to the MessageBuilder and haven't touched the WebhookMessageBuilder. Atm I think they are 1:1 except for the WebhookBuilder being able to set an avatar and display name. Therefore you can move the embed methods from the WebhookMessageBuilderDelegate to the MessageBuilderDelegate and be able reuse them as we currently have duplicated code.

Oh, move just the embed methods. I remember being asked about this but was confused over how to do it, taking a look at it again, the WebhookMessageBuilderDelegate does extend over MessageBuilderDelegate which means all I have to do is move the embed methods to the MessageBuilderDelegate, correct? (or am I getting something else confused?) >.<

KILLEliteMaste:
> > Thank you for the PR!
> > As far as I've seen you have just added the ability to add multiple embeds to the MessageBuilder and haven't touched the WebhookMessageBuilder. Atm I think they are 1:1 except for the WebhookBuilder being able to set an avatar and display name. Therefore you can move the embed methods from the WebhookMessageBuilderDelegate to the MessageBuilderDelegate and be able reuse them as we currently have duplicated code.
> 
> Oh, move just the embed methods. I remember being asked about this but was confused over how to do it, taking a look at it again, the WebhookMessageBuilderDelegate does extend over MessageBuilderDelegate which means all I have to do is move the embed methods to the MessageBuilderDelegate, correct? (or am I getting something else confused?) >.<

Yes, that's how inheritence works :)

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -153,11 +153,23 @@ public MessageBuilder setContent(String content) {
      * @return The current instance in order to chain call methods.
      */
     public MessageBuilder setEmbed(EmbedBuilder embed) {
```
KILLEliteMaste, line 155: You should add a new method `setEmbeds(EmbedBuilder... embeds)`

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
ShindouMihou: Basically `addEmbeds(...)` but overrides all the existing embeds?

KILLEliteMaste: Yes

ShindouMihou:
Also added a note on the Javadoc that tells them it will override all existing embeds that were set (to help them differentiate the difference between the two despite the names already explaining everything), that should be fine?

KILLEliteMaste (approved):
LGTM!

ShindouMihou:
Squashed and ready to go anytime. Thanks for reviewing!

PR #790, opened 2021-06-20
Add Membership Screening Pending getter and event.

### LINKED ISSUE #690: Add membership screening pending getter and event for when guild rules are accepted
Discord added a new membership screening functionality which requires people to accept the server rules before being able to interact with the server. 

Whether or not they have accepted that is exposed via the `pending` field on the [Guild Member Object](https://github.com/discord/discord-api-docs/blob/eeba53e05a5bb929e7fbf9487a02dceb34811586/docs/resources/Guild.md#guild-member-object). It would be nice to have a getter for that as well as an event which gets fired once users have done so in order to be able to automatically assign roles after that. (Assigning them before will make it so that they don't need to accept that)

ShindouMihou:
As the title implies, it adds the `pending` field of [Guild Member](https://github.com/discord/discord-api-docs/blob/eeba53e05a5bb929e7fbf9487a02dceb34811586/docs/resources/Guild.md#guild-member-object) which is a field that moderation bots or auto-role bots may require since it is needed to check whether the user has passed membership screening.

This PR has been tested (both the listener and get method: `User#isPending`, `Server#isPending` and `UserChangePendingListener`) though I feel like the JavaDocs and the name for the event may not fit well, feel free to leave your opinions.

Fixes #690

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -178,6 +178,20 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
     Optional<String> getNickname(User user);
 
     /**
+     * Gets the pending state of the user.
+     *
+     * <p>Membership screening pending state is only available for servers that have
+     * verification gate enabled, this will default to false if it isn't.
+     * You can check if a server has membership verification gate enabled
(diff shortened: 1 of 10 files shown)


KILLEliteMaste (approved):
LGTM!

ShindouMihou:
Squashed and ready to go anytime! Thanks for reviewing! 👍

PR #791, opened 2021-06-22
Add specific events for interactions

KILLEliteMaste:
- ApplicationCommandCreateEvent
- MessageComponentCreateEvent
- ButtonClickEvent

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/event/interaction/ApplicationCommandCreateEvent.java
+++ b/javacord-api/src/main/java/org/javacord/api/event/interaction/ApplicationCommandCreateEvent.java
@@ -1,0 +1,40 @@
+package org.javacord.api.event.interaction;
+
+import org.javacord.api.event.Event;
+import org.javacord.api.interaction.ApplicationCommandInteraction;
+import org.javacord.api.interaction.Interaction;
+import java.util.Optional;
+
+/**
+ * An application command create event.
+ */
+public interface ApplicationCommandCreateEvent extends Event {
+
+    /**
+     * Gets the created interaction.
+     *
+     * @return The interaction.
+     */
+    Interaction getInteraction();
+
+    /**
+     * Gets the created interaction as ApplicationCommandInteraction, if the interaction is of this type.
+     *
+     * @return The interaction.
+     */
+    default ApplicationCommandInteraction getApplicationCommandInteraction() {
+        return getInteraction().asApplicationCommandInteraction().get();
+    }
+
+    /**
+     * Gets the created interaction as ApplicationCommandInteraction, if the interaction is of this type and the
+     * command id equals the given command id.
+     *
+     * @param commandId The command it to match.
+     * @return The interaction.
+     */
+    default Optional<ApplicationCommandInteraction> getApplicationCommandInteractionWithCommandId(long commandId) {
+        return getInteraction().asApplicationCommandInteractionWithCommandId(commandId);
+    }
+
+}
--- a/javacord-api/src/main/java/org/javacord/api/event/interaction/ButtonClickEvent.java
+++ b/javacord-api/src/main/java/org/javacord/api/event/interaction/ButtonClickEvent.java
@@ -1,0 +1,40 @@
+package org.javacord.api.event.interaction;
+
+import org.javacord.api.event.Event;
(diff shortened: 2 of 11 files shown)


javacord-api/src/main/java/org/javacord/api/event/interaction/ApplicationCommandCreateEvent.java
```diff
@@ -0,0 +1,40 @@
...
+ * An application command create event.
+ */
+public interface ApplicationCommandCreateEvent extends Event {
+
+    /**
+     * Gets the created interaction.
+     *
+     * @return The interaction.
+     */
+    Interaction getInteraction();
```
coaxial1, line 18: Why do we have this? Shouldn't it always be an ApplicationCommandInteraction in this event?
This sould be returning the substype imo, and the following optional method can be removed

javacord-api/src/main/java/org/javacord/api/event/interaction/ButtonClickEvent.java
```diff
@@ -0,0 +1,40 @@
...
+ * A button click event.
+ */
+public interface ButtonClickEvent extends Event {
+
+    /**
+     * Gets the created interaction.
+     *
+     * @return The interaction.
+     */
+    Interaction getInteraction();
```
coaxial1, line 19: Same as in ApplicationCommandCreateEvent

javacord-api/src/main/java/org/javacord/api/event/interaction/MessageComponentCreateEvent.java
```diff
@@ -0,0 +1,40 @@
...
+ * A message component interaction create event.
+ */
+public interface MessageComponentCreateEvent extends Event {
+
+    /**
+     * Gets the created interaction.
+     *
+     * @return The interaction.
+     */
+    Interaction getInteraction();
```
coaxial1, line 18: Same

coaxial1 (changes requested):
BTW, I couldn't read it from the code somehow, but can we already attach a button click listener to a ``Button`` (e.g. after we added one to a message we just sent)?

javacord-api/src/main/java/org/javacord/api/event/interaction/ApplicationCommandCreateEvent.java
KILLEliteMaste: This is exactly the interaction I mentioned in Discord. You also said this should not be removed as this is the top level interaction which someone might want

coaxial1: I meant we shouldn't remove the top level interaction listener. On the top leven interaction create event, there should be a way to get the top level interaction. I thought you were talking about removing it from the top level create event in favor of the casting methods we added to it.

coaxial1: Somehow I didn't realize that you're talking about the new events :D

KILLEliteMaste:
> BTW, I couldn't read it from the code somehow, but can we already attach a button click listener to a `Button` (e.g. after we added one to a message we just sent)?
I haven't added this in the PR as I just wanted to have the more specific event. This could also be added in a followup PR

javacord-api/src/main/java/org/javacord/api/event/interaction/ApplicationCommandCreateEvent.java
KILLEliteMaste: But removing it from the new events result in the same. If someone want to pass the interaction to somewhere he might want to get the top level interaciton. The only difference is that the event is of this specific type but if the user wants to store it or has a global interacion handler he could get the top level interaction. So imo either remove it from all events or none

coaxial1: Okay, agreed.
Then my only objection is that the type specfic interaction getter shouldn't return optional here, but the interaction directly, like you did in the other events :)

KILLEliteMaste: Oh yeah, you are right. Missed that somehow :)

Issue #795, opened 2021-06-30
Couldn't handle packet of type INTERACTION_CREATE


Bastian:
```
2021-06-30 21:27:11.979+0200 WARN org.javacord.core.util.gateway.PacketHandler Couldn't handle packet of type INTERACTION_CREATE. Please contact the developer! (packet: {"version":1,"type":3,"token":"<redacted>","message":{"id":"859877451131256832","flags":64},"member":{"user":{"username":"Bastian","public_flags":768,"id":"157862224206102529","discriminator":"0310","avatar":"b69de4679e4ee28c5235e75a6eca4aa3"},"roles":["151325992038825984"],"premium_since":null,"permissions":"137438953471","pending":false,"nick":"Bastian","mute":false,"joined_at":"2016-03-11T15:31:25.546000+00:00","is_pending":false,"deaf":false,"avatar":null},"id":"859877752002052116","guild_id":"151037561152733184","data":{"custom_id":"success","component_type":2},"channel_id":"151326329839681536","application_id":"157889000391180288"}) {shard=0} 
 java.lang.AssertionError
	at java.util.Optional.orElseThrow(Optional.java:290)
	at org.javacord.core.interaction.InteractionImpl.<init>(InteractionImpl.java:60)
	at org.javacord.core.interaction.MessageComponentInteractionImpl.<init>(MessageComponentInteractionImpl.java:37)
	at org.javacord.core.interaction.ButtonInteractionImpl.<init>(ButtonInteractionImpl.java:18)
	at org.javacord.core.util.handler.interaction.InteractionCreateHandler.handle(InteractionCreateHandler.java:61)
	at org.javacord.core.util.gateway.PacketHandler.lambda$handlePacket$0(PacketHandler.java:51)
```

This happened when clicking on a button that was created with the following code:
```java
SlashCommand.with("example", "An example command").createForServer(api.getServerById(151037561152733184L).get()).join();

api.addSlashCommandCreateListener(event -> {
    SlashCommandInteraction slashCommandInteraction = event.getSlashCommandInteraction();
    if (slashCommandInteraction.getCommandName().equals("example")) {
        slashCommandInteraction.createImmediateResponder()
            .setContent("Slash commands are cool!")
            .setFlags(MessageFlag.EPHEMERAL)
            .addComponents(
                ActionRow.of(Button.success("success", "And"),
                    Button.danger("danger", "buttons"),
                    Button.secondary("secondary", "are too!")))
            .respond();
    }
});
```

[...]

PR #801, opened 2021-07-03
Fix NPE when invoking a slash command in DMs and fix for #795

KILLEliteMaste:


coaxial1 (changes requested):
LGTM excpet a minor thing

coaxial1 (approved):

PR #810, opened 2021-07-09
Added ability to edit message with message builder

Morazzer:
Added ability to edit messages with the message builder.

Issue: #575

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
@@ -1,35 +1,25 @@
 package org.javacord.api.entity.message;
 
 import org.javacord.api.DiscordApi;
-import org.javacord.api.entity.Icon;
-import org.javacord.api.entity.Mentionable;
 import org.javacord.api.entity.channel.TextChannel;
-import org.javacord.api.entity.message.component.HighLevelComponent;
-import org.javacord.api.entity.message.component.LowLevelComponent;
-import org.javacord.api.entity.message.embed.EmbedBuilder;
-import org.javacord.api.entity.message.internal.MessageBuilderDelegate;
-import org.javacord.api.entity.message.mention.AllowedMentions;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.entity.webhook.IncomingWebhook;
 import org.javacord.api.util.DiscordRegexPattern;
-import org.javacord.api.util.internal.DelegateFactory;
 
-import java.awt.image.BufferedImage;
-import java.io.File;
-import java.io.InputStream;
-import java.net.URL;
 import java.util.concurrent.CompletableFuture;
 import java.util.regex.Matcher;
 
 /**
  * This class can help you to create messages.
  */
-public class MessageBuilder {
+public class MessageBuilder extends MessageBuilderBase<MessageBuilder> {
 
     /**
-     * The message delegate used by this instance.
+     * Class constructor.
      */
-    protected final MessageBuilderDelegate delegate = DelegateFactory.createMessageBuilderDelegate();
+    public MessageBuilder() {
+        super(MessageBuilder.class);
+    }
 
     /**
      * Creates a message builder from a message.
@@ -55,133 +45,6 @@ public class MessageBuilder {
     }
 
     /**
-     * Add multiple high level components to the message.
-     *
-     * @param components The high level components.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder addComponents(HighLevelComponent... components) {
-        delegate.addComponents(components);
-        return this;
-    }
-
-    /**
-     * Add multiple low level components, wrapped in an ActionRow, to the message.
-     *
-     * @param components The low level components.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder addActionRow(LowLevelComponent... components) {
-        delegate.addActionRow(components);
-        return this;
-    }
-
-    /**
-     * Appends code to the message.
-     *
-     * @param language The language, e.g. "java".
-     * @param code     The code.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder appendCode(String language, String code) {
-        delegate.appendCode(language, code);
-        return this;
-    }
-
-    /**
-     * Appends a sting with or without decoration to the message.
-     *
-     * @param message The string to append.
-     * @param decorations The decorations of the string.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder append(String message, MessageDecoration... decorations) {
-        delegate.append(message, decorations);
-        return this;
-    }
-
-    /**
-     * Appends a mentionable entity (usually a user or channel) to the message.
-     *
-     * @param entity The entity to mention.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder append(Mentionable entity) {
-        delegate.append(entity);
-        return this;
-    }
-
-    /**
-     * Appends the string representation of the object (calling {@link String#valueOf(Object)} method) to the message.
-     *
-     * @param object The object to append.
-     * @return The current instance in order to chain call methods.
-     * @see StringBuilder#append(Object)
-     */
-    public MessageBuilder append(Object object) {
-        delegate.append(object);
-        return this;
-    }
-
-    /**
-     * Appends a new line to the message.
-     *
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder appendNewLine() {
-        delegate.appendNewLine();
-        return this;
-    }
-
-    /**
-     * Sets the content of the message.
-     * This method overwrites all previous content changes
-     * (using {@link #append(String, MessageDecoration...)} for example).
-     *
-     * @param content The new content of the message.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder setContent(String content) {
-        delegate.setContent(content);
-        return this;
-    }
-
-    /**
-     * Sets the embed of the message (overrides all existing embeds).
-     *
-     * @param embed The embed to set.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder setEmbed(EmbedBuilder embed) {
-        delegate.removeAllEmbeds();
-        delegate.addEmbed(embed);
-        return this;
-    }
-
-    /**
-     * Sets multiple embeds of the message (overrides all existing embeds).
-     *
-     * @param embeds The embed to set.
-     * @return The current instance in order to chain call methods.
-     */
-    public MessageBuilder setEmbeds(EmbedBuilder... embeds) {
-        delegate.removeAllEmbeds();
-        delegate.addEmbeds(embeds);
-        return this;
-    }
-
-    /**
-     * Adds an embed to the message.
-     *
-     * @param embed The embed to add.
-     * @return The current isntance in order to chain call methods.
-     */
-    public MessageBuilder addEmbed(EmbedBuilder embed) {
-        delegate.addEmbed(embed);
-        return this;
-    }
-
-    /**
      * Sets if the message should be text to speech.
      *
      * @param tts Whether the message should be text to speech or not.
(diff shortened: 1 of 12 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -670,4 +670,51 @@ public StringBuilder getStringBuilder() {
...
+     * @param message The message to edit.
+     * @param updateContent If the content of the message should be updated.
+     * @param updateEmbeds If the embeds of the message should be updated.
+     * @param updateComponents If the components of the message should be updated.
+     * @return The edited message.
+     */
+    public CompletableFuture<Message> edit(Message message,
+                                           boolean updateContent, boolean updateEmbeds, boolean updateComponents) {
+        return delegate.edit(message, updateContent, updateEmbeds, updateComponents);
+    }
```
KILLEliteMaste, line 718: Is there an actual use case for the booleans I didn't think of? Why can't you just remove the values from from the MessageBuilder or create a new?

coaxial1:
Since it seems like the original author isn't coming back to this (and I need this myself actually), I've changed this to remove the boolean flags and check for the content field to be filled at all.
@KILLEliteMaste Please lmk fi this makes sense for you now. :)

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
```diff
@@ -569,36 +569,32 @@ private static void executeWebhookRest(RestRequest<Message> request, boolean wai
...
+            for (int i = 0; i < embeds.size() && i < 10; i++) {
+                embedsNode.add(((EmbedBuilderDelegateImpl) embeds.get(i).getDelegate()).toJsonNode());
             }
+            body.set("embeds", embedsNode);
         }
 
-        if (updateComponents) {
-            prepareComponents(body);
-        }
+        prepareComponents(body);
```
KILLEliteMaste: I think we can reuse nearly everything from #send(TextChannel) for this. There are also a lot of things missing like the allowed mentions, flags...  See https://discord.com/developers/docs/resources/channel#edit-message-jsonform-params

If for some reason you do not reuse the code from the #send method you should also remove the check to update the message content as it is not possible atm to remove it.

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
coaxial1: For flags: We currently do not support flags at all, do we? I mean any of those: https://discord.com/developers/docs/resources/channel#message-object-message-flags

So all that's missing is allowed mentions and files/attachments.

> you should also remove the check to update the message content 

I don't understand this one. I've intentionally added the check. All requests trying to patch in an empty content will either fail (completely empty message won't work) or produce unexpected results (e.g. it'd remove content from a message that also contains an embed although this is not wanted), so we need to exclude it from the request if the content is empty / unset, right? This is important for requests where we e.g. only want to patch in some components:

```java
Message m = ...;
new MessageBuilder().addComponents(...).edit(m);
```
We do not use ``Message#toMessageBuilder()`` since we don't want to change anything else than components in this case, so we create a new one (like you suggested in your complaint about the boolean flags)

KILLEliteMaste: >  All requests trying to patch in an empty content will either fail (completely empty message won't work) or produce unexpected results (e.g. it'd remove content from a message that also contains an embed although this is not wanted), so we need to exclude it from the request if the content is empty / unset, right?

But how else should you be able to edit a message which contains a `content` and `embed` to only having an `embed`?

> For flags: We currently do not support flags at all, do we? I mean any of those:

We do: `EPHEMERAL` for example as this was the only Message flag available back this time we implemented it.

And taking your example
```java
Message m = ...;
new MessageBuilder().addComponents(...).edit(m);
```
You example assumes that we patch in the new stuff. I was thinking more about we replace the message with the the new MessageBuilder

coaxial1: > But how else should you be able to edit a message which contains a content and embed to only having an embed?

Yeah, that is a problem. But we should really avoid to send the entire message in the patch request, even tho we only change the components (for example) ... So how would we do it? We could compare everything against the Message object passed in as parameter to ``edit()`` and only send differences ... in theory ...

>  the only Message flag available back this time we implemented it.

I think you're confusing https://discord.com/developers/docs/interactions/slash-commands#interaction-response-object-interaction-application-command-callback-data-flags with the link I sent above; those are different types of flags! And the ones I linked above existed for a long time already when we added support for ephemeral interaction responses.

>  I was thinking more about we replace the message with the the new MessageBuilder

Yeah we could do that as well, but it would need to inlcude the difference checking stuff I mentioned above.

coaxial1: Alternatively to the difference chec which would be complex to do, we could track wether or not any of the methods affect ``content`` have been called on the MessageBuilder, and send an empty content field if they have been intentionally set to be empty.

Second alternative would be to have a new interface ``MessageUpdater``, similar to ``ServerUpdater``, to make the intention more clear. I agree that having a MessageBuilder with ``edit`` would imply that all fields of the builder are applied, even if empty (and thus replace the message).

KILLEliteMaste: > I think you're confusing...

Prorbably yeah. But since these are 2 different Enums our current implementation for `setFlags` in `InteractionMessageBuilderBase` and maybe some other places is wrong as this assumes MessageFlag which is then something different than the `Interaction Application Command Callback Data Flags`. 

I think introducing a message updater is a good idea. This clearly differenciates the MessageBuilder which sends a message to updating a message. And this also allows us to have different updating methods like `updateSetFields` and `replaceMessage` without cluttering everything in the MessageBuilder and it makes it clear how you are updating the message.

coaxial1: > our current implementation for setFlags ... is wrong

I don't think so. EPHEMERAL is indeed to only flag that can be set when sending interaction responses. Offering anything else would currently be misleading.
We could have another enum to represent the flags of an already sent message, maybe, for example to know wether or not a message has been crossposted or not, but I wouldn't consider that high priority. Might deserve an issue, though. Also, except for the SUPPRESS_EMBEDS flags nothing is updateable, so we could just have a method like ``Message#suppressEmbeds()`` to set the flag to true. That would be the same thing as the little "X" that appear on the top right of embeds in the UI.

> I think introducing a message updater is a good idea.

I'll create a new pull request for that approach. I also like it better than having an ``.edit()`` on the builder. I don't know wether or not we should make a difference between "UpdateSetFields" and "ReplaceMessage", though. The first one is how an updater usually works

KILLEliteMaste: > I don't think so. EPHEMERAL is indeed to only flag that can be set when sending interaction responses. Offering anything else would currently be misleading.

Because exactly of this the current implementation is not correct as MessageFlag also contains i.e. `SUPPRESS_EMBEDS`. So we have to create a 2nd enum which contains only the valid flags repsonding to an interaction.

> The first one is how an updater usually works

Yes, but I think there are actually more use cases for replacing a message than just adding e.g. a component to a message. It's also more clear what the message will actually look like when replacing it.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
```diff
@@ -0,0 +1,540 @@
...
+    private final Class<T> myClass;
+
+    public MessageBuilderBase(Class<T> myClass) {
+        this.myClass = myClass;
+    }
+
+    /**
+     * The message delegate used by this instance.
+     */
+    protected final MessageBuilderDelegate delegate = DelegateFactory.createMessageBuilderDelegate();
```
KILLEliteMaste, line 27: Name has to be updated correspondingly

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -0,0 +1,42 @@
...
+        return delegate.edit(message, true);
+    }
+
+    /**
+     * Edits the message, updating all fields.
+     * Fields that have not been set will be dropped and removed from the actual message on Discord.
+     *
+     * @return The edited message.
+     */
+    public CompletableFuture<Message> replaceMessage() {
```
KILLEliteMaste, line 39: I think calling them quiet similar would be better as you get all possible update methods when you type the first few letters. Something like `updateApplyChanges` and `updateReplace` but the mentioned names are probably not that nice. Maybe you know something better

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
```diff
@@ -567,17 +571,109 @@ private static void executeWebhookRest(RestRequest<Message> request, boolean wai
...
+        if (updateAll || attachmentsChanged) {
+            return checkForAttachmentsAndExecuteRequest(message.getChannel(), body, request, true);
+        } else {
+            return executeRequestWithoutAttachments(message.getChannel(), body, request);
+        }
+    }
+
+    private void prepareEmbeds(ObjectNode body, boolean evenIfEmpty) {
+        if (embeds.size() != 0 || evenIfEmpty) {
+            // Discord now supports multiple embeds for Discord bots.
```
KILLEliteMaste, line 603: Remove the comment as it's not helpful in any way

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
```diff
@@ -567,17 +571,109 @@ private static void executeWebhookRest(RestRequest<Message> request, boolean wai
         });
     }
 
+
```
KILLEliteMaste, line 574: Remove line

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
```diff
@@ -0,0 +1,540 @@
...
+import org.javacord.api.entity.message.internal.MessageBuilderDelegate;
+import org.javacord.api.entity.message.mention.AllowedMentions;
+import org.javacord.api.util.internal.DelegateFactory;
+
+import java.awt.image.BufferedImage;
+import java.io.File;
+import java.io.InputStream;
+import java.net.URL;
+
+abstract class MessageBuilderBase<T> {
```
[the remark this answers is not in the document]
coaxial1: But it's also not the base of a message ... :D
I think it's not critical to have a "correct" name here, even tho I think the name is fine anyways.
The updater sends a request to Discord, and is thereby building a new state of the message that's returned afterwards.
I don't really know a better name tbh.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
coaxial1: So you mean like "MessageBuilderBaseDelegate" (assuming that we keep the name of the base)?

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
coaxial1: Hmm...
What about sth like
``applyChanges``
and
``applyReplacing``

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderDelegateImpl.java
coaxial1: It's not from me, but yeah we can remove it

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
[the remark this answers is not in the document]
KILLEliteMaste: As this is also a class you wouldn't interact with normally I think if someone finds a better name we can change it without hesitation as this would normally be a breaking change. So I guess leave it for now 🤔

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
KILLEliteMaste: As I thought we would rename the class I just mentioned it here. But we should probably update it to include "base" as it's for exactly this class

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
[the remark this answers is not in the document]
coaxial1: Mhm. I mean it's still in the API package, but abstract so I wouldn't say that a user would usually use that type anywhere

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
KILLEliteMaste: It's always `replace` that doesn't sound good to me. What do you think about `applySwap` or `applyExchange` ? 
Maybe this is also a bad Idea and applyChanges and replaceMessage are just fine. Just thought about having a uniform update method would be easier for new users 😕

coaxial1: We could use ``updateChanges()`` or ``updateEverything()``?
Dunno. I also think that the existing names are fines. I wouldn't type any letter. I would type a dot and then have the IDE suggest the available methods, which would show both

KILLEliteMaste: Hm, then I guess we should leave it as it is currently.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -606,9 +700,7 @@ protected void prepareCommonWebhookMessageBodyParts(ObjectNode body) {
         if (strBuilder.length() != 0) {
             body.put("content", strBuilder.toString());
         }
-        if (allowedMentions != null) {
-            ((AllowedMentionsImpl) allowedMentions).toJsonNode(body.putObject("allowed_mentions"));
-        }
+        prepareAllowedMentions(body);
         if (embeds.size() != 0) {
```
KILLEliteMaste, line 704: You can replace this with `prepareEmbeds`. You should also replace the loop in the method with the used one here as the one in the method is has a hardcoded limit.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -567,17 +571,107 @@ private static void executeWebhookRest(RestRequest<Message> request, boolean wai
...
+
+    private void prepareEmbeds(ObjectNode body, boolean evenIfEmpty) {
+        if (embeds.size() != 0 || evenIfEmpty) {
+            ArrayNode embedsNode = JsonNodeFactory.instance.objectNode().arrayNode();
+            for (int i = 0; i < embeds.size() && i < 10; i++) {
+                embedsNode.add(((EmbedBuilderDelegateImpl) embeds.get(i).getDelegate()).toJsonNode());
+            }
+            body.set("embeds", embedsNode);
+        }
+    }
```
KILLEliteMaste, line 608: Should be moved to the other `prepare*` methods below

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):

PR #812, opened 2021-07-11
Add addIntents and setNonPrivilegedIntentsAnd method to DiscordApiBuilder

ShindouMihou:
## Description
This adds the two methods: `addIntents(...)` which adds the specified intents to the current list (or default) and also `setNonPrivilegedIntentsAnd(...)` which is basically setting the default intents together with the intents specified, though it's probably better to use `addIntents(...)` instead since it does the same thing (as long as you don't set anything else).

## Type of Change
- ✔️ New Feature

## Necessary Checks
- ✔️ Tested both methods with a test Discord bot
- ✔️ Codestyle

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
@@ -429,6 +429,29 @@ public class DiscordApiBuilder implements ChainableGloballyAttachableListenerMan
     }
 
     /**
+     * Sets all non privileged intents and the given intents.
+     *
+     * @param intentsToInclude One or more {@code Intent}s which should be included.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllNonPrivilegedIntentsAnd(Intent... intentsToInclude) {
+        setAllIntentsWhere(intent -> !intent.isPrivileged());
+        addIntents(intentsToInclude);
+        return this;
+    }
+
+    /**
+     * Adds the given intents to the already set.
+     *
+     * @param intents The intents to add.
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder addIntents(Intent... intents) {
+        delegate.addIntents(intents);
+        return this;
+    }
+
+    /**
      * Sets the intents where the given predicate matches.
      *
      * @param condition Whether the intent should be added or not.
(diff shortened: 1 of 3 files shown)


ShindouMihou:
Fixed conflicting merges and rebased successfully.

javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
```diff
@@ -428,6 +428,29 @@ public DiscordApiBuilder setAllNonPrivilegedIntentsExcept(Intent... intentsToOmi
...
+     * @return The current instance in order to chain call methods.
+     */
+    public DiscordApiBuilder setAllNonPrivilegedIntentsAnd(Intent... intentsToInclude) {
+        setAllIntentsWhere(intent -> !intent.isPrivileged());
+        addIntents(intentsToInclude);
+        return this;
+    }
+
+    /**
+     * Adds the intents specified.
```
KILLEliteMaste, line 444: Change to `Adds the given intents to the already set`

javacord-api/src/main/java/org/javacord/api/internal/DiscordApiBuilderDelegate.java
```diff
@@ -171,6 +171,14 @@
      */
     boolean isShutdownHookRegistrationEnabled();
 
+    /**
+     * Adds the intents specified to the current list of
+     * intents.
```
KILLEliteMaste, line 176: Change to `Adds the given intents to the already set`

javacord-core/src/main/java/org/javacord/core/DiscordApiBuilderDelegateImpl.java
```diff
@@ -390,6 +390,13 @@ public boolean isShutdownHookRegistrationEnabled() {
         return registerShutdownHook;
     }
 
+    @Override
+    public void addIntents(Intent... intents) {
+        this.intents.addAll(Arrays.stream(intents)
+                .filter(intent -> !this.intents.contains(intent))
```
KILLEliteMaste, line 396: No need to filter as `intents` is a Set

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/DiscordApiBuilderDelegateImpl.java
ShindouMihou: Realized that when I first sent the PR but forgot to change. 😓

KILLEliteMaste (approved):

PR #816, opened 2021-07-16
Add option to control user cache

Bastian:
And make it disabled by default (breaking change!).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 org.gradle.caching = true
-version = 3.3.3-SNAPSHOT
+version = 3.4.0-SNAPSHOT
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
@@ -440,6 +440,28 @@ public class DiscordApiBuilder implements ChainableGloballyAttachableListenerMan
     }
 
     /**
+     * Sets whether or not the user cache should be enabled.
+     *
+     * <p>By default, the user cache is disabled.
+     *
+     * @param enabled Whether or not the user cache should be enabled.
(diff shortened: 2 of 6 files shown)


KILLEliteMaste:
Was the breaking change necessary? Now everyone who wants to have a cache with the GUILD_MEMBERS intent has to activate the cache so 2 steps are now necessary vs 1 when you only had to set the intent. I also think there are are many more users that would need the cache or just want to get everything working instead of users facing issues with an active cache.

Bastian:
> Was the breaking change necessary?

It was not necessary, but I believe that this is the more sensible default:
1. Discord is giving hints that a fully populated user cache might not be possible in the future (see https://github.com/discord/discord-api-docs/issues/2184#issuecomment-816353472)
2. The user cache has a very large performance (especially memory) impact. I don't think that many bots actually need the user cache (<10% but that's just a guess). Something that expensive should be disabled by default and only enabled when needed.

KILLEliteMaste:
With # 1 I can totally agree with you. But on the other side there are already quiet a few questions always coming up why i.e. `getMembers` or `getRoles` is not working. The aim should be to provide an easy to use library but including another barrier isn't that useful for going that way (atm). 
We maybe have to in addition to this PR add some exceptions to the mentioned methods(and all others) and throw error messages why the method isn't working. This also shouldn't be an issue since the error would be always thrown if you didn't do both steps to activate the cache. After that it can be up to the user if he wants to activate it or remove the method instead.

Bastian:
> We maybe have to in addition to this PR add some exceptions to the mentioned methods(and all others) and throw error messages why the method isn't working.

Agreed. With a good error message, the user immediately knows how to fix the problem. Something like `"Cannot get members with a disabled user cache. You can enable it by calling DiscordApiBuilder#setUserCacheEnabled(true)."`

PR #818, opened 2021-07-21
Add method to manually request members + ServerMembersChunkEvent

Bastian:


javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -469,6 +470,13 @@ default String getDisplayName(User user) {
      */
     boolean hasAllMembersInCache();
 
+    /**
+     * Requests Discord to send the members for this server.
+     *
+     * <p>After calling this method, your bot will receive {@link ServerMembersChunkEvent}s.
+     */
+    void requestMembers();
```
KILLEliteMaste, line 478: I think this shouldn't be called `requestMembers` as this would probably be more appropiate for this endpoint https://discord.com/developers/docs/resources/guild#list-guild-members . You could rename it to i.e. `requestMemberChunks` or `initiateMemberChunksRequest`.

javacord-api/src/main/java/org/javacord/api/event/server/member/ServerMembersChunkEvent.java
```diff
@@ -0,0 +1,19 @@
...
+ * A server members chunk event.
+ */
+public interface ServerMembersChunkEvent extends ServerEvent {
+
+    /**
+     * Gets the chunk of users.
+     *
+     * @return The chunk of users.
+     */
+    List<User> getMembersChunk();
```
KILLEliteMaste, line 18: Although `getMembersChunk` is a bit more descriptive(and maybe also a bit wrong as you dont get a chunk but the members out of it) where you get the members from I think `getMembers` is sufficient as it should be clear that you get the members from the chunk event

javacord-core/src/main/java/org/javacord/core/event/server/member/ServerMembersChunkEventImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * The implementation of {@link ServerMembersChunkEvent}.
+ */
+public class ServerMembersChunkEventImpl extends ServerEventImpl implements ServerMembersChunkEvent {
+
+    private final List<User> membersChunk;
+
+    /**
+     * Creates a new server member join event.
```
KILLEliteMaste, line 19: Wrong Javadoc

javacord-core/src/main/java/org/javacord/core/util/handler/guild/GuildMembersChunkHandler.java
```diff
@@ -22,8 +27,16 @@ public GuildMembersChunkHandler(DiscordApi api) {
...
+                .map(server -> (ServerImpl) server)
+                .ifPresent(server -> {
+                    List<Member> members = server.addAndGetMembers(packet.get("members"));
+                    ServerMembersChunkEventImpl event = new ServerMembersChunkEventImpl(
+                            server,
+                            members.stream().map(Member::getUser).collect(Collectors.toList())
+                    );
+                    api.getEventDispatcher().dispatchServerMembersChunkEvent(server, server, event);
+                });
+
```
KILLEliteMaste, line 39: Remove line

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
AyuAi: I think `requestMembersChunk` makes the most sense, since `request` implies a REST request, and the data being acquired is a Member Chunk

javacord-core/src/main/java/org/javacord/core/event/server/member/ServerMembersChunkEventImpl.java
AyuAi:
```suggestion
     * Creates a new server members chunk event.
```

AyuAi (approved):
Mostly reviewing for extra visibility, as I now have a need for this feature; Would be nice to have it staged sooner rather than later :D (And huge ty for even making this PR for when I needed to make my own Member Cache <3)

PR looks good to me otherwise. I generally agree with @KILLEliteMaste's comments, but they are simply semantical tbf

Milo123459:
Any updates?

PR #821, opened 2021-07-23
Fixed inverted logic of User#getMutualServers

KILLEliteMaste:


javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java
```diff
@@ -241,15 +240,15 @@ public boolean hasDefaultAvatar() {
...
-            }
-            return servers;
+            return api.getEntityCache().get().getMemberCache().getServers(getId());
+        }
+        Set<Server> servers = new HashSet<>();
+        if (member != null) {
+            servers.add(member.getServer());
         }
-        return api.getEntityCache().get().getMemberCache().getServers(getId());
+        return servers;
```
coaxial1, line 251: I think we should return an immutable collection here, as I think the returned collection in case of an enabled cache is immutable as well.

Probably just like this
```java
return member == null ? Set.of() : Set.of(member.getServer());
```
What do you think?

coaxial1 (changes requested):
Just one minor thought :)

javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java
coaxial1: Damn, we have to support Java 8. Then let's use
```java
return member == null ? Collections.emptySet() : Collections.singleton(member.getServer());
```

KILLEliteMaste: Yeah, I agree

coaxial1 (approved):

PR #830, opened 2021-08-02
Expose method to get unknown custom emoji

coaxial1:
This will expose ``getKnownCustomEmojiOrCreateCustomEmoji`` to the public API of ``DiscordApi`` so one can get custom emoji objects that are not known to the current shard for use elsewhere in the Javacord API.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -1223,6 +1223,20 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
     }
 
     /**
+     * Creates or gets a custom emoji to be used in other Javacord APIs. Use this if the custom emoji you're looking
+     * for is hosted on a different shard, and can't be accessed through {@code getCustomEmojiById()}.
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -1222,6 +1222,19 @@ default String makeMentionsReadable(String content) {
                         .collect(Collectors.toList()));
     }
 
+    /**
+     * Creates or gets a custom emoji to be used in other Javacord APIs. Use this if the custom emoji you're looking
+     * for is hosted on a different shard, and can't be accessed through {@code getCustomEmojiById()}.
+     * If the custom emoji is known, the method will return the known custom emoji instead of creating a new one.
+     * The method will always return a non-null value, even if the emoji does not really exist.
```
KILLEliteMaste, line 1229: `...even if the emoji does not exist which will lead to a non functional custom emoji for further usage.`

KILLEliteMaste (changes requested):
You should add `@Override` to the DiscordApiImpl implementation of this method

KILLEliteMaste (approved):

PR #834, opened 2021-08-03
FIX: Separate Heartbeat Scheduler

### LINKED ISSUE #833: Heartbeats not answered during heavy load
Unfortunately, due to heavy load at times on the central scheduler, heartbeats are regularly missed at points of stress in Beemo (and thus the bot cannot stay connected to the gateway).

While it can be argued that this issue only exists because of the scheduler being used outside of Javacord, the threadpool is purposefully made accessible and it is commonly mentioned to use it for the sake of a cleaner shutdown / less responsibility in managing your threads if Javacord makes the threadpools / executors for you.

AyuAi:
Fixes #833 

This has been a bit of an ongoing discussion, as Beemo has been missing heartbeats during times of high-stress (such as login, queueing a bunch of REST calls, etc). While there may be a counterargument here that the Javacord schedulers should not be used for computations, it has definitely been a regular suggestion in the Javacord Discord to use the Javacord threadpools, and it doesn't say anywhere in the code that it is doing essential things like heartbeats (or else I doubt anyone would touch them).

This PR is an example of how I imagine a heartbeat scheduler would look in the current Javacord. While it is accessible to the wrapper user, it is properly labelled as the `Heart Scheduler` (notably, not `Heartbeat Scheduler`, because the class in Javacord for heartbeats is cutely named `Heart`!), so I highly doubt that anyone would make the mistake of using it for heavy computations, if even at all.

Note: While I don't think a heartbeat scheduler should be accessible to users, because of the way Javacord is currently designed around `ThreadPool.java` being accessible I don't think it would be a quick and easy PR to make otherwise.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/util/concurrent/ThreadPool.java
+++ b/javacord-api/src/main/java/org/javacord/api/util/concurrent/ThreadPool.java
@@ -24,6 +24,14 @@ public interface ThreadPool {
     ScheduledExecutorService getScheduler();
 
     /**
+     * Gets the used heart scheduler. 
+     * This scheduler is used to keep the bot connected to the gateway.
+     *
+     * @return The used heart scheduler.
+     */
+    ScheduledExecutorService getHeartScheduler();
+
+    /**
      * Gets the used daemon scheduler.
      *
      * @return The used daemon scheduler.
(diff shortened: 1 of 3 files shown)


Vampire:
Sorry, but I'm against this change and as explained in detail in the according issue I highly doubt it would have any effect on the actual problem in that issue. If the scheduler would be blocked, then those message would even be less likely displayed, while of course the heartbeats would still be not sent if the scheduler is blocked.

> While there may be a counterargument here that the Javacord schedulers should not be used for computations, it has definitely been a regular suggestion in the Javacord Discord to use the Javacord threadpools

This is the same as if you do blocking or long-running things in a listener thread, or in a `thenAccept` that then blocks REST sending, and so on.

And it was never suggested or recommended to do blocking or long-running things on the **scheduler**, but only on the **executor**. The executor is unbounded and just creates new threads if necessary to do the work that needs to be done. The scheduler is single-threaded and should exclusively used for scheduling. If you need to schedule a long-running or blocking task, you use the scheduler for the scheduling and then when the task triggers use the executor to execute the blocking or long-running stuff.

Just like event dispatching is blocked if you do long-running or blocking things in the wrong place or REST sending is blocked if you do long-running or blocking things in the wrong place, scheduling is blocked if you do long-running or blocking things in the wrong place.

Imho there should not be another scheduler added just to work-around buggy user code.
I think you need veeeeery many **non-blocking** tasks to cause any harm with the scheduler.

AyuAi:
Tested and proved no difference on Beemo :/

Issue #732, opened 2021-04-04
MessageCreateEvent does not trigger upon messages being sent in DMs


elainecao93:
Breaking bug in 3.2.0.

Using the example bot in the `readme.md`:

Expected behavior: When the bot is interacted with in DMs, the bot replies in DMs.

Actual behavior: The bot does not reply because the `MessageCreateEvent` does not trigger.

Vampire:
Duplicate of #707

Bastian:
I've just released version 3.3.0 that fixes this bug.

PR #835, opened 2021-08-04
Fix incoming interaction in unknown DM channels

coaxial1:
Similar to past issues with DMs ( #524, #707, #732, #708 ), interaction don't work if they are spawned in an uncached DM channel.

Currently, this spawns this exception:
```
java.lang.NullPointerException: Cannot invoke "org.javacord.api.entity.channel.TextChannel.getMessageCache()" because "channel" is null
    at org.javacord.core.entity.message.MessageImpl.<init>(MessageImpl.java:178) ~[javacord-3.4.0-SNAPSHOT-shaded.jar:na]
    at org.javacord.core.interaction.MessageComponentInteractionImpl.<init>(MessageComponentInteractionImpl.java:46) ~[javacord-3.4.0-SNAPSHOT-shaded.jar:na]
    at org.javacord.core.interaction.ButtonInteractionImpl.<init>(ButtonInteractionImpl.java:18) ~[javacord-3.4.0-SNAPSHOT-shaded.jar:na]
```

Steps to reproduce:
- Start test bot
- Send yourself a DM from the bot with a button component
- Restart bot
- Click button
--> The interaction cannot be processed

This PR contains the fix for this issue similar to how we fixed the other related issues.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
@@ -11,7 +11,10 @@ import org.javacord.api.event.interaction.MessageComponentCreateEvent;
 import org.javacord.api.event.interaction.SelectMenuChooseEvent;
 import org.javacord.api.event.interaction.SlashCommandCreateEvent;
 import org.javacord.api.interaction.InteractionType;
+import org.javacord.core.entity.channel.PrivateChannelImpl;
 import org.javacord.core.entity.server.ServerImpl;
+import org.javacord.core.entity.user.MemberImpl;
+import org.javacord.core.entity.user.UserImpl;
 import org.javacord.core.event.interaction.ButtonClickEventImpl;
 import org.javacord.core.event.interaction.InteractionCreateEventImpl;
 import org.javacord.core.event.interaction.MessageComponentCreateEventImpl;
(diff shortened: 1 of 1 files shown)


javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
```diff
@@ -44,7 +47,15 @@ public InteractionCreateHandler(DiscordApi api) {
...
-            channel = api.getTextChannelById(packet.get("channel_id").asLong()).orElse(null);
+            long channelId = packet.get("channel_id").asLong();
+
+            // Check if this interaction comes from a guild or a DM
+            if (packet.hasNonNull("guild_id")) {
+                channel = api.getTextChannelById(channelId).orElse(null);
+            } else {
+                UserImpl user = new UserImpl(api, packet.get("user"), (MemberImpl) null, null);
+                channel = PrivateChannelImpl.getOrCreatePrivateChannel(api, channelId, user.getId(), user);
+            }
```
KILLEliteMaste, line 58: I think this has the same issue as we had with EPHEMERAL messages. The packet should not include a guild id but was sent from a guild and we handle it as a private channel. Can you confirm this?

coaxial1: This is an interaction generated from an EPHEMERAL message that has been sent in a guild. Seems like it contains both "member" and "guild_id", so we should be fine.

```json
{
  "t": "INTERACTION_CREATE",
  "s": 8,
  "op": 0,
  "d": {
    "version": 1,
    "type": 3,
    "token": "aW50ZXJhY3Rpb24...YwcHdiNGtZTGE3T1JodVRy",
    "message": {
      "id": "872505001019338812",
      "flags": 64
    },
    "member": {
      "user": {
        "username": "Bad_Mate | Pat",
        "public_flags": 393280,
        "id": "386953515873992707",
        "discriminator": "5128",
        "avatar": "d28621be5548f02435d67c5151f15c84"
      },
      "roles": [
        "794275019416928307",
        "578903831908450308",
        "790708226235564044"
      ],
      "premium_since": null,
      "permissions": "274877906943",
      "pending": false,
      "nick": null,
      "mute": false,
      "joined_at": "2019-04-14T14:53:55.669000+00:00",
      "is_pending": false,
      "deaf": false,
      "avatar": null
    },
    "id": "872505008044785734",
    "guild_id": "566999586032320512",
    "data": {
      "custom_id": "Test2",
      "component_type": 2
    },
    "channel_id": "769960629888352286",
    "application_id": "597448122016202793"
  }
}
```

coaxial1: All that's missing is the details about the message, but this is expected, since Discord doesn't store the contents of ephemeral messages. At least we get the ID.

KILLEliteMaste:
LGTM! You just need to rebase on the latest version :)

PR #837, opened 2021-08-06
MessageBuilder and Component fixes

KILLEliteMaste:
- Fixed MessageBuilder logic that wasn't correctly adjusted for messages with multiple embeds
- Fixed hard coded actionrow size limit and `prepareComponents` in the MessageBuilder can now also add other HighLevelComponents(for the future)

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
@@ -27,7 +27,7 @@ import org.javacord.api.entity.message.mention.AllowedMentions;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.entity.webhook.IncomingWebhook;
 import org.javacord.core.DiscordApiImpl;
-import org.javacord.core.entity.message.component.ActionRowImpl;
+import org.javacord.core.entity.message.component.ComponentImpl;
 import org.javacord.core.entity.message.embed.EmbedBuilderDelegateImpl;
 import org.javacord.core.entity.message.mention.AllowedMentionsImpl;
 import org.javacord.core.entity.user.Member;
@@ -712,10 +712,8 @@ public class MessageBuilderBaseDelegateImpl implements MessageBuilderBaseDelegat
     protected void prepareComponents(ObjectNode body, boolean evenIfEmpty) {
         if (evenIfEmpty || !components.isEmpty()) {
             ArrayNode componentsNode = JsonNodeFactory.instance.objectNode().arrayNode();
-            for (int i = 0; i < components.size() && i < 5; i++) {
-                ActionRowImpl component = (ActionRowImpl) components.get(i);
-                componentsNode.add(component.toJsonNode());
-            }
+            components.forEach(
+                    highLevelComponent -> componentsNode.add(((ComponentImpl) highLevelComponent).toJsonNode()));
             body.set("components", componentsNode);
         }
     }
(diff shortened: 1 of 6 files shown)


javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -477,12 +475,12 @@ public StringBuilder getStringBuilder() {
...
      * Send a message to an incoming webhook.
      *
-     * @param webhookId    The id of the webhook to send the message to
+     * @param webhookId The id of the webhook to send the message to
      * @param webhookToken The token of the webhook to send the message to
-     * @param displayName  The display name the webhook should use
-     * @param avatarUrl    The avatar the webhook should use
-     * @param wait         If the completable future will be completed
-     * @param api          The api instance needed to send and return the message
+     * @param displayName The display name the webhook should use
```
coaxial1, line 480: Is it correct to remove all the indentation from param descriptions here (and elsewhere)? It looks better with spacing imo, and my IDE (IntelliJ) is always putting them in place if they are missing.

KILLEliteMaste: I once got exactly the same but opposite question
image

Javacord doesn't have any guideline about this currently and I don't mind doing it either way.

coaxial1: Hm okay. What should we agree on? Otherwise we'll continue to randomly change this in PRs

coaxial1: Also, I couldn't find a checkstyle config option for this, sp I'm not sure if we can enforce it at all

KILLEliteMaste: That's the issue. Back then i also had a look on checkstyle if it was possible to enforce it but sadly there isn't anything. 
If you like it better to align the Javadoc at the same identation we can go for it. That's something that's easily changed with auto formatting the project if it's unwanted at some point.

coaxial1: I like it better actually, yeah. Increases readability a lot. :)

KILLEliteMaste: Do you want to create a follwoup PR for this?

Bastian: Here's some information from the Google style guide: https://google.github.io/styleguide/javaguide.html#s4.6.3-horizontal-alignment
The downside of this kind of alignment is, that it can result in untidy diffs because single-line changes could also require changes in other lines to adjust alignment.

But like I said previously: It doesn't matter that much. This whole topic is just bike-shedding. It would be nice to enforce a consistent style but if that's not possible, it also doesn't really matter.

coaxial1: I mean, maybe it would be "untidy", but that is okay if we're changing that actual method.
What I'd consider to be much worse is if two main contributors have different settings on that and always push completely unrelated changes because their auto-formatter's are fighting each other ... :D
So we should just agree on anything, and I'll change (or not) it on my side as well

KILLEliteMaste: @coaxial1 is there anything else or does this look good to you so I it can get merged?

coaxial1 (approved):
LGTM!

PR #838, opened 2021-08-06
Allow ephemeral deferred interaction responses

coaxial1:
Deferred responses to interactions (type 5) can be ephemeral as well. This PR adds the ability to make the "XYZBot is thinking ... " loading state ephemeral.

I've chosen to overload the method to make this change not breaking.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
@@ -48,6 +48,18 @@ public interface InteractionBase extends DiscordEntity {
     CompletableFuture<InteractionOriginalResponseUpdater> respondLater();
 
     /**
+     * Sends an acknowledgement of the interaction to Discord and displays an (ephemeral) loading state to the user,
+     * indicating that you'll respond with a delay.
+     * Please note that you can only actually update this loading state message within 15 minutes after receiving the
+     * interaction.
+     *
+     * @param ephemeral wether or not the response should be ephemeral
+     * @return A CompletableFuture that completes as soon as the acknowledgement has been sent; it yields an updater
+     *     that should be used to update the message later on.
+     */
+    CompletableFuture<InteractionOriginalResponseUpdater> respondLater(boolean ephemeral);
+
+    /**
      * Create a message builder to send follow up messages for this interaction.
      * You can send, edit and delete follow up messages up to 15 minutes after you received the interaction.
      *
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
```diff
@@ -47,6 +47,18 @@
...
+     * Sends an acknowledgement of the interaction to Discord and displays an (ephemeral) loading state to the user,
+     * indicating that you'll respond with a delay.
+     * Please note that you can only actually update this loading state message within 15 minutes after receiving the
+     * interaction.
+     *
+     * @param ephemeral wether or not the response should be ephemeral
+     * @return A CompletableFuture that completes as soon as the acknowledgement has been sent; it yields an updater
+     *     that should be used to update the message later on.
+     */
+    CompletableFuture<InteractionOriginalResponseUpdater> respondLater(boolean ephemeral);
```
KILLEliteMaste, line 60: I would change the name to `respondLaterEphemeral` and omit the parameter as this would be cleaner

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
coaxial1: Actually I had this at first, but changed it to my version after thinking about it.

Doing it like I did allows for easy programmatic decisions about the ephemeral state which would be complicated with two fixed methods (like you suggested):

```java
boolean shouldBeEphemeral = ...; // Some kind of automated decision that is not always the same
Interaction interaction = ...;

(shouldBeEphemeral ? interaction.respondLaterEphemeral() : interaction.respondLater()).thenAcceptAsync(updater -> ...);
// vs my version
interaction.respondLater(shouldBeEphemeral).thenAcceptAsync(updater -> ...);
```

This looks weird, and is not necessary with an boolean parameter, whereas it doesn't make much of a difference to write these either one of these lines if you always want an ephemeral response:
```java
interaction.respondLater(true).thenAccept(updater -> ...);
// vs
interaction.respondLaterEphemeral().thenAccept(updater -> ...);
```
Keep in mind that both will be equally readable in a modern IDE due to parameter hints:
grafik

KILLEliteMaste: Okay, I can agree with you. LGTM!

KILLEliteMaste (approved):

PR #842, opened 2021-08-08
Fix message updater

coaxial1:
Fixing two bugs in the message updater:
- Inverted behavior of "applyChanges" and "replaceMessage"
- Returned message was outdated / taken from cache; now uses data from the Discord response

Adds ``Message#createUpdater()`` as convenience method.


KILLEliteMaste (approved):
LGTM!

PR #843, opened 2021-08-08
Elevate getAvatarHash to User

AyuAi:
This PR is to make the `UserImpl#getAvatarHash` from the `javacord-core` accessible via the `User` class.

This is important, at least for me (and anyone else who wants to manage their own cache), so that the raw hash can be easily accessible / stored on the Javacord user's end. I am actually currently running on a patch myself (https://github.com/AyuAi/Javacord/commit/55e8ecb8d70d9469ff26329a5afff32b028be4a0) that elevated this method because I needed access to it to implement my own member/user cache on top of Javacord.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/User.java
@@ -204,6 +204,13 @@ public interface User extends DiscordEntity, Messageable, Nameable, Mentionable,
     EnumSet<UserFlag> getUserFlags();
 
     /**
+     * Gets the avatar hash of the user.
+     *
+     * @return The avatar hash.
+     */
+    Optional<String> getAvatarHash();
+
+    /**
      * Gets the avatar of the user.
      *
      * @return The avatar of the user.
--- a/javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java
@@ -166,13 +166,9 @@ public class UserImpl implements User, InternalUserAttachableListenerManager {
         return Optional.ofNullable(member);
     }
 
-    /**
-     * Gets the avatar hash of the user.
-     *
-     * @return The avatar hash.
-     */
-    public String getAvatarHash() {
-        return avatarHash;
+    @Override
+    public Optional<String> getAvatarHash() {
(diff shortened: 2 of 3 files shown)


javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -203,6 +203,13 @@ default UserStatus getWebStatus() {
      */
     EnumSet<UserFlag> getUserFlags();
 
+    /**
+     * Gets the avatar hash of the user.
+     *
+     * @return The avatar hash.
+     */
+    public String getAvatarHash();
+
```
coaxial1, line 212: If you're converting this to a public API, this must return an ``Optional<String>``, as users without avatar won't have an avatar hash.

coaxial1 (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
AyuAi: Ah, you are right. That's a good point

AyuAi: https://github.com/Javacord/Javacord/commit/45ebb0437b87d93eaebddddc94cb7b1e88c0fc84

coaxial1:
This won't compile.
Also, make sure to check all locations where the original ``getAvatarHash`` method is used, and adjust all affected code accordingly to deal with the new optional return type. :)

AyuAi:
> This won't compile.
> Also, make sure to check all locations where the original `getAvatarHash` method is used, and adjust all affected code accordingly to deal with the new optional return type. :)

Aw heck, you right. Idk why I thought it was a quick change lol

AyuAi:
Unrelated tangent, but I hate how VSCode wont import Javacord as a Java project.. 🙃

coaxial1:
That's why you'd usually use IntelliJ for Java development xD

coaxial1 (approved):
LGTM!

AyuAi:
> That's why you'd usually use IntelliJ for Java development xD

Lame

coaxial1:
Hey @AyuAi if you rebase your PR onto latest dev, we can merge it :)

AyuAi:
> Hey @AyuAi if you rebase your PR onto latest dev, we can merge it :)

Rebased! c:

PR #845, opened 2021-08-11
Allow the bot to bypass disabled user cache.

ShindouMihou:
Personally, I don't know whether this is really needed but for many bots, the disabled user cache breaks methods like `canYouManageServer` or other similar methods for the bot. I personally needed those methods to function which is why I monkey-patched this, some bots may also need the bot to be on the cache, so I decided to make a PR.

If you think that this PR or the intention of the PR doesn't really fit well with Javacord, feel free to close this PR.

Checks:
- ✔️ Codestyle.

Tests:
- ✔️ Actual production deployment on a bot.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
@@ -943,9 +943,10 @@ public class DiscordApiImpl implements DiscordApi, DispatchQueueSelector {
      * @param member The member to add.
      */
     public void addMemberToCacheOrReplaceExisting(Member member) {
-        if (!isUserCacheEnabled()) {
+        if (!isUserCacheEnabled() && member.getId() != you.getId()) {
             return;
         }
+        
         entityCache.getAndUpdate(cache -> {
             Member oldMember = cache.getMemberCache()
                     .getMemberByIdAndServer(member.getId(), member.getServer().getId())


KILLEliteMaste:
I am not sure if this works correctly. What is your current setup? Is the GUILD_MEMBERS intent set? What do you have `setUserCacheEnabled` set to?

ShindouMihou:
> I am not sure if this works correctly. What is your current setup? Is the GUILD_MEMBERS intent set? What do you have `setUserCacheEnabled` set to?

My setup on the bot right now is `GUILD_MEMBERS` intent enabled with `setUserCacheEnabled(false)` and it seems to work from the image below:
image

Here is the eval with me as the user:
image

KILLEliteMaste:
> Does it perhaps not work without the GUILD_MEMBERS intent because that would be weird as the last version without GUILD_MEMBERS, the you variable on DiscordApiImpl shouldn't be null?

I am not sure what exactly you mean, but `you` isn't null because we get it from an initial packet which always contains the bot as a user.

Regardless of that I am not sure if this should be in Javacord as this adds some sort of semi filled cache behaviour we currently do not have I think. And this would most likely be superseeded by the cache module whhich could allow you to add specific ID's which should be cached all the time.

ShindouMihou:
> > Does it perhaps not work without the GUILD_MEMBERS intent because that would be weird as the last version without GUILD_MEMBERS, the you variable on DiscordApiImpl shouldn't be null?
> 
> I am not sure what exactly you mean, but `you` isn't null because we get it from an initial packet which always contains the bot as a user.
> 
> Regardless of that I am not sure if this should be in Javacord as this adds some sort of semi filled cache behaviour we currently do not have I think. And this would most likely be superseeded by the cache module whhich could allow you to add specific ID's which should be cached all the time.

I had the same thought as well whether this should be in Javacord or not since the cache module was indeed coming, I was also thinking about waiting for the cache module part which is still around the corner. I guess this PR can be closed for now. 🧇

PR #848, opened 2021-08-14
Move from deprecated "embed" field to "embeds" in UncachedMessageUtil.

KILLEliteMaste:
- Added MessageBuilderBase#addEmbeds(List<EmbedBuilder>)
- Switched the edit methods in Message from using the static edit methods to using the MessageUpdater.
- Message#edit, MessageEvent#editMessage and UncachedMessageUtil#edit methods now take in addition a list of embeds instead of a single embed and single embeds were enhanced to take at least 1 embeds.
- A few edit methods couldn't be enhanced due to the embed parameter being in the middle of params, so they introduce a breaking change as they have been refactored to take a list of embeds
- Messageable was enhanced with being able to take multiple embeds

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -44,9 +44,8 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Returns a {@code MessageBuilder} according to this {@code Message}.
      *
-     * @see MessageBuilder#fromMessage(Message)
-     *
      * @return The {@code MessageBuilder}.
+     * @see MessageBuilder#fromMessage(Message)
      */
     default MessageBuilder toMessageBuilder() {
         return MessageBuilder.fromMessage(this);
@@ -55,18 +54,17 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Returns a {@code WebhookMessageBuilder} according to this {@code Message}.
      *
-     * @see WebhookMessageBuilder#fromMessage(Message)
-     *
      * @return The {@code WebhookMessageBuilder}.
+     * @see WebhookMessageBuilder#fromMessage(Message)
      */
     default WebhookMessageBuilder toWebhookMessageBuilder() {
         return WebhookMessageBuilder.fromMessage(this);
     }
 
     /**
-     * Cross posts the message if it is in a announcement channel.
+     * Cross posts the message if it is in an announcement channel.
      *
-     * @param api The discord api instance.
+     * @param api       The discord api instance.
      * @param channelId The id of the message's channel.
      * @param messageId The id of the message.
      * @return The new message object.
@@ -76,7 +74,7 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     }
 
     /**
-     * Cross posts the message if it is in a announcement channel.
+     * Cross posts the message if it is in an announcement channel.
      *
      * @return The new message object.
      */
@@ -87,7 +85,7 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Deletes the message.
      *
-     * @param api The discord api instance.
+     * @param api       The discord api instance.
      * @param channelId The id of the message's channel.
      * @param messageId The id of the message.
      * @return A future to tell us if the deletion was successful.
@@ -99,7 +97,7 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Deletes the message.
      *
-     * @param api The discord api instance.
+     * @param api       The discord api instance.
      * @param channelId The id of the message's channel.
      * @param messageId The id of the message.
      * @return A future to tell us if the deletion was successful.
@@ -111,10 +109,10 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Deletes the message.
      *
-     * @param api The discord api instance.
+     * @param api       The discord api instance.
      * @param channelId The id of the message's channel.
      * @param messageId The id of the message.
-     * @param reason The audit log reason for the deletion.
+     * @param reason    The audit log reason for the deletion.
      * @return A future to tell us if the deletion was successful.
      */
     static CompletableFuture<Void> delete(DiscordApi api, long channelId, long messageId, String reason) {
@@ -124,10 +122,10 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     /**
      * Deletes the message.
      *
-     * @param api The discord api instance.
+     * @param api       The discord api instance.
      * @param channelId The id of the message's channel.
      * @param messageId The id of the message.
-     * @param reason The audit log reason for the deletion.
+     * @param reason    The audit log reason for the deletion.
      * @return A future to tell us if the deletion was successful.
      */
     static CompletableFuture<Void> delete(DiscordApi api, String channelId, String messageId, String reason) {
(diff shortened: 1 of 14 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -334,28 +335,28 @@ default WebhookMessageBuilder toWebhookMessageBuilder() {
...
     /**
      * Updates the embed of the message.
      *
-     * @param embed The new embed of the message.
+     * @param embeds An array of the new embeds of the message.
      * @return A future to check if the update was successful.
      */
-    default CompletableFuture<Message> edit(EmbedBuilder embed) {
-        return Message.edit(getApi(), getChannel().getId(), getId(), null, false, embed, true);
+    default CompletableFuture<Message> edit(List<EmbedBuilder> embeds) {
```
ShindouMihou, line 347: Can't we allow single embeds as well by simply using `Collections#singletonList` since changing this method up will probably break a pile load of stuff. I don't really see why adding a convenience method to prevent this from being a breaking is not possible?

KILLEliteMaste: It is of course possible but I've chosen to go this way as it is probably better to have 1 way of doing it than multiple and otherwise we would have to overload the method so many times you cannot really keep track of them anymore. Furthermore I think it is also more confusion to the users as they differentiate just for 1 argument.
If the demand is really that high for a single embed param we could add the methods back but I am against it.

ShindouMihou (approved):

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
coaxial1: Why don't we use ``EmbedBuilder... embeds`` instead to keep it non-breaking?

KILLEliteMaste: We could but is it really such a big deal? This PR will stay a breaking change even if we decide to change it because of
```java
static CompletableFuture<Message> edit(DiscordApi api, long channelId, long messageId, String content,
                                        boolean updateContent, EmbedBuilder embed, boolean updateEmbed)
```
which would need to be adjusted.
It may also be a personal preference but I think having a list as param is also more common to use instead of an array. And if I had to add multiple embeds dynamicaly I would have convert my embed list to an array which would look kinda weird `List.of(embed1, embed2).toArray(new EmbedBuilder[0])` and further more a list provides just a lot more flexibility.
This PR is also aimed for version 3.5 where a lot of breaking changes will be included so I don't see why we wouldn't want to improve these methods right away.

coaxial1: For example, I use the single embed method you're changing up here like at least 70 times and I'd need to refactor everything to ``message.edit(List.of(embed));`` which looks 100% weird as well.

I don't see why it's not a good idea to just offer both, if both is needed by different users with different requirements. We offer both in the MessageBuilder, so why not here?

I think it's totally fine to switch to a list for the complex edit() method you mentioned in your comment above as the list is in the middle of the parameter list (is anyone really using that method with all the flags? Probably an edge case anyways), but I really don't see why we should make the use of ``Message#edit`` less convenient than before. I want to edit my embed-only message with another embed. This was easy before, now I randomly have to wrap it in a list. This is not an improvement.

coaxial1: > having a list as param is also more common to use instead of an array

Yeah I mean the method gets served an array, but this is a variadic paraemter an as such a very common thing. As a developer I can easily pass in two, or three, or more embed builders without the need to use any sort of collection.
You are right that ``edit(EmbedBuilder[] embeds)`` would be a weird and uncommon thing.

KILLEliteMaste: I can see your point that you not necessarily want to always use `List.of()` if you only want to change 1 embed. But I also had the code maintainability in mind. So should we really offer both variants for each method? That would be 7 new methods. Or only for Message#edit(EmbedBuilder...) and #edit(String, EmbedBuilder...) as these are probably the most common used ones?

coaxial1: I think we should offer flexibility where possible. Look at all the convenience methods we have for components. :D

Why are 7 methods a problem if they are directly next to their related one with the list parameter, respectively, and the similarity is obvious to anyone who needs to change these methods in the future? :) 
In comparision, we have 24 methods just to create a button and I still believe we'll be able to maintain that.

I think there's no good reason to remove fully working, intuitive and easy APIs that have been present since the beginning (of v3, at least).
It's 7 quick methods for us, but potentially thousands of changes that need to be done in all the Javacord projects out there, as editing a message is likely one of the most common things to do with a bot. ^_^

coaxial1: For those methods where the list is in the middle of the param list, we could also still offer an overload with a singular instance like suggested by ShindouMihou above. I'd vote for this for the same reasons (offer flexibility, be easy to use). But since you don't like it, it's probably up to the others.

Side note: ``org.javacord.api.entity.message.Messageable.java`` probably deserves a similar refactoring :D

KILLEliteMaste: @coaxial1 I have reworked everything, readded and enhanced the single embed method. I also added and enhanced some methods in Messageable as you mentioned.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -214,117 +212,191 @@ default WebhookMessageBuilder toWebhookMessageBuilder() {
...
+     * @param api       The discord api instance.
      * @param channelId The id of the message's channel.
      * @param messageId The id of the message.
-     * @param content The new content of the message.
+     * @param content   The new content of the message.
      * @return A future to check if the update was successful.
      */
     static CompletableFuture<Message> edit(DiscordApi api, long channelId, long messageId, String content) {
-        return api.getUncachedMessageUtil().edit(channelId, messageId, content, true, null, false);
+        return api.getUncachedMessageUtil().edit(channelId,messageId,content);
```
coaxial1, line 222: No spaces? Maybe just hit the auto formatter on this file again :D

javacord-api/src/main/java/org/javacord/api/event/message/MessageEvent.java
```diff
@@ -55,19 +56,19 @@
     /**
      * Updates the embed of the message involved in the event.
      *
-     * @param embed The new embed of the message.
+     * @param embeds An array of the new embeds of the message.
      * @return A future to check if the update was successful.
      */
-    CompletableFuture<Message> editMessage(EmbedBuilder embed);
+    CompletableFuture<Message> editMessage(List<EmbedBuilder> embeds);
```
coaxial1, line 62: I think we should offer the ``Embed embed, Embed... embeds`` version you have everywhere here as well in addition to the list based methods.
Also, the javadocs here are still saying "the embed" :)

javacord-api/src/main/java/org/javacord/api/event/message/MessageEvent.java
```diff
@@ -55,19 +56,19 @@
...
     /**
      * Updates the content and the embed of the message involved in the event.
      *
      * @param content The new content of the message.
-     * @param embed The new embed of the message.
+     * @param embeds An array of the new embeds of the message.
      * @return A future to check if the update was successful.
      */
-    CompletableFuture<Message> editMessage(String content, EmbedBuilder embed);
+    CompletableFuture<Message> editMessage(String content, List<EmbedBuilder> embeds);
```
coaxial1, line 71: Here as well.

javacord-api/src/main/java/org/javacord/api/interaction/callback/InteractionMessageBuilder.java
```diff
@@ -176,7 +176,7 @@ public InteractionMessageBuilder addEmbed(EmbedBuilder embed) {
 
     @Override
     public InteractionMessageBuilder addEmbeds(EmbedBuilder... embeds) {
```
coaxial1, line 178: Should we have a method accepting a list here as well?

coaxial1 (changes requested):
A few last thoughts :)

javacord-api/src/main/java/org/javacord/api/interaction/callback/InteractionMessageBuilder.java
KILLEliteMaste: Yes. I am adding the method but i think we would have to rework this anyway to use the MessageBuilderBase.

coaxial1 (approved):
LGTM!

PR #850, opened 2021-08-15
Add missing methods to get the actual component

KILLEliteMaste:


coaxial1:
Looks good, but is it required? We are already implementing ``Specializable`` so we have a ``as()`` method, don't we? :)

KILLEliteMaste:
> Looks good, but is it required? We are already implementing `Specializable` so we have a `as()` method, don't we? :)

I also added the Specializable in this PR. But I think adding the methods explicit is better as you know directly what you can get like in the MessageComponentInteraction, Specializable was just a thought of if we forget to add the corresponding methods for a new component.

coaxial1:
Okay, got it. LGTM!

coaxial1 (approved):

PR #852, opened 2021-08-19
Add SlashCommandOptionType NUMBER.

KILLEliteMaste:
This PR is a breaking change in terms of the `getStringValue` now doesn't return a result anymore if the option type is not a string.
In place comes `getStringRepresentationValue` which also returns the boolean which was missing previously.

- Reworked SlashCommandInteractionOptionImpl to check for the "type" key value instead of the datatype of the value node to uniquely identify the option.
- Add getStringRepresentationValue to clearly differentiate between getStringValue which is now only present when the option "type" key is a string.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
@@ -1,12 +1,15 @@
 package org.javacord.core.interaction;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.logging.log4j.Logger;
 import org.javacord.api.DiscordApi;
 import org.javacord.api.entity.Mentionable;
 import org.javacord.api.entity.channel.ServerChannel;
 import org.javacord.api.entity.permission.Role;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.interaction.SlashCommandInteractionOption;
+import org.javacord.api.interaction.SlashCommandOptionType;
+import org.javacord.core.util.logging.LoggerUtil;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -30,6 +33,8 @@ public class SlashCommandInteractionOptionImpl implements SlashCommandInteractio
 
     private final List<SlashCommandInteractionOption> options;
 
+    private static final Logger LOGGER = LoggerUtil.getLogger(SlashCommandInteractionOptionImpl.class);
+
     /**
      * Class constructor.
      *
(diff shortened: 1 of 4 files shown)


javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
```diff
@@ -35,22 +41,97 @@ public SlashCommandInteractionOptionImpl(DiscordApi api, JsonNode jsonData) {
...
             stringValue = valueNode.asText();
             intValue = null;
             booleanValue = null;
-        } else if (valueNode != null && valueNode.isInt()) {
-            intValue = valueNode.asInt();
+            userValue = null;
+            channelValue = null;
+            roleValue = null;
+            mentionableValue = null;
+            numberValue = null;
```
montlikadani, line 53: Redundant reference initialization, these are already initialized as `null` by the compiler.

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
```diff
@@ -35,22 +41,97 @@ public SlashCommandInteractionOptionImpl(DiscordApi api, JsonNode jsonData) {
         name = jsonData.get("name").asText();
         JsonNode valueNode = jsonData.get("value");
 
-        if (valueNode != null && valueNode.isTextual()) {
+        int type = jsonData.get("type").asInt();
+        if (type == 3) {
```
montlikadani, line 45: Consider to use `switch case` instead of if-else

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
KILLEliteMaste: When they are declared as final they need to be initialized or the compiler will throw an error

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
KILLEliteMaste: Why? Except having 1 more statement in each branch and the possibility of forgetting to add a `break` when adding a new type I don't see a benefit to switch over

Joshix-1: switch is faster (I think) and looks nicer

ShindouMihou: > switch is faster (I think) and looks nicer

"Faster" would be another case of pre-mature optimization as we are not dealing with large number of conditions and if we want to go for readability, we can assign temporary variables that are initialized as null then assign those variables with values accordingly before ending with the temporary variables' value being used as the final value (opinion).

An example which I don't think would be needed is:
```java
String tempS = null;
int tempN = null;

if (condition == 1) {
   tempS = condition.asString();
} 

if (condition == 2) {
    tempN = condition.asInt();
}

finalS = tempS;
finalN = tempN;
```

Joshix-1: https://github.com/Joshix-1/Javacord/commit/c6add6f79d899d7d84140683f10dc7243a8f8542
That looks really nice.

Bastian: It would be better to use the enum instead of [magic numbers](https://en.m.wikipedia.org/wiki/Magic_number_(programming)) though.

KILLEliteMaste: I have adjusted the code and it now includes your (@Bastian and @Joshix-1) suggestions.

coaxial1 (approved):
LGTM

PR #854, opened 2021-08-20
Fix typos and grammars

KILLEliteMaste:


javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -354,7 +354,7 @@ default String getDiscriminatedName() {
     /**
      * Mutes this user on the given server.
      *
-     * @param server The server to umute this user on.
+     * @param server The server to unmute this user on.
```
Saladoc, line 357: This contradicts the method.
```suggestion
     * @param server The server to mute this user on.
```

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -364,7 +364,7 @@ default String getDiscriminatedName() {
...
```
Saladoc, line 367: Same as above
```suggestion
     * @param server The server to mute this user on.
```

javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
```diff
@@ -53,15 +53,15 @@
      * Please note that you can only actually update this loading state message within 15 minutes after receiving the
      * interaction.
      *
-     * @param ephemeral wether or not the response should be ephemeral
+     * @param ephemeral weather or not the response should be ephemeral
```
Saladoc, line 56:
```suggestion
     * @param ephemeral Whether the response should be ephemeral
```

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandPermissions.java
```diff
@@ -20,18 +20,18 @@
     SlashCommandPermissionType getType();
 
     /**
-     * Whether this permission is enabled or disabled for the this slash command permissions.
+     * Whether this permission is enabled or disabled for the slash command permissions.
      *
-     * @return True if the command is enabled for this user or role. Otherwise false.
+     * @return True if the command is enabled for this user or role otherwise, false.
```
Saladoc, line 25: This looks wrong. It appears that the otherwise belongs to the left-hand side of the statement.

Saladoc (changes requested):
Thank you for the time and effort that went into this PR. I have marked a few points where I spotted small mistakes and things that I suspect might be small mistakes.

I'm also unsure about whether all the hyphenating you did is 100% correct, but since neither way looks particularly wrong today and readability is not inhibited, I didn't mark those.

javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
```diff
@@ -53,15 +53,15 @@
...
      * @return A CompletableFuture that completes as soon as the acknowledgement has been sent; it yields an updater
      *     that should be used to update the message later on.
      */
     CompletableFuture<InteractionOriginalResponseUpdater> respondLater(boolean ephemeral);
 
     /**
-     * Create a message builder to send follow up messages for this interaction.
-     * You can send, edit and delete follow up messages up to 15 minutes after you received the interaction.
+     * Create a message builder to send follow-up messages for this interaction.
+     * You can send, edit and delete follow-up messages to 15 minutes after you received the interaction.
```
Saladoc, line 64: Second 'up' was necessary.

```suggestion
     * You can send, edit and delete follow-up messages for up to 15 minutes after you received the interaction.
```

Saladoc (changes requested):

Saladoc (approved):

PR #855, opened 2021-08-20
Remove members from cache when the bot leaves the server

KILLEliteMaste:
Came across on this while adding the exceptions for member cache access if the cache is not enabled. 
If I am not totally wrong/blind we never remove members from the cache if the bot leaves the server. I tested my assumption by adding a scheduler which prints every 5 seconds the amount of all users + the amount users for the server
```
Executors.newSingleThreadScheduledExecutor().scheduleWithFixedDelay(() -> {
            System.out.println("Server: "+ getEntityCache().get().getMemberCache().getMembersByServer(SERVER_ID).size() +
            "\tAll: "+ getEntityCache().get().getMemberCache().getMembers().size());
        },5,5,TimeUnit.SECONDS);
```

Before and after kicking the test bot from the server it prints still the **same** amount of members for the server and total members.
After applying this patch I got this result:
```
Server: 2		All: 8 //this is the server to be kicked
Server: 0		All: 6
//After being kicked
Server: 0		All: 6
Server: 0		All: 6
```
There are each 2 results because I was running on 2 shards.


coaxial1:
Just asking to be safe: This will not affect users that are present on another server on the same shard as well, correct?

KILLEliteMaste:
> Just asking to be safe: This will not affect users that are present on another server on the same shard as well, correct?

Just a small update.
Now I am on a single shard:
Server: 2		All: 14		Cached Users: 7
Server: 0		All: 12		Cached Users: 5
Before and after kicking the bot from the server. So it actually removes the users too even when are still being used but this behaviour has always been in Javacord as the ServerMemberLeaveEvent uses the same method as I am after chaging the 1 line in this PR to call removeMember from the server instead on the api (which basically just in addition removes the member from the muted / deafened set in the server). 
The 2 cached users that are now missing is me and the bot. So I though calling things like `getMemberById` 
```
public Optional<User> getMemberById(long id) {
        return api.getEntityCache().get().getMemberCache()
                .getMemberByIdAndServer(id, getId())
                .map(Member::getUser);
    }
```
on a different server where I am on and the bot should actually return an empty optional as the user is now missing in the user cache but surprisingly `getMemberById().isPresent()` still returns true. So I think that might also be the cause why Javacord uses a lot more (~ double) the amount of RAM compared to other Libraries as it saves the user once in the user cache and once in the member as it seems

coaxial1:
> on a different server where I am on and the bot should actually return an empty optional as the user is now missing in the user cache

Well, that would be disastrous. That's basically what I've been asking: The library must not remove the user from the user cache if they are still present on another server. It should only remove the member from the member cache, but not touch the user in this case.
Other than that, it's a good idea to clear the *member* cache if the bot gets kicked but I'm surprised that this didn't happen already tbh, maybe we do oversee sth?

Additionally, I don't understand this:
https://github.com/Javacord/Javacord/blob/5b5cfa3e8eb7a484338e8fe9176c7a8c3fa2fdb6/javacord-core/src/main/java/org/javacord/core/util/cache/MemberCache.java#L68
Why is the user removed from the user cache here, even though we're only removing a member? What if the user is needed in a DM, or on antoher server?
Does every server have it's own member cache incl underlying user cache (doubt :D)?

Maybe @Bastian could help us thinking here?

Bastian:
> Why is the user removed from the user cache here, even though we're only removing a member? What if the user is needed in a DM, or on antoher server?
Does every server have it's own member cache incl underlying user cache (doubt :D)?

This indeed looks like a bug.

coaxial1:
That may explain some of the anomalies we've seen with missing users in the cache.

KILLEliteMaste:
>  The library must not remove the user from the user cache if they are still present on another server

As I said the user is still present in the member object as the optional `getMemberById()` which returns a user is **not** empty. The only 2 methods that require the usercache are `getCachedUsers` and `getCachedUserById` maybe that's why no one every noticed something.

coaxial1:
Yeah, the reference is still there, which mitigated most of the effect by this bug. But I doubt that this leads to double RAM usage - both the user cache and the member objects are just referencing the same user object in memory, so it won't consume more RAM

coaxial1:
However, getCachedUserById() is used by many internal methods, and I use it in my own code from time to time as well.
Seems like this needs to be fixed asap

KILLEliteMaste:
> Why is the user removed from the user cache here, even though we're only removing a member? What if the user is needed in a DM, or on antoher server?

I think it's okay to remove the user if theres no other server the user is in. In case of you need the user in DM's you should get the user back once he DM's the bot,

> But I doubt that this leads to double RAM usage

Yeah that was a wrong assumption by me. But first it looked like that ^^

PR #856, opened 2021-08-20
Change message methods that take embed as argument only to needing an array of embeds

KILLEliteMaste:
In #848 I introduced the new methods with needing at least 1 argument and and inifite amount of more embeds. This was changed to only needing an array of embeds due to method overloading which also take an inifnite amount of argument of a type, so it is not possible to pass 0 embeds, unless you pass an empty array


coaxial1 (approved):
LGTM!

PR #860, opened 2021-09-04
Fix message cache deadlock by using central lock

coaxial1:
We saw a few reports of deadlocks related to different lock being acquired in the central message map in DiscordApiImpl and the message caches of channels. This PR uses the same lock for all message cache related synchronization to avoid deadlocks.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
@@ -362,7 +362,12 @@ public class DiscordApiImpl implements DiscordApi, DispatchQueueSelector {
     /**
      * A map with all cached messages.
      */
-    private final Map<Long, WeakReference<Message>> messages = Collections.synchronizedMap(new ConcurrentHashMap<>());
+    private final Map<Long, WeakReference<Message>> messages = new ConcurrentHashMap<>();
+
+    /**
+     * A lock for accessing the message cache.
+     */
+    private final ReentrantLock messageCacheLock = new ReentrantLock();
 
     /**
      * A map to retrieve message IDs by the weak ref that point to the respective
@@ -401,17 +406,17 @@ public class DiscordApiImpl implements DiscordApi, DispatchQueueSelector {
      * Creates a new discord api instance that can be used for auto-ratelimited REST calls,
      * but does not connect to the Discord WebSocket.
      *
-     * @param token                         The token used to connect without any account type specific prefix.
-     * @param globalRatelimiter             The ratelimiter used for global ratelimits.
-     * @param gatewayIdentifyRatelimiter    The ratelimiter used to respect the 5-second gateway identify ratelimit.
-     * @param proxySelector                 The proxy selector which should be used to determine the proxies that
-     *                                      should be used
-     *                                      to connect to the Discord REST API and websocket.
-     * @param proxy                         The proxy which should be used to connect to the Discord REST API and
-     *                                      websocket.
-     * @param proxyAuthenticator            The authenticator that should be used to authenticate against proxies that
-     *                                      require it.
-     * @param trustAllCertificates          Whether to trust all SSL certificates.
+     * @param token                      The token used to connect without any account type specific prefix.
+     * @param globalRatelimiter          The ratelimiter used for global ratelimits.
+     * @param gatewayIdentifyRatelimiter The ratelimiter used to respect the 5-second gateway identify ratelimit.
+     * @param proxySelector              The proxy selector which should be used to determine the proxies that
+     *                                   should be used
+     *                                   to connect to the Discord REST API and websocket.
+     * @param proxy                      The proxy which should be used to connect to the Discord REST API and
+     *                                   websocket.
+     * @param proxyAuthenticator         The authenticator that should be used to authenticate against proxies that
+     *                                   require it.
+     * @param trustAllCertificates       Whether to trust all SSL certificates.
      */
     public DiscordApiImpl(String token, Ratelimiter globalRatelimiter, Ratelimiter gatewayIdentifyRatelimiter,
                           ProxySelector proxySelector, Proxy proxy, Authenticator proxyAuthenticator,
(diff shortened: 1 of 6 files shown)


AyuAi:
I think it generally LGTM; As an aside, kinda cool you can try/finally without a catch, I did not know that. Does it just run the finally after exceptions and do nothing with them? Or would this break the lock if an exception occurs 👀

ShindouMihou:
> I think it generally LGTM; As an aside, kinda cool you can try/finally without a catch, I did not know that. Does it just run the finally after exceptions and do nothing with them? Or would this break the lock if an exception occurs 👀

Try finally has been on Java for as long as I can remember though. It shouldn't catch the exceptions at all from what I can remember but my memory could be wrong.

coaxial1:
Yup. It allows exceptions to propagate normally, but still runs your finally block in any case to release the lock in any case :)

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -664,6 +669,7 @@ public DiscordApiImpl(
             // After minimum JDK 9 is required this can be switched to use a Cleaner
             getThreadPool().getScheduler().scheduleWithFixedDelay(() -> {
                 try {
+                    messageCacheLock.lock();
```
KILLEliteMaste, line 672: I am not commenting every single line but move everywhere the `lock` process before the try. The ReentrantLock Javadoc suggests it like that and even though it is unlikely that the lock will throw an error (and something would be very wrong in the code) because of 
`This lock supports a maximum of 2147483647 recursive locks by the same thread. Attempts to exceed this limit result in Error throws from locking methods` 
we should follow this practice as otherwise `unlock` would be called even though the lock wasn't acquired.

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
coaxial1: Yeah it prob makes sense to follow the suggested best practice

coaxial1:
I've applied the lock best practice as suggested.
Additionally, I changed a few lines where removing messages from the cache was very inconsistent (e.g. before, in the bulk delete handler, messages were removed from both the API and the channel cache, whereas it was only removed from the API but not from the channel cache in the regular delete handler, for no apparent reason).

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageDeleteHandler.java
```diff
@@ -45,6 +46,8 @@ public void handle(JsonNode packet) {
             MessageDeleteEvent event = new MessageDeleteEventImpl(api, messageId, channel);
 
             api.removeMessageFromCache(messageId);
+            api.getCachedMessageById(messageId)
+                    .ifPresent(((MessageCacheImpl) channel.getMessageCache())::removeMessage);
```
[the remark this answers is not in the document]
coaxial1: lol true

KILLEliteMaste (approved):

coaxial1:
I'll deploy this final state of the PR to Yunite tomorrow and see if any problems come in production.

Good news is that I'm running the initial version of this PR since today, and I haven't seen a single shard freeze so far (as opposed to the 3 years before xD). @AyuAi, you might want to test this PR on your bot as well and see if it fixes your issues

ShindouMihou:
Going to try this PR out (albeit, my bot is smaller than both of yours) 😆 hope that the auto restart does not trigger anymore~ thanks for the PR 🎖️

AyuAi:
> I'll deploy this final state of the PR to Yunite tomorrow and see if any problems come in production.
> 
> Good news is that I'm running the initial version of this PR since today, and I haven't seen a single shard freeze so far (as opposed to the 3 years before xD). @AyuAi, you might want to test this PR on your bot as well and see if it fixes your issues

@coaxial1 @ShindouMihou Sounds like a plan. I will give it a go when I get a new version out today-ish x)

coaxial1:
The PR is running fine on my system. Haven't seen any shard freezes and there doesn't seem to be any performance penalty coming with the central lock strategy. Any more results from you guys? :)

ShindouMihou:
> Going to try this PR out (albeit, my bot is smaller than both of yours) 😆 hope that the auto restart does not trigger anymore~ thanks for the PR 🎖️ 

Since twenty minutes after sending this message, no restarts so far and performance is more or less the same as far as I can see (on both runs and on the statistics). Looks to be good on my side.

@AyuAi results would be a good measure as well (if she has deployed a version with this PR).

AyuAi:
> > Going to try this PR out (albeit, my bot is smaller than both of yours) 😆 hope that the auto restart does not trigger anymore~ thanks for the PR 🎖️
> 
> Since twenty minutes after sending this message (the one quoted), no restarts so far and performance is more or less the same as far as I can see (on both runs and on the statistics). Looks to be good on my side.
> 
> @AyuAi results would be a good measure as well (if she has deployed a version with this PR).

I didn't get around to deploying this change yesterday, but I am fixing to get it out today with some other changes. Quite hyped if this is the fix 👉 👈

AyuAi:
Kk, I finally deployed Beemo onto this branch :D Will update on if any issues arise

AyuAi:
Hmm, I am still getting automated restarts for inactive message consumers. I don’t even use the JC threadpool for anything other than JC, either. I switched to my own implementation for my own stuff 👀

ShindouMihou:
> Hmm, I am still getting automated restarts for 240-second inactive message consumers. I don’t even use the JC threadpool for anything other than JC, either. I switched to my own implementation for my own stuff 👀

Can you try thread dumping with jstack? 🤔

AyuAi:
https://drive.google.com/file/d/1OVJlRp3UmJQgp6X-m792Z3Y7rC2JLEKR/view?usp=drivesdk

Running 14 total shards on one JVM atm

coaxial1:
Hm, if they restarted *before* you dumped it won't help; we'd need a dump while it's in the stuck state.
I've added [this](https://hastebin.com/axagucekaz.java) to my system to detect deadlocks during runtime (although I haven't seen any, applied it after this PR). 
You could try to see if you run into any.

But 240 seconds isn't much tbh - wouldn't it be totally possible that there has not been any message in this period?

@ShindouMihou How do you trigger auto-restarts? As well by looking at message listener activity?

AyuAi:
> Hm, if they restarted _before_ you dumped it won't help; we'd need a dump while it's in the stuck state.
> I've added [this](https://hastebin.com/axagucekaz.java) to my system to detect deadlocks during runtime (although I haven't seen any, applied it after this PR).
> You could try to see if you run into any.
> 
> But 240 seconds isn't much tbh - wouldn't it be totally possible that there has not been any message in this period?
> 
> @ShindouMihou How do you trigger auto-restarts? As well by looking at message listener activity?

I think it is not really plausible for this to happen, since there were periods where Beemo was much smaller and had a lot less messages/shard but still did not trigger such a mechanism.

ShindouMihou:
> iirc, @ShindouMihou probably watches message listener too, since she asked me what I did for Beemo and I showed her that code; I think she mentioned it before as well

Yeah, I was previously running on MCL monitor but added reaction add, remove and message edit and delete listeners since if a shard is frozen then those events would fail as well (and all four additional are related to message somehow). I only added them as an extra measure since my bot is smaller than both of yours 😆 so false positives could happen.

Also added an automatic thread dump (JSON format) for mine before it restarts, just in case: [Source](https://gist.github.com/ShindouMihou/1ae91efc7f1615c0423ee4a018ab924a)

coaxial1:
Are we merging this? @KILLEliteMaste

KILLEliteMaste:
Since there seems to be no issue with a central lock as you tested it, sure. You just need to rebase to the latest version

PR #864, opened 2021-09-11
Threads patch

PhoenixInvertigo:


javacord-core/src/main/java/org/javacord/core/entity/channel/ChannelThreadImpl.java
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+    The timestamp when the thread's archive status was last changed.
+     */
+    private final Instant archiveTimestamp;
+
+    /**
+     An indication of whether a user has joined a thread or not.
+     */
+    private final ThreadMember member;
```
ShindouMihou, line 66: Isn't a Javadoc comment supposed to be:
```java
/**
 * Comment here
 */
```

Instead of:
```java
/*
Comment
*/
```

javacord-core/src/main/java/org/javacord/core/entity/channel/ThreadMemberImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    public Long getId() {return id;}
+
+    @Override
+    public Long getUserId() {return userId;}
+
+    @Override
+    public Instant getJoinTimestamp() {return joinTimestamp;}
+
+    @Override
+    public int getFlags() {return flags;}
```
ShindouMihou, line 41: I don't think the code style checker encourages these where everything is in a single line.

This PR also failed the code style check, might have to fix that first.

ShindouMihou (commented):
For immediate return methods, please follow codestyle:
```java
public Class someReturnMethod() {
     return someReturnValue;
}
```

Instead of:
```java
public Class someReturnMethod() { return someReturnValue; }
```

Also, the other comments I left. Other than that, I'll leave it to the others.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelThread.java
```diff
@@ -0,0 +1,120 @@
...
+    long getOwnerId();
+
+    /**
+     * Gets the creator of the thread.
+     *
+     * @return The owner.
+     */
+    default Optional<User> getOwner() {
+        return getApi().getCachedUserById(getOwnerId());
+    }
```
ShindouMihou, line 90: Also, I am wondering whether to add `requestOwner` here as well since, for example, my bot doesn't use a user cache and this would only end up as Empty all the time.

But then again, `getOwnerId` exists though, I don't think it would hurt to add the `requestOwner` method like `Server` class has.

Feel free to state your opinions if any.

javacord-api/src/main/java/org/javacord/api/event/channel/thread/ThreadMembersUpdateEvent.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @return The users who were added to the thread.
+     */
+    public ThreadMember[] getAddedMembers();
+
+    /**
+     * @return The ids of the users who were removed from the thread.
+     */
+    public Long[] getRemovedMemberIds();
```
ShindouMihou, line 32: Does these have to be an Array instead of a Collection or List (since most of the time, Javacord only uses Lists or Collections as far as I can remember)?

Also, I think we can have `getRemovedMembers` instead of just the IDs as well? (Don't take my words for this granted though).

PhoenixInvertigo: Nope! They can totally be Lists; I was just implementing literally from the documentation. Probably better to make them Lists, so I did. Good call!

As far as the removed members, it's passing in member ids rather than member objects, so we'd need to do a lookup for that. I might make the change later, but I was kinda assuming the intended structure was, the api passes in all the ids, then you fetch the members at a later point.

(Admittedly, I haven't looked into much of the business logic yet, so it might be that you're right and we should just handle the user lookupById before creating this object)

javacord-core/src/main/java/org/javacord/core/entity/channel/ThreadMemberImpl.java
PhoenixInvertigo: Fixed! Thanks.

After I get all of the code written, I'll look at whatever formatting issues are left with the style checker and fix all of them.

javacord-core/src/main/java/org/javacord/core/entity/channel/ChannelThreadImpl.java
PhoenixInvertigo: Fixed.

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelThread.java
PhoenixInvertigo: Sounds good. I'll add it to my to-do list.

PhoenixInvertigo: Added it.

PhoenixInvertigo: Added.

KILLEliteMaste:
There are some changes for thread permissions which have to be considered https://github.com/discord/discord-api-docs/commit/a674ee95e9b72ceb480f753efde5320fdfaf5886

PhoenixInvertigo:
I'll put it on my to-do list. Gonna try and get some more of this tested and implemented this weekend.

KILLEliteMaste:
@PhoenixInvertigo please allow pushs to your branch

PhoenixInvertigo:
> @PhoenixInvertigo please allow pushs to your branch

How can I turn that on?

KILLEliteMaste:
My rework is not finished but I don't have that much time currently to further work on it.
It would be greatly appreciated if someone else can continue to work on it. From what I saw that needs to be done:
- Don't let the ServerThreadChannel extends the ServerTextChannel as they are not really related
- Move the "position" methods from the `ServerChannel` interface to somewhere else as ServerThreadChannels do not have a position
- Implement the REST endpoints

KILLEliteMaste:
@PhoenixInvertigo on the right side of this PR there's a checkbox "Allow edits...." is this activated? If not please do so.

PR #867, opened 2021-09-22
Context Menus

Mysterypotatoguy:
Implements Context Menus and renames some Slash Command components (SlashCommandPermissions) to generify them into Application Command components


javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandBuilder.java
```diff
@@ -11,7 +11,7 @@
 /**
  * This class is used to create new slash commands.
  */
-public class SlashCommandBuilder {
+public class SlashCommandBuilder implements ApplicationCommandBuilder<SlashCommand> {
```
KILLEliteMaste, line 14: Again, can't select it, `setName` `setDescription` `setDefaultPermissions` and both create methods can be moved to the ApplicationCommandBuilder class

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandUpdater.java
```diff
@@ -8,7 +8,7 @@
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
-public class SlashCommandUpdater {
+public class SlashCommandUpdater implements ApplicationCommandUpdater<SlashCommand> {
```
KILLEliteMaste, line 11: Same here, `setDescription` `setDefaultPermission` and the updater methods can be moved to the `ApplicationCommandUpdater`

javacord-api/src/main/java/org/javacord/api/interaction/UserContextMenu.java
```diff
@@ -0,0 +1,13 @@
...
+public interface UserContextMenu extends ContextMenu {
+
+    /**
+     * Creates a user context menu updater from this UserContextMenu instance.
+     *
+     * @return The user context menu updater for this UserContextMenu instance.
+     */
+    default UserContextMenuUpdater createUserContextMenuUpdater() {
+        return new UserContextMenuUpdater(this.getId());
+    }
```
KILLEliteMaste, line 12: Can be moved to the ContextMenu interface

javacord-api/src/main/java/org/javacord/api/interaction/UserContextMenuBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+import org.javacord.api.interaction.internal.UserContextMenuBuilderDelegate;
+import org.javacord.api.util.internal.DelegateFactory;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * This class is used to create new user context menus.
+ */
+public class UserContextMenuBuilder implements ApplicationCommandBuilder<UserContextMenu> {
```
KILLEliteMaste, line 14: Same as mentioned in the `MessageContextMenuBuilder`

javacord-api/src/main/java/org/javacord/api/interaction/UserContextMenuUpdater.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.javacord.api.DiscordApi;
+import org.javacord.api.entity.server.Server;
+import org.javacord.api.interaction.internal.UserContextMenuUpdaterDelegate;
+import org.javacord.api.util.internal.DelegateFactory;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public class UserContextMenuUpdater implements ApplicationCommandUpdater<UserContextMenu> {
```
KILLEliteMaste, line 11: Same as mentioned in the `MessageContextMenuUpdater`

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1344,117 +1351,212 @@ public ThreadPool getThreadPool() {
...
+            }
+        }
+        return Collections.unmodifiableList(messageContextMenus);
+    }
+
     @Override
-    public CompletableFuture<List<SlashCommand>> bulkOverwriteGlobalSlashCommands(
-            List<SlashCommandBuilder> slashCommandBuilderList) {
+    public CompletableFuture<List<ApplicationCommand>> bulkOverwriteGlobalApplicationCommands(
+            List<ApplicationCommandBuilder<?>> applicationCommandBuilderList) {
```
KILLEliteMaste, line 1533: This probably should be changed to 
```suggestion
            List<ApplicationCommandBuilder<? extends ApplicationCommand>> applicationCommandBuilderList) {
```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1344,117 +1351,212 @@ public ThreadPool getThreadPool() {
...
         ArrayNode body = JsonNodeFactory.instance.arrayNode();
-        for (SlashCommandBuilder slashCommandBuilder : slashCommandBuilderList) {
-            body.add(((SlashCommandBuilderDelegateImpl) slashCommandBuilder.getDelegate())
-                    .getJsonBodyForSlashCommand());
+        for (ApplicationCommandBuilder<?> applicationCommandBuilder : applicationCommandBuilderList) {
+            body.add(((ApplicationCommandBuilderDelegateImpl<?>) applicationCommandBuilder.getDelegate())
```
KILLEliteMaste, line 1536: Same here
```suggestion
        for (ApplicationCommandBuilder<? extends ApplicationCommand> applicationCommandBuilder : applicationCommandBuilderList) {
            body.add(((ApplicationCommandBuilderDelegateImpl<? extends ApplicationCommand>) applicationCommandBuilder.getDelegate())
```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1344,117 +1351,212 @@ public ThreadPool getThreadPool() {
...
                 .setBody(body)
-                .execute(result -> jsonToSlashCommandList(result.getJsonBody()));
+                .execute(result -> jsonToApplicationCommandList(result.getJsonBody()));
     }
 
     @Override
-    public CompletableFuture<List<SlashCommand>> bulkOverwriteServerSlashCommands(
-            Server server, List<SlashCommandBuilder> slashCommandBuilderList) {
+    public CompletableFuture<List<ApplicationCommand>> bulkOverwriteServerApplicationCommands(
+            Server server, List<ApplicationCommandBuilder<?>> applicationCommandBuilderList) {
```
KILLEliteMaste, line 1548:
```suggestion
            Server server, List<ApplicationCommandBuilder<? extends ApplicationCommand>> applicationCommandBuilderList) {
```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1344,117 +1351,212 @@ public ThreadPool getThreadPool() {
...
         ArrayNode body = JsonNodeFactory.instance.arrayNode();
-        for (SlashCommandBuilder slashCommandBuilder : slashCommandBuilderList) {
-            body.add(((SlashCommandBuilderDelegateImpl) slashCommandBuilder.getDelegate())
-                    .getJsonBodyForSlashCommand());
+        for (ApplicationCommandBuilder<?> applicationCommandBuilder : applicationCommandBuilderList) {
+            body.add(((ApplicationCommandBuilderDelegateImpl<?>) applicationCommandBuilder.getDelegate())
```
KILLEliteMaste, line 1551:
```suggestion
        for (ApplicationCommandBuilder<? extends ApplicationCommand> applicationCommandBuilder : applicationCommandBuilderList) {
            body.add(((ApplicationCommandBuilderDelegateImpl<? extends ApplicationCommand>) applicationCommandBuilder.getDelegate())
```

javacord-core/src/main/java/org/javacord/core/interaction/CommandType.java
```diff
@@ -0,0 +1,39 @@
+package org.javacord.core.interaction;
+
+public enum CommandType {
```
KILLEliteMaste, line 3:
```suggestion
public enum ApplicationCommandType {
```

javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
```diff
@@ -105,16 +128,51 @@ public void handle(JsonNode packet) {
...
-                        server == null ? api : server,
-                        server,
-                        interaction.getChannel().orElse(null),
-                        interaction.getUser(),
-                        slashCommandCreateEvent
-                );
+            case APPLICATION_COMMAND:
+                int applicationCommandTypeId = packet.get("data").get("type").asInt();
+                CommandType applicationCommandType = CommandType.fromValue(applicationCommandTypeId);
+                switch (applicationCommandType) {
```
KILLEliteMaste, line 134: What about an `ApplicationCommandEvent` ?

KILLEliteMaste (changes requested):
I didn't mention anything in the delegates as you can reflect / adjust the changes from the builder classes.

In `PermissionType` your forgot to change `USE_SLASH_COMMANDS` to `USE_APPLICATION_COMMANDS`

In `MessageType` you forgot to add `ContextMenuCommand` https://discord.com/developers/docs/resources/channel#message-object-message-types

In `RestRequestResultErrorCodes` your forgot to change `UNKNOWN_SLASH COMMAND` to `UNKNOWN_SLASH_COMMAND` and the description. And if you want you can also add the missing codes as there are quiet a few new ones or you can leave it open for another PR so we can create an issue

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
```diff
@@ -0,0 +1,52 @@
+package org.javacord.api.interaction;
+
+import org.javacord.api.entity.DiscordEntity;
+import org.javacord.api.entity.server.Server;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface ApplicationCommand extends DiscordEntity {
+
```
KILLEliteMaste, line 9: You should add a `getType` method

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1344,117 +1351,212 @@ public ThreadPool getThreadPool() {
 
     @Override
     public CompletableFuture<List<SlashCommand>> getGlobalSlashCommands() {
```
KILLEliteMaste, line 1353: Another thing I came across, I think you should remove the other methods for the context menu and change this one to `getGlobalApplicationCommands`. This removes quiet a lot of code and the user can simply get the needed application commands by themself if you add the above mentioned method by `list.stream().filter(command -> command.getType == CommandType.SLASH).map(command -> (SlashCommand)command).toList()` . If we decide to not remove any methods there should be at least a `getGlobalApplicationCommands` method otherwise 3 request are needed (and at some time even more with more application command types which also need to be patched in to acces them)

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/MessageType.java
```diff
@@ -58,53 +58,17 @@
...
-                return SERVER_DISCOVERY_GRACE_PERIOD_FINAL_WARNING;
-            case 19:
-                return REPLY;
-            case 20:
-                return SLASH_COMMAND;
-            default:
-                return UNKNOWN;
+        for (final MessageType messageType : values()) {
+            if (messageType.type == type) {
+                return messageType;
```
Saladoc, line 68: How does this correctly differentiate between `NORMAL_WEBHOOK` and `NORMAL`?

Saladoc (changes requested):
looks good (except for the one question), thanks for your work

javacord-api/src/main/java/org/javacord/api/entity/message/MessageType.java
KILLEliteMaste: Somehow I missed that. But what is the actual reason we have this type for? I would remove this and instead add a `isWebhookMessage` in the message or don't do it at all as you can check if the message is from a webhook by `message.getAuthor().isWebhook()`. What do you think?

Saladoc: Possible, but it should not be done on the fly as a side effect of this PR. 

We could include an if clause to catch that specific case for now, and bring forward a change request to remove `NORMAL_WEBHOOK` or adjust the enum instances to also have the proper `isWebhook` properties and check those in the loop as well.

KILLEliteMaste: > Possible, but it should not be done on the fly as a side effect of this PR.

Yeah, I will revert this change and open another PR that will remove this for the version 3.5 breaking change release

Saladoc (approved):

PR #868, opened 2021-09-30
Always include the GUILDS intent

KILLEliteMaste:
The `GUILDS` intent should always be included as otherwise Javacord is not functional. Therefore it doesn't make sense to not include it


coaxial1 (approved):
LGTM

PR #869, opened 2021-09-30
Add new appender types to the MessageBuilder.

KILLEliteMaste:
- Added appendNamedLink
- Added appendTimestamp
- Added TimestampStyle enum

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/TimestampStyle.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/TimestampStyle.java
@@ -1,0 +1,112 @@
+package org.javacord.api.entity.message;
+
+import java.time.Instant;
+
+/**
+ * The enum contains all different types of timestamp styles.
+ */
+public enum TimestampStyle {
+
+    /**
+     * Displayed as: 16:20.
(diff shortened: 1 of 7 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/TimestampStyle.java
```diff
@@ -0,0 +1,100 @@
...
+     * @return The tag of the timestamp.
+     */
+    public String getTimestampTag(final long epochSeconds) {
+        return "<t:"
+                + epochSeconds
+                + ":"
+                + timestampStyle
+                + ">";
+    }
+
```
Saladoc, line 99: Can we also add a method taking an `Instant` an argument?

Saladoc (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/TimestampStyle.java
KILLEliteMaste: Added methods that take an instant as an argument for the TimestampStyle and for the MessageBuilder

PR #872, opened 2021-09-30
Changed isSubcommandOrGroup to check the stringRepresentation instead of stringValue

KILLEliteMaste:


javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteractionOption.java
```diff
@@ -27,7 +27,7 @@
      * @return If the option is a subcommand or group.
      */
     default boolean isSubcommandOrGroup() {
-        return !getStringValue().isPresent();
+        return !getStringRepresentationValue().isPresent();
```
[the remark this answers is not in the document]
KILLEliteMaste: Gonna change the commit message afterwards. The string value is present if the SlashCommandOptionType is of type `STRING` and the string representation is present once there is any value `BOOLEAN`, `INTEGER`, `USER` and so on.

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
```diff
@@ -41,11 +41,11 @@
      * @param api      The DiscordApi instance.
      * @param jsonData The json data of the option.
      */
-    public SlashCommandInteractionOptionImpl(DiscordApi api, JsonNode jsonData) {
+    public SlashCommandInteractionOptionImpl(final DiscordApi api, final JsonNode jsonData) {
```
coaxial1, line 44: Is there a specific reason to add all the final modifiers? 
+ At least for method parameters, it seems like this is not something we're usually doing

KILLEliteMaste: Well, I think we should introduce final more often because theres not really any disadvantage? It's an easy way to add some micro optimization and it can help to prevent errors of reassigning variables and such. 
I was even thinking of quick fixing everything via IntelliJ to add final where possible. What do you think about it or do you in general disagree to introduce it more often?

coaxial1: Mh I don't think necessarily it's bad thing to do, it was just feeling unrelated to the PR :D and I think it "looks and feels" a bit weird when being used as parameters, but may objectively be a good thing. On the other side, doing it everywhere is polluting the code with lots and lots of final modifiers, so idk
Wouldn't necessarily do it in my own code :)

KILLEliteMaste: Im also not using it that excessively in my code too, but considering this is a library it may be worth? Maybe another opinion on this would be nice 🙂

PR #885, opened 2021-10-13
Add size parameter to User#getAvatar

Mysterypotatoguy:
Add the ability to change the size of the avatar when fetching as an Icon

Ideally I would have made this an option on the Icon itself to pair/combine with #879 but URL objects do not allow adding query parameters after they are constructed


ShindouMihou:
I'll actually leave this behind, depending on what (CDN) the Javacord one is pointing to, the image will not actually be resized as far as my bot had experienced. It seems like Discord changed something back then where `cdn.discordapp.com` wouldn't resize the images but `media.discordapp.com` would. I don't know if this issue only applied to my bot but it shouldn't be.

PR #887, opened 2021-10-15
Upgrade discord api version to v9

felldo:


Saladoc:
This happens already in #864.

KILLEliteMaste:
> This happens already in #864.

In your mentioned PR only the Gateway version has been upgraded but the API is still v8.

Issue #865, opened 2021-09-14
Add new "channel_types" field for CHANNEL option of SlashCommands


KILLEliteMaste:
https://github.com/discord/discord-api-docs/commit/72bf5acb5de5d804efb70039df7db1daf0c4b33b

PR #888, opened 2021-10-18
Added channel_type field for SlashCommandOption

ILGO0413:
Fixes https://github.com/Javacord/Javacord/issues/865


KILLEliteMaste:
Furthermore you probably checked out the master branch. Please rebase to the development branch

KILLEliteMaste:
It's okay. I will do it later when I have some time

Saladoc:
> I have tried to squash commits in my branch, but it made even few more instead of only 1 squashed commit. Is it possible to use 'squash and merge' option on pull request merging?

Squashing and rebasing means rewriting the commit history. Old commits technically get lost as their changes are now residing in different commits. Since git is very eager to never let you lose any information, you will have to explicitly tell it to forcefully push over the previous branch. It is likely you did not do this and instead git tried to merge your (seemingly) different changesets.

Issue #889, opened 2021-10-18
Add min_value and max_value for slash command options


KILLEliteMaste:
See: https://github.com/discord/discord-api-docs/pull/3967

ILGO0413:
I made a pull request, but it should be discussed probably. 
New min_value and max_value properties should be integer for `INTEGER` options, double for `NUMBER` options according to discord api docs. Therefore I implemented it as a pair of `INTEGER` min/max values (between -2^53 and 2^53, therefore long is used), getters for `INTEGER` values check command option type and throw an UnsupportedOperationException if it is different. `NUMBER` logic can be implemented the same way later (based on double primitive type).
It is probably possible to use only double min/max values, and it would even work for small values, but I am not sure if any issues with double precision can be faced.

PR #891, opened 2021-10-18
Add min_value and max_value

ILGO0413:
Fixes https://github.com/Javacord/Javacord/issues/889

Keep in mind that NUMBER option type can be implemented later, and depending on option type, the min/max_value format can be different.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
@@ -5,6 +5,7 @@ import org.javacord.api.entity.channel.ChannelType;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -64,6 +65,34 @@ public interface SlashCommandOption {
      * @return A set with the channel types that are shown.
      */
     Set<ChannelType> getChannelTypes();
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#INTEGER} type, the minimum value permitted.
+     *
+     * @return The minimum value permitted.
+     */
+    Optional<Long> getIntegerMinValue();
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#INTEGER} type, the maximum value permitted.
+     *
+     * @return The maximum value permitted.
+     */
+    Optional<Long> getIntegerMaxValue();
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#NUMBER} type, the minimum value permitted.
+     *
+     * @return The minimum value permitted.
+     */
+    Optional<Double> getNumberMinValue();
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#NUMBER} type, the maximum value permitted.
+     *
+     * @return The maximum value permitted.
+     */
+    Optional<Double> getNumberMaxValue();
 
     /**
      * Create a new slash command option to be used with a slash command builder.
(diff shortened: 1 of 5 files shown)


ILGO0413:
> 
> 
> This PR is not functional, maybe your forgot to commit something?
> 
> To keep the conversation at one place in #889 you metnioned that the methods for `INTEGER` take arguments of type `long`, I don't really get why you did it like that. And why should the methods for the `NUMBER` option implemented later and not now?

Yep, my commit is lost somehow and I am investigating what's happened. Gonna update it soon.
If 'NUMBER' option is already implemented, gonna implement min max values for it as well.

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandOptionImpl.java
```diff
@@ -128,35 +128,23 @@ public boolean isRequired() {
...
     }
 
     @Override
-    public double getNumberMaxValue() {
-        if (type != SlashCommandOptionType.NUMBER) {
-            throw new UnsupportedOperationException("The method should be invoked for NUMBER option only");
-        }
-        return numberMaxValue;
+    public Optional<Double> getNumberMaxValue() {
+        return Optional.of(numberMaxValue);
```
KILLEliteMaste: Use `Optional#ofNullable` instead of `Optional#of`

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandOptionImpl.java
```diff
@@ -20,6 +21,10 @@
     private final boolean required;
     private final List<SlashCommandOptionChoice> choices;
     private final List<SlashCommandOption> options;
+    private Long integerMinValue;
+    private Long integerMaxValue;
+    private Double numberMinValue;
+    private Double numberMaxValue;
```
KILLEliteMaste, line 27: Can be final but then you have to set them explicitly `null` depending on what type the option is in the `if` statement or `else` block

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandOptionImpl.java
ILGO0413: It can be final, but in this case we should handle 3 options:
1) type is INTEGER, set 2 integer values and set both number values as null
2) type is NUMBER. set 2 number values and set both integer as null
3) type is neither INTEGER nor NUMBER, set all 4 values as null. 
Is it OK to use 3 logical blocks for 4 variables initialization? AFAIK, Long and Double are immutable and no setters are provided

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandOptionImpl.java
ILGO0413: Shame on me. Fixed

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandOptionImpl.java
KILLEliteMaste: Yes, that should just be ~8 more lines for variable initialization. But what does the immutability of Longs and Doubles have to do with this?

ILGO0413: Well, I suppose its the main idea to define internal variable as final - to make sure it can be provided only when the object is created and it cannot be reinitialized later. Previous implementation had no 'final' keyword, but user had no possibility to set variable after object creation, and had no possibility to change the variable itself due to Long/Double immutability.

New implementation is provided, it covers all 3 options. Please compare and choose what implementation is more preferable

KILLEliteMaste: > Previous implementation had no 'final' keyword, but user had no possibility to set variable after object creation, and had no possibility to change the variable by reference itself due to Long/Double immutability.

It's not only about the user being able to change the assignment of a variable but also to prevent in the first place developers to do something that may be wrong because if you use final you have to think twice about it if that what you are doing really makes sense.

KILLEliteMaste (approved):
LGTM!
Please rebase to the latest version and resolve the merge conflicts.

ILGO0413:
And I failed on the rebase again

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
```diff
@@ -66,6 +67,35 @@
...
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#NUMBER} type, the maximum value permitted.
+     *
+     * @return The maximum value permitted.
+     */
+    Optional<Double> getNumberMaxValue();
+
+    /**
+
```
KILLEliteMaste, line 98: Remove line

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
```diff
@@ -214,4 +244,97 @@ static SlashCommandOption createChannelOption(String name,
                 .setChannelTypes(channelTypes)
                 .build();
     }
+
+
```
KILLEliteMaste, line 248: Remove line

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandOptionImpl.java
```diff
@@ -121,6 +155,26 @@ public boolean isRequired() {
...
+    }
+
+    @Override
+    public Optional<Double> getNumberMinValue() {
+        return Optional.of(numberMinValue);
+    }
+
+    @Override
+    public Optional<Double> getNumberMaxValue() {
+        return Optional.of(numberMaxValue);
```
KILLEliteMaste, line 175: Change to `Optional#ofNullable`

KILLEliteMaste (changes requested):
I have no idea what you are doing but the rebase really messed this up. Some previous changes are nore not applied anymore and there are still conflicts

Vampire:
I hopefully resolved your merge mess and rebased onto latest `development`.
So now the history should be nice and intact.
Be aware to **NOT** do a pull with merge now, even if Git suggests to do so.

If someone made a force push, rewriting history, and you try to push a new commit, Git suggests to do a pull with merge, but this is one of the worst things you can do as you then somehow merge the old and new history and create an even bigger mess which I believe you did before to get to that state this PR was in. Just reset your local branch to the remote branch and if you had additional local commits, cherry-pick them on top for example.

KILLEliteMaste (approved):
LGTM!
@Vampire would you also be so kind to squash the commits into a single one? :)

Vampire:
Sure, no problem, done.
As you approved it I'll also push to dev when the checks are through.
I just did not do a content review, just fixed the merge-mess as I need that feature. :-D

PR #903, opened 2021-10-24
Add ServerThreadChannelUpdater & ServerThreadChannelBuilder

ruiner189:
Adds an updater for threads.
Some breaking changes: 
- Moved canCreateInstantInvite & permission methods from the ServerChannel to the new RegularServerChannel
- Moved permission methods from ServerChannel-Updater/Builder to RegularServerChannel-Updater/Builder


KILLEliteMaste:
@ruiner189 after a short internal discussion we came up with a different name `RegularServerChannel` which could replace the `PositionableServerChnnel` pretty well. Then you can move the permission methods to it's updater

javacord-api/src/main/java/org/javacord/api/entity/channel/internal/ServerThreadChannelUpdaterDelegate.java
```diff
@@ -0,0 +1,54 @@
...
+ * You usually don't want to interact with this object.
+ */
+public interface ServerThreadChannelUpdaterDelegate extends ServerChannelUpdaterDelegate {
+
+    /**
+     * Queues the name to be updated.
+     *
+     * @param name The new name of the thread.
+     */
+    void setName(String name);
```
Vampire, line 17: This is unnecessary, it is already inherited from super interface

javacord-api/src/main/java/org/javacord/api/entity/channel/internal/ServerThreadChannelUpdaterDelegate.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Queues the archived to be updated.
```
Vampire, line 20: "flag" is missing

javacord-api/src/main/java/org/javacord/api/entity/channel/internal/ServerThreadChannelUpdaterDelegate.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Queues the auto archive duration to be updated.
+     *
+     * @param autoArchiveDuration The new auto archive duration of the thread.
+     */
+    void setAutoArchiveDuration(AutoArchiveDuration autoArchiveDuration);
+
+    /**
+     * Queues the locked to be updated.
```
Vampire, line 34: "flag" is missing

javacord-api/src/main/java/org/javacord/api/entity/channel/internal/ServerThreadChannelUpdaterDelegate.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param locked The new locked flag of the thread.
+     */
+    void setLockedFlag(boolean locked);
+
+    /**
+     * Queues the invitable to be updated.
```
Vampire, line 41: "flag" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+import org.javacord.api.entity.channel.internal.RegularServerChannelBuilderDelegate;
+import org.javacord.api.entity.permission.Permissions;
+
+/**
+ * This class is used to create new regular server channels.
+ */
+public class RegularServerChannelBuilder extends ServerChannelBuilder {
+
+    /**
+     * The regular server channel delegate used by this instance.
```
Vampire, line 14: "builder" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    private final RegularServerChannelBuilderDelegate delegate;
+
+    /**
+     * Creates a new regular server channel builder without delegate.
+     */
+    protected RegularServerChannelBuilder() {
+        delegate = null;
```
Vampire, line 22: What is the point of having a `delegate` field that is always `null`?
If it ever is used, it will throw `NullPointerException`.
It only works if every subclass overwrites the methods using this field.
I guess a more correct solution would be if the subclasses supply the `delegate` in the super constructor call and do not overwrite the methods.
Or if they are expected to overwrite the methods, the field should be removed and the methods made abstract, but I think that would be the worse solution.

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+            T permissionable, Permissions permissions) {
+        delegate.addPermissionOverwrite(permissionable, permissions);
+        return this;
+    }
+
+    /**
+     * Removes a permission overwrite for the given entity.
+     *
+     * @param <T>            The type of entity to hold the permission, usually <code>User</code> or <code>Role</code>
+     * @param permissionable The entity which permission overwrite should be removed.
```
Vampire, line 43: "which" is not correct, is it?

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelUpdater.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * The server channel delegate used by this instance.
+     */
+    private final RegularServerChannelUpdaterDelegate delegate;
+
+    /**
+     * Creates a new server channel updater without delegate.
+     */
+    protected RegularServerChannelUpdater() {
+        delegate = null;
```
Vampire, line 20: Same as for `RegularServerChannelBuilder` just that here a second constructor exists

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelUpdater.java
```diff
@@ -0,0 +1,71 @@
...
+            T permissionable, Permissions permissions) {
+        delegate.addPermissionOverwrite(permissionable, permissions);
+        return this;
+    }
+
+    /**
+     * Removes a permission overwrite for the given entity.
+     *
+     * @param <T> The type of entity to hold the permission, usually <code>User</code> or <code>Role</code>
+     * @param permissionable The entity which permission overwrite should be removed.
```
Vampire, line 64: "which" is not correct, is it?

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelUpdater.java
```diff
@@ -0,0 +1,71 @@
...
+import org.javacord.api.entity.DiscordEntity;
+import org.javacord.api.entity.Permissionable;
+import org.javacord.api.entity.channel.internal.RegularServerChannelUpdaterDelegate;
+import org.javacord.api.entity.permission.Permissions;
+import org.javacord.api.util.internal.DelegateFactory;
+
+public class RegularServerChannelUpdater extends ServerChannelUpdater {
+
+    /**
+     * The server channel delegate used by this instance.
```
Vampire, line 12: "regular" and "updater" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelUpdater.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    private final RegularServerChannelUpdaterDelegate delegate;
+
+    /**
+     * Creates a new server channel updater without delegate.
```
Vampire, line 17: "regular" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelUpdater.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    protected RegularServerChannelUpdater() {
+        delegate = null;
+    }
+
+    /**
+     * Creates a new server channel updater.
```
Vampire, line 24: "regular" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.javacord.api.entity.DiscordEntity;
+import org.javacord.api.entity.Permissionable;
+import org.javacord.api.entity.channel.internal.RegularServerChannelBuilderDelegate;
+import org.javacord.api.entity.permission.Permissions;
+
+/**
+ * This class is used to create new regular server channels.
+ */
+public class RegularServerChannelBuilder extends ServerChannelBuilder {
```
Vampire, line 11: methods for (raw)position are missing

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannelUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+import org.javacord.api.util.internal.DelegateFactory;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * This class can be used to update server thread channels.
+ */
+public class ServerThreadChannelUpdater extends ServerChannelUpdater {
+
+    /**
+     * The server thread channel delegate used by this instance.
```
Vampire, line 13: "updater" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannelUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param name The new name of the thread.
+     * @return The current instance in order to chain call methods.
+     */
+    public ServerThreadChannelUpdater setName(String name) {
+        delegate.setName(name);
+        return this;
+    }
+
+    /**
+     * Queues the archived to be updated.
```
Vampire, line 39: "flag" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannelUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param autoArchiveDuration The new auto archive duration of the thread.
+     * @return The current instance in order to chain call methods.
+     */
+    public ServerThreadChannelUpdater setAutoArchiveDuration(AutoArchiveDuration autoArchiveDuration) {
+        delegate.setAutoArchiveDuration(autoArchiveDuration);
+        return this;
+    }
+
+    /**
+     * Queues the locked to be updated.
```
Vampire, line 61: "flag" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannelUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param locked The new locked flag of the thread.
+     * @return The current instance in order to chain call methods.
+     */
+    public ServerThreadChannelUpdater setLockedFlag(boolean locked) {
+        delegate.setLockedFlag(locked);
+        return this;
+    }
+
+    /**
+     * Queues the invitable to be updated.
```
Vampire, line 72: "flag" missing

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannelUpdater.java
```diff
@@ -0,0 +1,102 @@
+package org.javacord.api.entity.channel;
+
+import org.javacord.api.entity.channel.internal.ServerThreadChannelUpdaterDelegate;
+import org.javacord.api.util.internal.DelegateFactory;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * This class can be used to update server thread channels.
+ */
+public class ServerThreadChannelUpdater extends ServerChannelUpdater {
```
Vampire, line 10: What about a `ServerThreadChannelBuilder`, is that not possible?

javacord-api/src/main/java/org/javacord/api/util/internal/DelegateFactoryDelegate.java
```diff
@@ -212,13 +214,13 @@
     ServerChannelUpdaterDelegate createServerChannelUpdaterDelegate(ServerChannel channel);
 
     /**
-     * Creates a new positionable server channel updater delegate.
+     * Creates a new regular server channel updater delegate.
      *
      * @param channel The channel to update.
      * @return A new server channel updater delegate.
```
Vampire, line 220: "regular" missing

javacord-core/src/main/java/org/javacord/core/entity/channel/RegularServerChannelBuilderDelegateImpl.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    protected void prepareBody(ObjectNode body) {
+        super.prepareBody(body);
+        ArrayNode permissionOverwrites = null;
+        if (overwrittenUserPermissions.size() + overwrittenRolePermissions.size() > 0) {
+            permissionOverwrites = body.putArray("permission_overwrites");
+        }
+        if (permissionOverwrites != null) {
```
Vampire, line 61: Is there a reason for this unnecessary check?
The only time this is `null` is, if the added sizes are 0 which means the for-loops below will not be entered anyway.

javacord-core/src/main/java/org/javacord/core/entity/channel/RegularServerChannelImpl.java
```diff
@@ -1,19 +1,19 @@
...
 
 import com.fasterxml.jackson.databind.JsonNode;
-import org.javacord.api.entity.channel.PositionableServerChannel;
+import org.javacord.api.entity.channel.RegularServerChannel;
 import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.entity.server.ServerImpl;
 import java.util.Comparator;
 
-public class PositionableServerChannelImpl extends ServerChannelImpl implements PositionableServerChannel {
+public class RegularServerChannelImpl extends ServerChannelImpl implements RegularServerChannel {
```
Vampire, line 9: Shouldn't also here all the permission related methods be moved down to `RegularServer...`?

javacord-core/src/main/java/org/javacord/core/entity/channel/RegularServerChannelUpdaterDelegateImpl.java
```diff
@@ -0,0 +1,147 @@
...
+        populatePermissionOverwrites();
+        if (permissionable instanceof Role) {
+            overwrittenRolePermissions.remove(permissionable.getId());
+        } else if (permissionable instanceof User) {
+            overwrittenUserPermissions.remove(permissionable.getId());
+        }
+    }
+
+    @Override
+    public CompletableFuture<Void> update() {
```
Vampire, line 85: Why is this overwritten?
The only difference to the superclass is, that it does not support audit log reason.
Does it not support audit log reason?
Then it would be bad that you can still set it on the updater but it is just silently ignored.

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerThreadChannelUpdaterDelegateImpl.java
```diff
@@ -0,0 +1,136 @@
...
+            body.put("rate_limit_per_user", delay);
+            patchThread = true;
+        }
+        return patchThread;
+    }
+
+    @Override
+    public CompletableFuture<Void> update() {
+        ObjectNode body = JsonNodeFactory.instance.objectNode();
+        super.prepareUpdateBody(body);
```
Vampire, line 125: This should be called in this classes `prepareUpdateBody` architecturally.
And also like it is, its return value is ignored, so if you just want to update a threads name, the update request will not be done.

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerThreadChannelUpdaterDelegateImpl.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+        if (delay != null) {
+            body.put("rate_limit_per_user", delay);
+            patchThread = true;
+        }
+        return patchThread;
+    }
+
+    @Override
+    public CompletableFuture<Void> update() {
```
Vampire, line 123: Why is this overwritten? The only difference to the superclass is, that it calls the superclass `prepareUpdateBody` which is misplaced here anyway.

Vampire (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelBuilder.java
KILLEliteMaste: Actually that's how it has been done in the past. The methods are also overwritten in i.e. ServerTextChannelBuilder and the other types but I will change this

Vampire: Yeah, seen that after I wrote the comment, but perfect opportunity to improve the code quality here. :-)

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannelUpdater.java
KILLEliteMaste: Possible for sure. Currently it's done through a `Message` or `ServerTextChannel` where you can call `createThread` or `createThreadForMessage` and I am not sure if it makes sense to add a builder for threads too atm as they would also split into 2 builder because of a thread with message and without. As they also have not that many possible arguments atm I would vote for only having the `createThread` methods. 
https://discord.com/developers/docs/resources/channel#start-thread-with-message
What do you think?

Vampire: Hm, well, it just felt a bit inconsistent.
For all other types we have a builder and we have an updater than can be constructed directly or through the object to be updated.

Vampire: Also why would we need two builders?
One could handle both cases, couldn't it?
You have one constructor that takes a message and one constructor that takes type and invitable flag.

KILLEliteMaste: Yeah but I thought as they differentiate in the possible arguments we should not offer these methods to set. Or is it just for the sake of completeness and be able to do whatever you want with the builder in case something changes from discords side?

Vampire: Not sure what you mean.
Besides one takes type and invitable and the other takes a message, they have the same possibilities, don't they?
There would be no setters for message, invitable, and type.

Btw. while talking about it, the updater has a field for invitable flag.
This is probably also bad if that flag is not valid for message-based threads?

KILLEliteMaste: I am not sure but the docs don't mention anything https://discord.com/developers/docs/resources/channel#modify-channel-json-params-thread

Vampire: grafik

Vampire: It seems to even not only be in non-message threads, but only on private threads.
Actually that's also what the "start a thread" docs say:
grafik

Vampire: This "only for private threads" might fall under the category "we don't check it in Javacord, but let Discord validate that" though.

KILLEliteMaste: Ah yeah just saw it. I guess mentioning in the javadoc that this only works for private threads might be enough

Vampire (changes requested):
A few more minor things

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelUpdater.java
```diff
@@ -1,37 +1,32 @@
...
+public class ServerChannelUpdater<T extends ServerChannelUpdater<T>> {
 
     /**
      * The server channel delegate used by this instance.
      */
-    private final ServerChannelUpdaterDelegate delegate;
+    private final ServerChannelUpdaterDelegate serverChannelUpdaterDelegate;
 
     /**
      * Creates a new server channel updater without delegate.
```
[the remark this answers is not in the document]
KILLEliteMaste: Not sure, you wrote this :D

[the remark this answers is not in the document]
Vampire: Well, there it was probably correct as delegate was null, but you kept it and even added it to the other. :-P

javacord-api/src/main/java/org/javacord/api/entity/channel/RegularServerChannelBuilder.java
```diff
@@ -16,7 +16,7 @@
     protected final RegularServerChannelBuilderDelegate delegate;
 
     /**
-     * Creates a new regular server channel builder without delegate.
+     * Creates a new regular server channel builder delegate.
```
Vampire, line 19: -"delegate" too

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelBuilder.java
```diff
@@ -15,7 +15,7 @@
     protected final ServerChannelBuilderDelegate delegate;
 
     /**
-     * Creates a new server channel builder without delegate.
+     * Creates a new server channel builder delegate.
```
Vampire, line 18: -"delegate" too

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelUpdater.java
```diff
@@ -15,14 +15,14 @@
     private final ServerChannelUpdaterDelegate serverChannelUpdaterDelegate;
 
     /**
-     * Creates a new server channel updater without delegate.
+     * Creates a new server channel updater delegate.
```
Vampire, line 18: -"delegate" too

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerChannelUpdater.java
```diff
@@ -15,14 +15,14 @@
...
      */
     protected ServerChannelUpdater(ServerChannelUpdaterDelegate serverChannelUpdaterDelegate) {
         this.serverChannelUpdaterDelegate = serverChannelUpdaterDelegate;
     }
 
     /**
-     * Creates a new server channel updater without delegate.
+     * Creates a new server channel updater delegate.
```
Vampire, line 25: -"delegate" too

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerStageVoiceChannel.java
```diff
@@ -17,4 +18,23 @@ default ChannelType getType() {
      * @return The topic of this channel.
      */
     Optional<String> getTopic();
+
+    @Override
+    default Optional<? extends ServerStageVoiceChannel> getCurrentCachedInstance() {
```
Vampire, line 23: Here no extends is necessary as long as there is no further subclass

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerStageVoiceChannel.java
```diff
@@ -17,4 +18,23 @@ default ChannelType getType() {
...
+        return getApi().getServerById(getServer().getId()).flatMap(server -> server.getChannelById(getId()))
+                .flatMap(Channel::asServerStageVoiceChannel);
+    }
+
+    @Override
+    default CompletableFuture<? extends ServerStageVoiceChannel> getLatestInstance() {
```
Vampire, line 29: Here no extends is necessary as long as there is no further subclass

javacord-core/src/main/java/org/javacord/core/audio/AudioConnectionImpl.java
```diff
@@ -255,7 +255,8 @@ public void setPrioritySpeaking(boolean prioritySpeaking) {
         return Collections.unmodifiableSet(speakingFlags);
     }
 
-    /**M
+    /**
+     * M
```
Vampire, line 259: "M"?

Vampire (changes requested):
Aaaalmost there :-)

KILLEliteMaste:
Okay, but now there shouldn't be anything left 😃

Vampire (approved):
Well, I'm not going to surch further. :-D
lgtm now.
Feel free to bring it over after you combined it into 1 or 2 commits, the sooner the better for me. :-)

PR #909, opened 2021-10-30
Update PermissionType.java

tomsaul:
Added new permissions values from Discord API.  Not sure if these changes MUST be added elsewhere, but this at least lets them be seen/used.


KILLEliteMaste:
You probably took the master branch (where this branch is also targeted on). If you have a look at the dev branch the permissions are already up to date as you can see here https://github.com/Javacord/Javacord/blob/58f8f5f2756bec90c16f7996e52c02da19e110da/javacord-api/src/main/java/org/javacord/api/entity/permission/PermissionType.java

Issue #822, opened 2021-07-23
Message#edit should edit the whole messages instead of only the parameter


KILLEliteMaste:
Currently i.e. `Message#edit(String)` only updates the message content instead of removing the embed and components. The method should be changed to "replace" the message with the given new content and corresponding new methods i.e. `Message#editContent` should be added to only update the specific field.

jyp17:
I'd like to take over this issue if that's alright. Should I ask questions concerning the issue here or on the discord?

KILLEliteMaste:
> I'd like to take over this issue if that's alright. Should I ask questions concerning the issue here or on the discord?

If you have general questions our Discord may be a better place to get answers faster.

PR #914, opened 2021-10-31
Add editContent and editEmbed to Message and UncachedMessageUtil

MatthewMerrill:
Adds content/embed specific methods to Message and UncachedMessageUtilImpl.

 - Message#editContent
 - Message#editEmbed
 - UncachedMessageUtil#editContent(long, long, String)
 - UncachedMessageUtil#editContent(String, String, String)
 - UncachedMessageUtil#editEmbed(long, long, EmbedBuilder)
 - UncachedMessageUtil#editContent(String, String, EmbedBuilder)

Blocker for resolving #822. Sent as a separate PR so discussions on how to handle the breaking change can be reserved for the commit that makes the breaking change.


KILLEliteMaste:
Your branch is based on the master instead of the development branch. So you have to rebase your branch on the development to get the latest changes.
Furthermore I don't see any benefit over doing a 2nd PR for the actual breaking changes as the changes included in this PR are basically just copies of the already existing methods but with other naming as far as i've seen. In this manner I think the brekaing changes should be included right away.

PR #917, opened 2021-11-26
Add an event for any gateway events.

e042:
Hey,
this PR contains only one small change. I added an event `GatewayDispatchEvent` which will be fired on any gateway dispatch. This can be helpful if you want to use e.g. new, not yet implemented features or debug something.

Since this raises some security issues, this event is only fired if the environment variable `ENABLEINSECURE` has an arbitrary content.

I hope this PR fits into your concept 😄


ShindouMihou:
I don't know if enforcing environment variables is a good choice for Javacord.

PR #919, opened 2021-11-30
Fix inconsistency in slash command option naming

Vampire:


javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteractionOption.java
```diff
@@ -135,7 +135,7 @@ default boolean isSubcommandOrGroup() {
      *
      * @return The mentionable value of this option.
      */
```
KILLEliteMaste, line 137: Can't select it all but the Javadoc is wrong

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteractionOptionsProvider.java
```diff
@@ -168,11 +168,11 @@
      *
      * @return An Optional with the double value of such an option if it exists; an empty Optional otherwise
      * @deprecated Use {@link #getOptions() getOptions().get(0).getNumberValue()}
-     *     or {@link #getOptionNumberValueByIndex(int) getOptionNumberValueByIndex(0)} instead.
+     *     or {@link #getOptionDecimalValueByIndex(int) getOptionDecimalValueByIndex(0)} instead.
      */
     @Deprecated
     default Optional<Double> getFirstOptionNumberValue() {
```
KILLEliteMaste, line 174: Won't mark every place but the same should be done for `Second` and `Third` NumberValue
```suggestion
    default Optional<Double> getFirstOptionDecimalValue() {
```

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteractionOptionsProvider.java
```diff
@@ -168,11 +168,11 @@
      *
      * @return An Optional with the double value of such an option if it exists; an empty Optional otherwise
      * @deprecated Use {@link #getOptions() getOptions().get(0).getNumberValue()}
```
KILLEliteMaste, line 170: Won't mark every place but the same should be done for `Second` and `Third` NumberValue
```suggestion
     * @deprecated Use {@link #getOptions() getOptions().get(0).getDecimalValue()}
```

javacord-api/src/main/java/org/javacord/api/interaction/internal/SlashCommandOptionBuilderDelegate.java
```diff
@@ -86,32 +86,32 @@
...
      */
-    void setIntegerMaxValue(long integerMaxValue);
+    void setLongMaxValue(long longMaxValue);
 
     /**
-     * Sets the {@link SlashCommandOptionType#NUMBER} min value for the slash command option.
+     * Sets the {@link SlashCommandOptionType#DECIMAL} min value for the slash command option.
      *
-     * @param numberMinValue The integer min value.
+     * @param decimalMinValue The number min value.
```
KILLEliteMaste, line 105:
```suggestion
     * @param decimalMinValue The decimal min value.
```

javacord-api/src/main/java/org/javacord/api/interaction/internal/SlashCommandOptionBuilderDelegate.java
```diff
@@ -86,32 +86,32 @@
...
      */
-    void setNumberMinValue(double numberMinValue);
+    void setDecimalMinValue(double decimalMinValue);
 
     /**
-     * Sets the {@link SlashCommandOptionType#NUMBER} max value for the slash command option.
+     * Sets the {@link SlashCommandOptionType#DECIMAL} max value for the slash command option.
      *
-     * @param numberMaxValue The integer max value.
+     * @param decimalMaxValue The number max value.
```
KILLEliteMaste, line 112:
```suggestion
     * @param decimalMaxValue The decimal max value.
```

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOptionType.java
```diff
@@ -18,7 +18,7 @@
     /**
      * Any double between -2^53 and 2^53.
```
KILLEliteMaste, line 19:
```suggestion
     * Any double between -2^53 and 2^53. This is the NUMBER option according to the Discord docs.
```

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionOptionImpl.java
```diff
@@ -35,7 +35,7 @@
     private final Long channelValue;
     private final Long roleValue;
     private final Long mentionableValue;
-    private final Double numberValue;
+    private final Double decimalValue;
```
KILLEliteMaste, line 38:
```suggestion
    /**
    * This is the NUMBER option according to the Discord docs.
    */
    private final Double decimalValue;
```

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteractionOptionsProvider.java
Vampire: I intentionally left those out, as they are deprecated anyway.
Renaming them now would immediately remove the old ones and introduce new methods that are right-away deprecated.
That felt strange to me, so I kept them with the old name.
What do you think?

KILLEliteMaste: Right, that would be a bit strange. But that applies for both options, having inconsistent naming again or like you said introducing new methods that are right away deprecated. But inconsistent naming might be a better than a breaking change for this case, so leave them as they are

KILLEliteMaste (approved):
LGTM!

PR #922, opened 2021-12-12
Update log4j-api and recommended log4j-core version

Bastian:
See https://github.com/advisories/GHSA-jfh8-c2jp-5v3q

While Javacord itself is not affected by the security vulnerability (since we are only using the api and have our own "fallback" logger implementation), the README recommended Log4J with a vulnerable version.

This PRupdates the README to recommend the use of a patched version. It also updates the log4j-api dependency to the latest version just to make sure that there aren't any version conflicts.

### DIFF
--- a/.github/README_template.md
+++ b/.github/README_template.md
@@ -121,7 +121,7 @@ log levels per class, and much more.
 
 For example, Log4j Core in Gradle
 ```groovy
-dependencies { runtimeOnly 'org.apache.logging.log4j:log4j-core:2.11.0' }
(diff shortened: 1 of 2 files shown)


Vampire:
As you target the `development` branch it should be sufficient to update the readme template, not the actual readme. Noone looks at the readme of the development branch. :-D

Vampire:
Also we should go to 2.16.0

Bastian:
2.17.0 now 😉

Vampire (approved):
LGTM

PR #926, opened 2021-12-26
Fix NPE in SelectMenuInteractionImpl

KILLEliteMaste:
Not tested but it should be pretty straight forward. A link button does not have a custom id so a simple check whether the component is just a select menu by it's type id should fix this.


Vampire (approved):
LGTM

PR #929, opened 2021-12-30
Skip & log missing channels in Server voice_states

AyuAi:
I recently encountered an issue where the `voice_states` had channel ids that were not present in the `channels` property. This caused the `ServerImpl` to fail, throwing an assertion error, which rendered the entire guild unusable in Javacord.

This patch is what I used to fix the issue, which simply takes the assertion error and logs it to the logger. This allows the server to still create, it just doesn't process the unknown `voice_states`'s channel id.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
@@ -417,10 +417,21 @@ public class ServerImpl implements Server, Cleanupable, InternalServerAttachable
 
         if (data.hasNonNull("voice_states")) {
             for (JsonNode voiceStateJson : data.get("voice_states")) {
-                ServerVoiceChannelImpl channel =
-                        (ServerVoiceChannelImpl) getVoiceChannelById(voiceStateJson.get("channel_id").asLong())
-                                .orElseThrow(AssertionError::new);
-                channel.addConnectedUser(voiceStateJson.get("user_id").asLong());
+                Optional<ServerVoiceChannelImpl> channel =
+                        getVoiceChannelById(voiceStateJson.get("channel_id").asLong())
+                                .map(ch -> (ServerVoiceChannelImpl) ch);
+                // This gracefully disregards any channels in voice_states that are not present in the channels field.
+                // Bug occurred on a particular guild and prevented the guild from using the bot entirely.
+                // This log should happen almost never but should protect any guilds from failing creation due to
+                // mismatching guild data from Discord. https://github.com/discord/discord-api-docs/issues/4455
+                if (channel.isPresent()) {
+                    channel.get().addConnectedUser(voiceStateJson.get("user_id").asLong());
+                } else {
+                    logger.warn("Channel " + voiceStateJson.get("channel_id").asLong()
+                            + " was found in the voice_states property for server " + this.id
+                            + " but was not found in the channels property. It will not be"
+                            + " loaded.");
+                }
             }
         }
 


AyuAi:
For assurance sake, this is currently working at scale on https://github.com/AyuAi/Javacord/commit/e1f9eca42d46e297306e346786d6e440ac3832e1

Vampire:
Well, that's exception for flow control at its finest and will for sure not be merged in its current state. But I also wonder whether it is a good idea anyway. Warnings will typically just get ignored and that wouldn't be good.

AyuAi:
> Well, that's exception for flow control at its finest and will for sure not be merged in its current state. But I also wonder whether it is a good idea anyway. Warnings will typically just get ignored and that wouldn't be good.

Why is that not good? This is something that is bugged on Discords end, which means the dev cant do much about it to fix anything. The better solution would be to make the code more robust within Javacord, no?

Also, yes, the flow control being a trycatch is suspicious, lol. Primarily an example, I mainly want it to be a discussion. Definitely open to changing it to something cleaner

Vampire:
> Why is that not good? This is something that is bugged on Discords end, which means the dev cant do much about it to fix anything. The better solution would be to make the code more robust within Javacord, no?

Ah, ok, didn't get that. Then please also report an issue to Discord at https://github.com/discord/discord-api-docs if there is none yet. (API bugs get reported there) and add the link to the ticket in a comment that explains why that code is done and enables to check whether it can be removed again.

> Also, yes, the flow control being a trycatch is suspicious, lol. Primarily an example, I mainly want it to be a discussion. Definitely open to changing it to something cleaner

Then you should mention that, otherwise someone could simply merge it accidentally. :-P

Please replace it by a proper if-check instead of using `orElseThrow` and then right away catching the exception.

KILLEliteMaste:
Another solution to this would be to wait i.e. 5 -10 seconds and then try to refetch this guild via this endpoint https://discord.com/developers/docs/resources/guild#get-guild . 
When it comes to deciding what's better - to have an inconsistent but working state of a guild or waiting a bit and hopefully getting a valid guild JSON I think it might be better to wait. But this also has to be carefully tested before to see if some delay to refetching the guild really fixes this issue. For now I can only assume that for the time the JSON was generated by discord some sort of inconsistency was in their system and that it might have been fixed.

AyuAi:
> Please replace it by a proper if-check instead of using orElseThrow and then right away catching the exception.

I've gone ahead and cleaned up this code based on your suggestion. I also added a comment explaining its existence and a discord-api-docs issue I submitted.

I'm looking to have this merged at this point in time if it is approvable to prevent the issue from cropping up in the future.

Vampire:
Please don't do an `orElse(null)` just to check for `null` right after.
Use `Optional#isPresent` instead. :-)

AyuAi:
> Please don't do an `orElse(null)` just to check for `null` right after.
> Use `Optional#isPresent` instead. :-)

Updated to use an Optional

AyuAi:
> A second thought on that. It would be interesting to know if the additional voice states are actually correct. That means do the voice channels exist and are missing in the JSON or have the channels maybe been deleted with users connected to and discord may not update voice states in time?

@KILLEliteMaste I have no idea, honestly. I didn't really get the opportunity to thoroughly test.

@wasdennnoch mentioned to me that "it can happen that discord's guild cache gets out of sync; for instance, every now and then there are guilds that have a channel in their channel list that when trying to do anything with just 404s with "unknown channel". Discord generally tells server owners to contact support to request them to reload the guild cache."

If this is true, then I think that it should be addressed within the Javacord code so as to not ruin the entire guild on startup.

Vampire (approved):
LGTM, @KILLEliteMaste WDYT?

KILLEliteMaste (approved):
LGTM too

Vampire:
Can you please combine to one commit?

AyuAi:
Well that took me more effort than I anticipated lol

> Can you please combine to one commit and rebase?

👍

Vampire:
thanks :-)

Issue #935, opened 2022-01-14
SelectMenuChooseEvent doesn't implement Event


Lauriichan:
I noticed that the [SelectMenuChooseEvent](https://docs.javacord.org/api/v/3.3.2/org/javacord/api/event/interaction/SelectMenuChooseEvent.html) doesn't implement the Event class.
And since it is indeed a Event and it normally should.

PR #936, opened 2022-01-14
Add missing superclasses Event and ServerEvent ( Fixes #934 and #935 )

### LINKED ISSUE #934: ServerChangeRulesChannelEvent doesn't implement ServerEvent
I noticed that the [ServerChangeRulesChannelEvent](https://docs.javacord.org/api/v/3.3.2/org/javacord/api/event/server/ServerChangeRulesChannelEvent.html) doesn't implement the ServerEvent or Event class.
And since it is indeed a Event and it normally should.

Lauriichan:


Vampire (approved):
LGTM
Thanks for the contribution

PR #938, opened 2022-01-14
Add a Java9+ alternative for MessageSetImpl that works with Stream#takeWhile instead of Stream#filter

Vampire:


KILLEliteMaste (approved):
LGTM!

PR #944, opened 2022-01-20
Update dependencies

KILLEliteMaste:
For some reason `com.fasterxml.jackson.databind` is no longer a valid module name although jackson has a module named exactly like this.

I ran this change for 4h+and so far everything seems to be still working


Saladoc:
The suggestion [here](https://github.com/FasterXML/jackson-databind/issues/3380#issuecomment-1018129505) to use 2.12.6 for now would work. 
See https://github.com/FasterXML/jackson/issues/96 for further info / possible resolution in future releases.

feddevanderlist:
Well good and bad news.
Using okhttp3 version 4.8.1 everything works.
I am now trying to analyse what changed between kotlin 1.3.71 that could cause this.

PR #948, opened 2022-01-21
Add member timeouts

KILLEliteMaste:


Tigerpanzer02:
Just adding some information to that pr from the official discord documentation
https://dis.gd/timeout

javacord-core/src/main/java/org/javacord/core/entity/server/ServerUpdaterDelegateImpl.java
```diff
@@ -595,6 +610,13 @@ public void removeAllRolesFromUser(User user) {
                 }
             }
 
+            if (userTimeouts.containsKey(member)) {
+                String timeout = DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.from(ZoneOffset.UTC))
+                        .format(userTimeouts.get(member));
+                updateNode.put("communication_disabled_until",timeout);
```
Saladoc, line 616: According to the [docs](https://discord.com/developers/docs/resources/guild#modify-guild-member) the timeout, if to be removed, should be set to `null`. Even if we internally use `Instant.MIN` as a magic value, we should try to conform to discord's API to the best of our ability

javacord-api/src/main/java/org/javacord/api/listener/user/UserChangeTimeoutListener.java
```diff
@@ -0,0 +1,21 @@
...
+import org.javacord.api.event.user.UserChangeTimeoutEvent;
+import org.javacord.api.listener.GloballyAttachableListener;
+import org.javacord.api.listener.ObjectAttachableListener;
+import org.javacord.api.listener.server.ServerAttachableListener;
+
+/**
+ * This listener listens to user timeout changes.
+ */
+@FunctionalInterface
+public interface UserChangeTimeoutListener extends ServerAttachableListener, UserAttachableListener,
```
Saladoc, line 12: Does this also fire when a timeout expires or only if it is manually unset? Do we get gateway events for expiring timeouts?

javacord-api/src/main/java/org/javacord/api/entity/user/User.java
```diff
@@ -359,6 +359,72 @@ default String getDiscriminatedName() {
...
+        return server.timeoutUser(this, Instant.MIN, reason);
+    }
+
+    /**
+     * Gets the timestamp of when the user's timeout will expire
+     * and the user will be able to communicate in the server again.
+     * The returned Instant may be in the past which indicates that the user is not timed out.
+     *
+     * @param server The server to get the timeout for the user from.
+     * @return The timestamp of when this will no longer be timeout or an Instant in the past.
```
Saladoc, line 424:
```suggestion
     * @return The timestamp of when this user will no longer be timed out. Empty, if user is currently not timed out.
```

javacord-api/src/main/java/org/javacord/api/entity/server/ServerUpdater.java
```diff
@@ -579,6 +580,18 @@ public ServerUpdater setNickname(User user, String nickname) {
...
+     *
+     * @param user The user whose nickname should be updated.
+     * @param timeout The new timeout of the user.
+     * @return The current instance in order to chain call methods.
+     */
+    public ServerUpdater setUserTimeout(User user, Instant timeout) {
+        delegate.setUserTimeout(user, timeout);
+        return this;
+    }
+
```
Saladoc, line 594: Also include a method to queue the user's timeout to be unset.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -1519,6 +1529,62 @@ default ServerUpdater createUpdater() {
...
+     *
+     * @param user    The user.
+     * @param timeout The timestamp until the user should be timeout,
+     * @param reason  The audit log reason for this update.
+     * @return A future to check if the update was successful.
+     */
+    default CompletableFuture<Void> timeoutUser(User user, Instant timeout, String reason) {
+        return createUpdater().setUserTimeout(user, timeout).setAuditLogReason(reason).update();
+    }
+
```
Saladoc, line 1560: What do you think of also adding convenience methods with a `User, Duration` signature to make "timeout this user for 48 hours starting now" style calls easier?

```suggestion

    /**
     * Timeout the given user on this server.
     *
     * <p>If you want to update several settings at once, it's recommended to use the
     * {@link ServerUpdater} from {@link #createUpdater()} which provides a better performance!
     *
     * @param user    The user.
     * @param duration The duration how long the user should be timed out.
     * @param reason  The audit log reason for this update.
     * @return A future to check if the update was successful.
     */
    default CompletableFuture<Void> timeoutUser(User user, Duration timeout, String reason) {
        return createUpdater().setUserTimeout(user, Instant.now().plus(duration)).setAuditLogReason(reason).update();
    }
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -1519,6 +1529,62 @@ default ServerUpdater createUpdater() {
...
     }
 
+    /**
+     * Timeout the given user on this server.
+     *
+     * <p>If you want to update several settings at once, it's recommended to use the
+     * {@link ServerUpdater} from {@link #createUpdater()} which provides a better performance!
+     *
+     * @param user    The user.
+     * @param timeout The timestamp until the user should be timeout,
```
Saladoc, line 1539:
```suggestion
     * @param timeout The timestamp until the user should be timed out,
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -1519,6 +1529,62 @@ default ServerUpdater createUpdater() {
...
+    }
+
+    /**
+     * Timeout the given user on this server.
+     *
+     * <p>If you want to update several settings at once, it's recommended to use the
+     * {@link ServerUpdater} from {@link #createUpdater()} which provides a better performance!
+     *
+     * @param user    The user.
+     * @param timeout The timestamp until the user should be timeout,
```
Saladoc, line 1553:
```suggestion
     * @param timeout The timestamp until the user should be timed out,
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -181,6 +181,16 @@
      */
     Optional<String> getNickname(User user);
 
+    /**
+     * Gets the timestamp of when the user's timeout will expire
+     * and the user will be able to communicate in the server again.
+     * The returned Instant may be in the past which indicates that the user is not timed out.
```
Saladoc, line 187: If the user is not timed out, the returned optional should be empty.

javacord-core/src/main/java/org/javacord/core/entity/user/Member.java
```diff
@@ -133,4 +132,13 @@ default String getDisplayName() {
      */
     boolean isPending();
 
+    /**
+     * Gets the timestamp of the timeout when the timeout will expire
+     * and the user will be able to communicate in the server again.
+     * The returned Instant may be in the past which indicates that the user is not timed out.
+     *
+     * @return The timestamp of when this will no longer be timeout or an Instant in the past.
```
Saladoc, line 140:
```suggestion
     * @return The timestamp of when this member will no longer be timed out.
```

javacord-core/src/main/java/org/javacord/core/entity/user/MemberImpl.java
```diff
@@ -300,6 +312,11 @@ public boolean isPending() {
         return pending;
     }
 
+    @Override
+    public Optional<Instant> getTimeout() {
+        return Optional.ofNullable(communicationDisabledUntil);
```
Saladoc, line 317: If the timeout is expired, the returned optional should be empty
```suggestion
        return Optional.ofNullable(communicationDisabledUntil).filter(Instant.now()::isBefore);
```

Saladoc (changes requested):
Looks good overall, but I've marked some spots that seemed odd to me. 
Some suggestions I made would need to be applied in more than one place.

javacord-api/src/main/java/org/javacord/api/listener/user/UserChangeTimeoutListener.java
KILLEliteMaste: It doesn't. Seems to fire only when its manually changed

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
KILLEliteMaste: Should this be added to the Javadoc? Or do you mean to change the behavior?

Saladoc: Both. The behavior is changed in another suggestion, but the javadoc should also be changed accordingly. I feel that if you get a non-empty optional the question should not be whether but rather how long the user is timed out.

javacord-core/src/main/java/org/javacord/core/entity/user/MemberImpl.java
Saladoc: I see, that's a legit case too. Maybe two functions? `getTimeout()` and `getActiveTimeout()`?

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
KILLEliteMaste: Personally I think the use case would be minimal and just clutter the API. I can imagine the use case of timeouts would be to integrate it into moderation bots where mods can timeout a user with a command `/timeout Domme 12` and then the convenience methods are useless. And the time range is just so huge (+28days) so what would be the most used times to make convenience methods for?

bassner: I think Saladocs suggestion (the code above) is a good idea. With your current implementation, you need to calculate the target instant / timestamp yourself.
With the suggest additional method, you can go like
```java
server.timeoutUser(user, Duration.ofHours(12));
```

KILLEliteMaste: Oh, I totally misunderstood what he meant. Indeed, this is a good idea

bassner (approved):
Code LGTM! 👍

javacord-api/src/main/java/org/javacord/api/entity/server/ServerUpdater.java
```diff
@@ -579,6 +581,40 @@ public ServerUpdater setNickname(User user, String nickname) {
...
+    /**
+     * Queues a user's timeout to be updated.
+     *
+     * @param user The user whose nickname should be updated.
+     * @param duration The duration of the timeout.
+     * @return The current instance in order to chain call methods.
+     */
+    public ServerUpdater setUserTimeout(User user, Duration duration) {
+        return setUserTimeout(user, Instant.now().plusSeconds(duration.getSeconds()));
```
Saladoc, line 604: Is there a particular reason to not just use `Instant.now().plus(duration)`?

javacord-api/src/main/java/org/javacord/api/entity/server/internal/ServerUpdaterDelegate.java
```diff
@@ -368,6 +369,14 @@
...
 
+    /**
+     * Queues a user's timeout to be updated.
+     *
+     * @param user The user whose nickname should be updated.
+     * @param timeout The new timeout of the user.
+     */
+    void setUserTimeout(User user, Instant timeout);
+
     /**
```
Saladoc, line 380: Should we also have the `setUserTimeout(User, Duration)` method here?

Saladoc (approved):
LGTM overall. Two quick questions, but if you don't think they warrant further changes we can skip them and merge.

javacord-api/src/main/java/org/javacord/api/entity/server/internal/ServerUpdaterDelegate.java
KILLEliteMaste: You have missed it, I added it with the previous changes too

javacord-api/src/main/java/org/javacord/api/entity/server/ServerUpdater.java
KILLEliteMaste: When I tried what you did, it was red underlined by the IDE. Now when I tried it again it works. Maybe it was because at that moment my system was under heavy load and it took way to long to visualize it as valid

javacord-api/src/main/java/org/javacord/api/entity/server/internal/ServerUpdaterDelegate.java
Saladoc: ah yeah nevermind it should not be on the delegate as that's internal - I didn't properly read the class name.

PR #957, opened 2022-01-29
Fix `BotInviteBuilder` shortcomings with additional features.

### LINKED ISSUE #908: `BotInviteBuilder` shortcomings
The `BotInviteBuilder` should allow to select for which scopes an invite link should be generated and whether a link for adding the bot or for just adding a scope should be generated.

ShindouMihou:
This is sort of an entire revamp of the BotInviteBuilder that allows the addition of `BotInviteScope`, `Prompt Consent`, and `Redirect URIs` for completeness. This also adds a few more conveience methods such as `addScopes`, `addPermissions`, and so forth with an extra builder for `DiscordApi`.

> 🟥 I think this change would be marked as "breaking" because the `BASE_LINK` was changed to utilize a StringBuilder approach into building the links. I could use `String#Format` plus concatenation but considering how this is 3.5.0, I was thinking of just going with this.

> 🟨  This also follows the suggestion of which it shouldn't throw an exception when **REDIRECT_URI** is not set while using specific scopes that require the **REDIRECT_URI** to be set.

I also have no idea where to place the Enum for `BotInviteScope` but since the other common enums were found in the `api.entity` package, I decided to follow that instead. Feel free to correct me if I need to place it somewhere else instead.

✅  This commit has been tested for all functionality, thoroughly.
🧷 Fixes #908

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/BotInviteBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/BotInviteBuilder.java
@@ -1,7 +1,15 @@
 package org.javacord.api;
 
+import org.javacord.api.entity.permission.PermissionType;
 import org.javacord.api.entity.permission.Permissions;
 import org.javacord.api.entity.permission.PermissionsBuilder;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.stream.Collectors;
 
 /**
  * A tool to create bot invites.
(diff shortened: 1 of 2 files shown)


Saladoc (changes requested):
I don't think this is a breaking change as all existing api will continue to work without any difference in observable behavior (as evidenced by the fact that you did not have to change the `DiscordApi#createBotInvite` methods. 

The only change I want to request is that you move the enum, as it does not pertain to any discord entity. If no other suitable package exists, either create a dedicated  package and move both the enum and the builder there or leave both in `org.javacord.api`.

Currently I'm in favor of just leaving them in `org.javacord.api` since I'm unhappy with any dedicated package name I can think of.

Thank you for this quality contribution.

ShindouMihou:
> I don't think this is a breaking change as all existing api will continue to work without any difference in observable behavior (as evidenced by the fact that you did not have to change the `DiscordApi#createBotInvite` methods.

I was more looking into how changing the `BASE_LINK`'s value might affect some people using it since it is sort of a public static variable, although then again, the amount of people using such is most likely down to a few fingers or even none at that which makes me think this should be sort of safe to do(?)

Thanks for the review though, I've moved the enum as requested and squashed the commits already. 👍🏻

Saladoc (approved):
LGTM, but I'm waiting for someone smarter than me to confirm this needs to be listed as a breaking change. Won't affect the release it'll be in but be relevant for the patch notes.

KILLEliteMaste:
Imo this is a breaking change because you changed the BASE_LINK. It is publicly available in the API and therefore if someone uses this it might not work after merging this PR.
Another thing: Something doesn't match anymore. Either It is a `BotInviteBuilder` or just a `LinkBuilder`. They would differentiate in the `BotInviteBuilder` having the `bot` scope by default. Is there actually any other use case for this builder except for creating a bot invite?

ShindouMihou:
> They would differentiate in the `BotInviteBuilder` having the `bot` scope by default.

It does use `bot` and `application. commands` by default.

> Imo this is a breaking change because you changed the BASE_LINK.

It is kind of impossible to not make this a breaking one imo since the bot scope was hardcoded. I was thinking earlier to just remove any queries pre-built and make the static variable closer to a BASE_LINK than one filled with # queries beforehand.

> Is there actually any other use case for this builder except for creating a bot invite?

The builder can still be used for bots, for instance, my bot uses the `redirect_uri` and identify to send the user to a location that simply redirects them to the wiki after inviting with `application_commands` and `bot` scope.

The original BotInviteBuilder doesn't have any form of support for additional scopes which made it incompatible with making bot invite links for bots with slash commands since the `application_commands` scope is kind of a requirement.

But overall, this is still a bot invite builder since it defaults to two scopes `bot` and `application_commands` unless overridden although the defaulting happens in the wrong part of the world (should happen during variable initialization).

Feel free to correct me if I am misunderstanding any of the questions.

Saladoc:
Alright, then this just needs a quick rebase and then will be good to merge.

ShindouMihou:
> Alright, then this just needs a quick rebase and then will be good to merge.

Rebased 👍🏻

PR #963, opened 2022-02-16
Do not return a message from cache if it was reqeusted from a channel by ID if the message belongs to a different channel

Vampire:


KILLEliteMaste (changes requested):
I think this is a breaking change because it changes the behavior of the method?

And the Javadoc of this and all dependent methods (like in DiscordApi#getMessageById(long, TextChannel)) should be changed to include something like
```
Gets a message by its id.
Only returns a message, if the message belongs to this channel.
```

Vampire:
JavaDoc is added.

Regaring breaking, that's probably a matter of PoV.
It imho was a faulty behavior that the message was returned.
It is like an `appleTree.getFruit("green")` that returns you a green pear.
On the other hand it is a change in behavior of that method.
I actually don't care much whether we list it as breaking or not.

Vampire:
JavaDoc is added.

Regaring breaking, that's probably a matter of PoV.
It imho was a faulty behavior that the message was returned.
It is like an `appleTree.getFruit("green")` that returns you a green pear.
On the other hand it is a change in behavior of that method.
I actually don't care much whether we list it as breaking or not.

KILLEliteMaste (approved):
LGTM!

PR #615, opened 2020-09-23
[WIP] Add Support for teams

Joshix-1:
Not tested. But should function properly. I want to wait for the next version and feedback to improve this. 
Based on #614

PR #968, opened 2022-03-07
Add teams

Saladoc:
This is #615 updated and expanded. 

Probably the most controversial change is how `getOwner()` and `getOwnerId()` are handled. No existing behavior is broken, however:

If an application (like our bot) is owned by a Team, the ownerId field will be the team ID for the owner, there may be a pseudo user with the team id, the username `team[teamid]` etc. 

To preserve api stability, if a team is the owner of the application, the `getOwner()` and `getOwnerId`  methods will return the owner of the team. This way, we don't need to break `getOwner()` and make it return `Optional<User>`.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -1,6 +1,7 @@
 package org.javacord.api;
 
 import org.javacord.api.entity.ApplicationInfo;
+import org.javacord.api.entity.ApplicationOwner;
 import org.javacord.api.entity.Icon;
 import org.javacord.api.entity.activity.Activity;
 import org.javacord.api.entity.activity.ActivityType;
@@ -536,40 +537,56 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
      * Gets the id of the application's owner.
      *
      * @return The id of the application's owner.
+     * @see ApplicationInfo#getOwner()
+     * @see ApplicationOwner#getId()
      */
-    long getOwnerId();
+    default Optional<Long> getOwnerId() {
+        ApplicationInfo applicationInfo = getCachedApplicationInfo();
+        return applicationInfo.getOwner()
+                .map(ApplicationOwner::getId);
+    }
 
     /**
      * Gets the owner of the application.
      *
-     * @return The owner of the application.
+     * @return The owner of the application, if applicable.
+     * @see ApplicationInfo#getOwner()
+     * @see ApplicationOwner#requestUser()
      */
-    default CompletableFuture<User> getOwner() {
-        return getUserById(getOwnerId());
+    default Optional<CompletableFuture<User>> getOwner() {
+        return getCachedApplicationInfo().getOwner()
+                .map(ApplicationOwner::requestUser);
     }
 
     /**
      * Gets the team of the application.
      *
-     * @return The team of the application.
+     * @return The team of the application, if applicable.
+     * @see ApplicationInfo#getTeam()
      */
-    Optional<Team> getCachedTeam();
+    default Optional<Team> getCachedTeam() {
+        return getCachedApplicationInfo().getTeam();
+    }
 
     /**
      * Requests the team of the application.
      *
      * @return The team of the application.
+     * @see ApplicationInfo#getTeam()
      */
     default CompletableFuture<Optional<Team>> requestTeam() {
-        return getApplicationInfo().thenApply(ApplicationInfo::getTeam);
+        return requestApplicationInfo().thenApply(ApplicationInfo::getTeam);
     }
 
     /**
      * Gets the client id of the application.
      *
      * @return The client id of the application.
+     * @see ApplicationInfo#getClientId()
      */
-    long getClientId();
+    default long getClientId() {
+        return getCachedApplicationInfo().getClientId();
+    }
 
     /**
      * Disconnects the bot.
@@ -623,12 +640,22 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
     int getReconnectDelay(int attempt);
 
     /**
+     * Gets the cached application info of the bot.
+     *
+     * @return The cached application info.
+     * @see #requestApplicationInfo()
+     */
+    ApplicationInfo getCachedApplicationInfo();
+
+    /**
      * Gets the application info of the bot.
-     * The method only works for bot accounts.
+     *
+     * <p>This will always request the up to date application info and update the cache.</p>
      *
      * @return The application info of the bot.
+     * @see #getCachedApplicationInfo()
      */
-    CompletableFuture<ApplicationInfo> getApplicationInfo();
+    CompletableFuture<ApplicationInfo> requestApplicationInfo();
 
     /**
      * Gets a webhook by its id.
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -29,6 +29,7 @@ import org.javacord.api.entity.server.ServerBuilder;
 import org.javacord.api.entity.server.invite.Invite;
 import org.javacord.api.entity.sticker.Sticker;
 import org.javacord.api.entity.sticker.StickerPack;
+import org.javacord.api.entity.team.Team;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.entity.user.UserStatus;
 import org.javacord.api.entity.webhook.IncomingWebhook;
@@ -545,6 +546,22 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
      */
     default CompletableFuture<User> getOwner() {
         return getUserById(getOwnerId());
+    }
+
+    /**
+     * Gets the team of the application.
+     *
+     * @return The team of the application.
+     */
+    Optional<Team> getCachedTeam();
+
+    /**
+     * Requests the team of the application.
+     *
+     * @return The team of the application.
+     */
+    default CompletableFuture<Optional<Team>> requestTeam() {
+        return getApplicationInfo().thenApply(ApplicationInfo::getTeam);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -1864,7 +1872,12 @@ public User getYourself() {
 
     @Override
     public long getOwnerId() {
-        return ownerId;
+        return team != null ? ownerId : team.getOwnerId();
```
KILLEliteMaste, line 1875: `applicationInfo.flatMap(ApplicationInfo::getTeam).map(Team::getOwnerId).orElseGet(() -> ownerId)`

javacord-core/src/main/java/org/javacord/core/entity/team/TeamImpl.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    private final String name;
+
+    /**
+     * The members of the team.
+     */
+    private final List<TeamMember> members = new LinkedList<>();
+
+    /**
+     * Creates a new user.
```
KILLEliteMaste, line 60:
```suggestion
     * Creates a new Team.

```

javacord-core/src/main/java/org/javacord/core/entity/team/TeamImpl.java
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @param api The discord api instance.
+     * @param data The json data of the user.
```
KILLEliteMaste, line 63:
```suggestion
     * @param data The json data of the Team.

```

javacord-core/src/main/java/org/javacord/core/entity/activity/ApplicationInfoImpl.java
```diff
@@ -37,8 +42,9 @@ public ApplicationInfoImpl(DiscordApi api, JsonNode data) {
         description = data.get("description").asText();
         publicBot = data.get("bot_public").asBoolean();
         botRequiresCodeGrant = data.get("bot_require_code_grant").asBoolean();
-        ownerId = data.get("owner").get("id").asLong();
-        ownerName = data.get("owner").get("username").asText();
+        team = data.hasNonNull("team") ? new TeamImpl((DiscordApiImpl) api, data.get("team")) : null;
+        ownerId = (team == null) ? data.get("owner").get("id").asLong() : team.getOwnerId();
+        ownerName = (team == null) ? data.get("owner").get("username").asText() : team.getName();
         ownerDiscriminator = data.get("owner").get("discriminator").asText();
```
KILLEliteMaste, line 48: Imo we should change the datatype to return an Optional. There should be either a user owner or a team owner. Mixing these fields is not a good idea as it also somewhat doesn't make sense to return the team name as the owner name or a weird self created owner discriminator.

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -516,19 +517,39 @@ default String createBotInvite(Permissions permissions) {
...
+
+    /**
+     * Requests the team of the application.
+     *
+     * @return The team of the application.
+     */
+    default CompletableFuture<Optional<Team>> requestTeam() {
+        return getApplicationInfo().thenApply(ApplicationInfo::getTeam);
+    }
+
```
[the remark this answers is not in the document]
Saladoc: So do we remove the `getOwnerId()` and `getOwner()` methods from Discord Api as well? because that too is part of the Application Info.
The main problem here is that before, an applications owner was static. Now, applications can be transferred from single users to teams and the team members can change. The application info is requested once on startup and then cached. If `getApplicationInfo()` is called, the cache is updated. For a long time, that only was relevant for when the bot description changed or the application name, whether the bot was public etc. 

I'll think this over once more. Obviously, owner info and team info are the most relevant parts of the ApplicationInfo when it comes to bots since they might be relevant for permissions etc so if it is feasible, they should be made available directly from the api object. But we also shouldn't duplicate information all over the place.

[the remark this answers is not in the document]
KILLEliteMaste: Or to be more precise I basically wonder why we don't provide a cached ApplicationInfo rather than the cached team. Currently we only have a method to request it, but why shouldn't we allow the user to access a cached instance? Doing this would also save us 2 variables in the `DiscordApiImpl` which then are no longer needed anymore. And this would also automatically update everything dependent like the team without needing to assign it again.

KILLEliteMaste:
What's the state of this? As far as I see only the last requested change needs to be done and then it can be merged

Saladoc:
Alright, time to get this under wraps. Here's the short notes:
 - Previous owner info has been encapsulated in a separate class
 - Previous methods have been kept (some with changed signatures -> breaking change), deprecated and marked for removal in the subsequent release
 - Either `Owner` or `Team` is present on an Application Info instance. Both are `Optional<>`

javacord-core/src/main/java/org/javacord/core/entity/team/TeamImpl.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Creates a new team.
+     *
+     * @param api The discord api instance.
+     * @param data The json data of the team.
+     */
+    public TeamImpl(DiscordApiImpl api, JsonNode data) {
+        this.api = api;
+
+        id = Long.parseLong(data.get("id").asText());
```
KILLEliteMaste, line 67:
```suggestion
        id = data.get("id").asLong();

```

javacord-core/src/main/java/org/javacord/core/entity/team/TeamImpl.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+        iconHash = data.has("icon") ? data.get("icon").asText() : null;
```
KILLEliteMaste, line 70:
```suggestion
        iconHash = data.path("icon").asText(null);

```

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
```diff
@@ -650,14 +646,14 @@ public DiscordApiImpl(
...
-                            getApplicationInfo().whenComplete((applicationInfo, exception) -> {
+                            requestApplicationInfo().whenComplete((applicationInfo, exception) -> {
                                 if (exception != null) {
                                     logger.error("Could not access self application info on startup!", exception);
+                                    threadPool.shutdown();
+                                    ready.completeExceptionally(exception);
                                 } else {
-                                    clientId = applicationInfo.getClientId();
-                                    ownerId = applicationInfo.getOwnerId();
+                                    ready.complete(this);
```
KILLEliteMaste, line 655: Shouldn't you assign the `ApplicationInfo` in here?

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -537,24 +540,64 @@ default String createBotInvite(Permissions permissions) {
...
      *
-     * @return The owner of the application.
+     * <p>If the application does not have an owner, for example if it is owned by a team,
+     * the future completes with an empty Optional.</p>
+     *
+     * @return The owner of the application, if applicable.
+     * @see ApplicationInfo#getOwner()
+     * @see ApplicationOwner#requestAsUser()
+     */
+    default CompletableFuture<Optional<User>> getOwner() {
```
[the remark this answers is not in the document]
Saladoc: The alternative I first went with, but then never commited was returning `CompletableFuture<User>` and returning a failed future with `NoSuchElementException` if there was no owner. Would you prefer that? Personally I found it more important to encode the possiblilty that there might not be an owner.

The best way here would probably be to mimic `EmbedAuthor#downloadIconAsBufferedImage` and return `Optional<Completablefuture<User>>` to drive home the point that there might be situations where there is no chance of obtaining a user at all.

javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
Saladoc: That *could* prove helpful, yes.

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -537,24 +540,57 @@ default String createBotInvite(Permissions permissions) {
      * Gets the id of the application's owner.
      *
      * @return The id of the application's owner.
+     * @see ApplicationInfo#getOwner()
+     * @see ApplicationOwner#getId()
      */
-    long getOwnerId();
+    default OptionalLong getOwnerId() {
```
Mysterypotatoguy, line 546: We don't really have precedent for using `OptionalLong` over `Optional<Long>` anywhere else, is this something we want to start doing?

Saladoc: Probably not. If we do, we'll want to change it everywhere but that's out of scope for this PR. I honestly forgot we had `Optional<Long>` elsewhere.

Mysterypotatoguy (approved):

javacord-core/src/main/java/org/javacord/core/entity/team/TeamImpl.java
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @param api The discord api instance.
+     * @param data The json data of the team.
+     */
+    public TeamImpl(DiscordApiImpl api, JsonNode data) {
+        this.api = api;
+
+        id = data.get("id").asLong();
+
+
```
KILLEliteMaste, line 69: Remove lines

KILLEliteMaste (changes requested):
Otherwise it LGTM! Rebase + squashing commits.

KILLEliteMaste (approved):

Saladoc:
@all-contributors please add @Joshix-1 for code

KILLEliteMaste:
@all-contributors please add @Joshix-1 for code

PR #972, opened 2022-03-08
Let interactions and webhooks not consume the global ratelimit

bassner:
Webhooks and interactions work indepedently of the bot account. Therefore, it's not required, but rather contraproductive to let webhook and interaction requests consume a quota from the global ratelimiter that are then not available for actual bot requests, even though they aren't really consuming a quota on Discord's side.

For the same reason, it's not required to add the general bot authorization header to these requests.

Yunite runs on this patch for a few weeks now without any issues (even though it is not using actual webhooks anywhere, only interactions; it's using follow up messages though, which go through the WEBHOOK_SEND endpoint).


KILLEliteMaste (commented):
Is there any reason why `InteractionImpl#respondWithModal` does not ignore the global ratelimit?

bassner:
Nope, I think I just overlooked it. Thanks! @KILLEliteMaste 
Added in c5ac5af

Saladoc (approved):

PR #974, opened 2022-03-14
Adjust collection (and subtypes) method return type to the correct ones in the API

KILLEliteMaste:


Saladoc (changes requested):
Looks good.

However, I'm not a fan of the documentation stating the collection type over and over again when it's clear from the return type. As far as I am concerned, all javadoc comments can be changed from `(Gets )? a (set|list|collection) with XY` to just `XY` unless we provide additional beneficial information, like the ordering of a list.

Saladoc (approved):

PR #976, opened 2022-03-14
Have git ignore OS X system files

viennaltran:
.DS_Store is added to .gitignore because .DS_Store files are automatically generated on OS X and should not be checked in.
https://en.wikipedia.org/wiki/.DS_Store

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@
 *.iml
 *.ipr
 *.iws
+
+# System files
+.DS_Store
+


Saladoc (approved):

Vampire:
@Saladoc why did you merge this?
I strongly vote to revert this commit.

System-dependent files have nothing to do in a project-specific gitignore file.
Next person comes and asks to add `*.bak` to gitignore because he is using a editor generating such files as backup.
Then the next person comes and asks to add `*~` for the same reason.

Even the IDE specific files are questionable to have in a project specific gitignore file.
But in case of IDE specific files it is actually project-specific whether those files are kept in VCS or not so it is ok.

For system-dependent things like backup files of editors, or files a specific OS is generating and so on users can either use the project specific `.git/info/exclude` file or the global `~/.config/git/ignore` (or whatever the Git config `core.excludesFile` points at) file.

Saladoc:
I wasn't aware that there's such a strong hierarchy to how gitignore files should be used.

Vampire:
Well, it's not a hard rule, just what makes sense.
OS-specific files are local to the developer, so the developer who has them should ignore them.
Anything that is generated by the build or is project-specific should be in the project-specific ignore file imho.
That's why there are different levels of gitignore files after all. :-)

PR #981, opened 2022-03-21
Add a check if the `nsfw` field is present for ChannelCategory updates

KILLEliteMaste:


javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -302,7 +302,7 @@ private void handleChannelCategory(JsonNode jsonChannel) {
         long channelCategoryId = jsonChannel.get("id").asLong();
         api.getChannelCategoryById(channelCategoryId).map(ChannelCategoryImpl.class::cast).ifPresent(channel -> {
             boolean oldNsfwFlag = channel.isNsfw();
-            boolean newNsfwFlag = jsonChannel.get("nsfw").asBoolean();
+            boolean newNsfwFlag = jsonChannel.hasNonNull("nsfw") && jsonChannel.get("nsfw").asBoolean();
```
Saladoc, line 305: Shouldn't the default value overload take care of this?

```suggestion
            boolean newNsfwFlag = jsonChannel.get("nsfw").asBoolean(false);
```

KILLEliteMaste: This would require that the `get` was actually sucessfull. The default value is just for the case the the value node is i.e. of another type iirc

Saladoc: Ah right, I mixed up `get` and `path`  - `jsonChannel.path("nsfw").asBoolean(false)` should work.

KILLEliteMaste: Oh, I did not know that this exists. So far the project always uses the way like seen in this PR. Was there maybe a reason to avoid this method?

itsTyrion: If I had to guess, either not known or that it's more obvious/less obfuscated what it's doing

Issue #647, opened 2020-10-07
Message cache issue when size is set to 0


KILLEliteMaste:
Altough the message cache size is set to 0 this warning gets thrown sometimes
`[WARN ] 2020-10-06 03:31:24 [Javacord - Central Scheduler - 1] MessageCacheImpl - Heap memory was too low to hold all configured messages in the cache. Removed 15 messages from the cache due to memory shortage. Either increase your heap settings or decrease your message cache settings!`

The day before and after (so far) didn't have a single warning
https://pastebin.com/eQMLbqrt

KILLEliteMaste:
A while ago there was a discussion in our Discord where another problem is related to this issue. Multiple users experiences this. The issue is that even though the messsage cache is set, for some channels the cleanup just doesn't work anymore after some time and the channels message cache gets filled up.
![image](https://user-images.githubusercontent.com/31566807/163115400-2abf7b7a-44a2-449e-9553-a03f4ff9eaa8.png)

KILLEliteMaste:
Maybe related to #359 ?

PR #984, opened 2022-04-12
make the adding of all messages to the cache optional

twonirwana:
This allows it to disable the adding of all messages to the cache and thereby reducing the resources.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -834,6 +834,13 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
     }
 
     /**
+     * Checks if the add all messages to cache is enabled.
+     *
+     * @return Whether the add all messages to cache is enabled;
+     */
+    boolean isAddAllMessageToCacheEnabled();
+
+    /**
      * Gets a collection with all currently cached users.
      *
      * @return A collection with all currently cached users.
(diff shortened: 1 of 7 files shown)


KILLEliteMaste:
Isn't this the effectively the same as `DiscordApi#setMessageCacheSize(0,0)` ? Even if they are cached for 1 minute as the [Javadoc](https://github.com/Javacord/Javacord/blob/754e7d7f6dad443ea48827c30521d5e0ee61b46c/javacord-api/src/main/java/org/javacord/api/DiscordApi.java#L393-L403) states, I think that isn't that bad?

twonirwana:
My bot is not that popular (on around 300 servers) and even with ` setMessageCacheSize(0,0)` but i have up to 1000 messages in the cache and get ` Heap memory was too low to hold all configured messages in the cache.` messages (even with 5gig of heap). I did not look at the threads but i think all the cleanup schedules will also be quite active.

KILLEliteMaste:
There seems to be an issue with this message see https://github.com/Javacord/Javacord/issues/647 so this message is not 100% reliable as far as I have witnessed. Imo the best would be that the issue gets fixed instead of adding a workaround as this seems 
 to be when looking at the linked issue.

twonirwana:
`Heap memory was too low to hold all configured messages in the cache.` was only the reason i looked into the caching. I think adding hundreds of messages to a cache only to remove all of them 60s later is not a optimal usage of resources. If you are interested i could also create a pull request where i replace your self written caching with a Guava Cache, that should be more reliable.

PR #986, opened 2022-04-25
Add a toggle to change if events can be dispatched or not

KILLEliteMaste:
This PR suggests to add a toggle which allows the user to turn the dispatch for events on/off. 
The intention of this is to be able to get closer to a 100% uptime of a bot which is currently not (easily) posible with the time a bot takes to start. In addition to that it becomes even harder when events gets dispatched and they may be handled twice when having a second bot process running.
For example a CD workflow:
1. Start the 2nd process with dispatchEvents turned off
2. Wait until the bot is fully logged in
3. Switch the toggle dispatch on both running bot processes
4. Stop the old running process

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
@@ -191,6 +191,10 @@ public abstract class EventDispatcherBase {
      * @param <T>           The type of the listener.
      */
     protected <T> void dispatchEvent(DispatchQueueSelector queueSelector, List<T> listeners, Consumer<T> consumer) {
+        if (!api.canDispatchEvents()) {
+            return;
+        }
+
         api.getThreadPool().getSingleThreadExecutorService("Event Dispatch Queues Manager").submit(() -> {
             if (queueSelector != null) { // Object dependent listeners
                 // Don't allow adding of more events while there are unfinished object independent tasks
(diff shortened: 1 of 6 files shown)


bassner:
Hm. Sounds like you could instead just *not* attach your listeners until you want to receive events? :)

KILLEliteMaste:
Yes but then there's also the need to disable the event dispatching. You can also do it with removing the listeners but this is a lot trickier because i.e. `api.getListeners()` does not return listeners that are attached directly to an entity i.e. `ServerTextChannel`. 
So essentially this PR suggest to add a convenience feature for toggleable event dispatching.

bassner:
I see
Yeah I mean I don't have objections against this as long as event dispatching remains enabled by default

KILLEliteMaste:
@bassner did you have a look at the PR yet? If so please approve :)

javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
```diff
@@ -191,24 +191,26 @@ public void setExecutionTimeCheckingEnabled(boolean enable) {
...
-                    try {
-                        synchronized (queuedListenerTasks) {
-                            // Just to be on the safe side, we use a timeout of 5 seconds
-                            queuedListenerTasks.wait(5000);
-                        }
-                    } catch (InterruptedException ignored) { }
+        if (api.canDispatchEvents()) {
+            api.getThreadPool().getSingleThreadExecutorService("Event Dispatch Queues Manager").submit(() -> {
+                if (queueSelector != null) { // Object dependent listeners
+                    // Don't allow adding of more events while there are unfinished object independent tasks
```
bassner, line 197: I suggest to return early to avoid indenting everything below:
```java
if (!api.canDispatchEvents()) {
    return;
}
```

bassner (approved):
LGTM 👍

PR #993, opened 2022-05-06
Add bans pagination

bassner:
Discord randomly added paging to the bans endpoint: https://discord.com/developers/docs/change-log#mar-31-2022

This is causing that Javacord doesn't return more than 1000 bans at the moment. This is only a problem for large guilds.

This PR fetches all bans again (not a breaking change) by just fetching all available pages. There are rarely usecases where you would only want to fetch like the first one- or two thousand of bans, so the method has been set up to always fetch all bans. Individual bans can still be fetched using ``requestBan()``.

I've successfully tested this patch on a server with about 2.5k bans.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
@@ -73,6 +73,7 @@ import org.javacord.core.util.rest.RestEndpoint;
 import org.javacord.core.util.rest.RestMethod;
 import org.javacord.core.util.rest.RestRequest;
 import org.javacord.core.util.rest.RestRequestResult;
+
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.Instant;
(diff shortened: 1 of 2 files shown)


javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1556,15 +1557,49 @@ public CompletableFuture<Ban> requestBan(long userId) {
 
     @Override
     public CompletableFuture<Collection<Ban>> getBans() {
```
KILLEliteMaste, line 1559: The Javadoc should inform the user that this method takes up to `1000 bans / total server bans` requests just in case there are larger bots calling this method frequently to prevent unnecessary debugging why they may run into rate limits :)

Mysterypotatoguy:
Perhaps also a `getBans(int limit, Long after)` for manual pagination

bassner:
@KILLEliteMaste @Mysterypotatoguy I've added your suggestions in 8d565c0 :)

KILLEliteMaste (approved):
LGTM!

PR #994, opened 2022-05-13
Updated Discord api version

### LINKED ISSUE #960: Upgrade to gateway version `10`
https://github.com/discord/discord-api-docs/discussions/4510

RealYusufIsmail:
I have updated the discord api version to 10

closes #960

TODO:
- [x] tweak identifying payload
- [x] add all the changes for the new version.


KILLEliteMaste (commented):
1. Your change does not work, it should be `10`
2. You have not done any changes mentioned in the changelog of the v10. The link is in the issue.

RealYusufIsmail:
oh ok

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1564,6 +1571,9 @@ public CompletableFuture<Collection<Ban>> getBans(Integer limit, Long after) {
         ).setUrlParameters(getIdAsString());
 
         if (limit != null) {
+            if (limit < 0 || limit > 1000) {
+                request.addQueryParameter("limit", "1000");
+            }
```
KILLEliteMaste, line 1576: Why?

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1555,6 +1555,13 @@ public CompletableFuture<Ban> requestBan(long userId) {
                 .execute(result -> new BanImpl(this, result.getJsonBody()));
     }
 
+    /**
+     * This method is used the get the {@link Member} who are banned from the server.
+     * @param limit how many bans should be returned at most. Must be within [0, 1000]. If null, it will default to 1000
+     * @param after should be a snowflake to only take bans of users with IDs higher
+     *              than this parameter into account; can be null
+     * @return a list of bans
+     */
```
KILLEliteMaste, line 1564: Why did you add / edit the Javadoc in the implementation and not in the interface available in the API module?

javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
```diff
@@ -159,7 +159,26 @@ public enum Intent {
...
      */
-    DIRECT_MESSAGE_TYPING(14, false);
+    DIRECT_MESSAGE_TYPING(14, false),
+
+    /**
+     * MESSAGE_CONTENT is a special case as it doesn't represent individual events,
+     * but rather affects the data sent for most events that could contain message content fields
+     * (content, attachments, embeds, and components).
+     */
+    MESSAGE_CONTENT(15, false),
```
KILLEliteMaste, line 169: This is a privileged intent, so it has to be true

KILLEliteMaste (changes requested):
Since the attachments for messages are now removed when editing their attachments, this is a breaking change and the MessageUpdater needs to be updated properly to be able to use the new feature.
The changelog also states `/channels/<channel_id>/threads/active has been removed` which you did not

KILLEliteMaste:
> Screenshot 2022-05-16 at 16 43 12
> 
> It is not there

Then we have this endpoint not implemented. I thought we support everything for threads but then it doesn't seem like that.

RealYusufIsmail:
I will work on MessageUpdater  later today.

RealYusufIsmail:
@KILLEliteMaste what would I change with the Message updater. Like do i need to change anything in MessageBuilderBase or some other class ?

KILLEliteMaste:
> @KILLEliteMaste what would I change with the Message updater. Like do i need to change anything in MessageBuilderBase or some other class ?

I don't know, that's something you have to figure out. Since the MessageBuilderBase contains some functionality that's used by the MessageBuilder/Updater it may be possible that you need to change something even in the MessageBuilder to correctly adjust to changes in the MessageBuilderBase

RealYusufIsmail:
Ok

RealYusufIsmail:
Will start working on this today.

PR #996, opened 2022-05-14
Updated activity flags

### LINKED ISSUE #915: Update activity flags
For more information see: https://github.com/discord/discord-api-docs/pull/3877

RealYusufIsmail:
This update introduces the Activity flags enum which includes
```
| Name                        | Value  |
|-----------------------------|--------|
| INSTANCE                    | 1 << 0 |
| JOIN                        | 1 << 1 |
| SPECTATE                    | 1 << 2 |
| JOIN_REQUEST                | 1 << 3 |
| SYNC                        | 1 << 4 |
| PLAY                        | 1 << 5 |
| PARTY_PRIVACY_FRIENDS       | 1 << 6 |
| PARTY_PRIVACY_VOICE_CHANNEL | 1 << 7 |
| EMBEDDED                    | 1 << 8 |
```

Also closes #915

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/activity/Activity.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/activity/Activity.java
@@ -4,6 +4,7 @@ import org.javacord.api.entity.Nameable;
 import org.javacord.api.entity.emoji.Emoji;
 
 import java.time.Instant;
+import java.util.EnumSet;
 import java.util.Optional;
 
 /**
@@ -81,4 +82,11 @@ public interface Activity extends Nameable {
      */
     Optional<Emoji> getEmoji();
 
+    /**
+     * Gets the activity flags of this activity.
+     * The flags describe what the payload includes.
+     *
+     * @return the activity flags.
+     */
+    EnumSet<ActivityFlag> getFlags();
 }
--- a/javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlag.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlag.java
@@ -1,0 +1,48 @@
+package org.javacord.api.entity.activity;
+
+public enum ActivityFlag {
+
+    INSTANCE(1 << 0),
+    JOIN(1 << 1),
+    SPECTATE(1 << 2),
+    JOIN_REQUEST(1 << 3),
+    SYNC(1 << 4),
+    PLAY(1 << 5),
+    PARTY_PRIVACY_FRIENDS(1 << 6),
+    PARTY_PRIVACY_VOICE_CHANNEL(1 << 7),
+    EMBEDDED(1 << 8),
+    UNKNOWN(-1);
+
+    private final int flag;
+
+    /**
+     * The class constructor.
+     * @param flag The bitmask of the flag.
+     */
+    ActivityFlag(int flag) {
+        this.flag = flag;
+    }
+
+    /**
+     * Gets the activity flag of the given integer value.
+     * @param flags The integer value of the flag.
+     * @return The activity flag of the given integer value.
+     */
+    public static ActivityFlag getActivityFlagById(int flags) {
+        for (ActivityFlag activityFlag : values()) {
+            if (activityFlag.flag == flags) {
+                return activityFlag;
+            }
+        }
+        return UNKNOWN;
+    }
+
+    /**
+     * Gets the integer value of the activity flag.
+     * @return Gets the integer value of the activity flag.
+     */
+    public int asInt() {
+        return flag;
+    }
+}
+
--- a/javacord-core/src/main/java/org/javacord/core/entity/activity/ActivityImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/activity/ActivityImpl.java
@@ -3,6 +3,7 @@ package org.javacord.core.entity.activity;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.javacord.api.entity.activity.Activity;
 import org.javacord.api.entity.activity.ActivityAssets;
+import org.javacord.api.entity.activity.ActivityFlag;
 import org.javacord.api.entity.activity.ActivityParty;
 import org.javacord.api.entity.activity.ActivityType;
 import org.javacord.api.entity.emoji.Emoji;
@@ -10,6 +11,7 @@ import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.entity.emoji.UnicodeEmojiImpl;
 
 import java.time.Instant;
+import java.util.EnumSet;
 import java.util.Objects;
 import java.util.Optional;
 
@@ -29,6 +31,7 @@ public class ActivityImpl implements Activity {
     private final Long startTime;
     private final Long endTime;
     private final Emoji emoji;
+    private final EnumSet<ActivityFlag> flags = EnumSet.noneOf(ActivityFlag.class);
 
     /**
      * Creates a new activity object.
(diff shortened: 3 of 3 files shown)


RealYusufIsmail:
@KILLEliteMaste can you please check if I am missing anything.

KILLEliteMaste (commented):
No idea what you did here, but you have edited the completely wrong Enum. The linked PR for the changes is not meant for the Message Activity Type as you can see here [Discord Docs](https://discord.com/developers/docs/resources/channel#message-object-message-activity-types)

RealYusufIsmail:
Ok sorry for the inconvenience

RealYusufIsmail:
I cant find the enum for this.

javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlags.java
```diff
@@ -0,0 +1,47 @@
+package org.javacord.api.entity.activity;
+
+public enum ActivityFlags {
```
KILLEliteMaste, line 3:
```suggestion
public enum ActivityFlag {
```

javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlags.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param flags The integer value of the flag.
+     * @return The flags' id of the given integer value.
+     */
+    public static ActivityFlags getActivityFlagsById(int flags) {
+        for (ActivityFlags activityFlag : values()) {
+            if (activityFlag.flag == flags) {
+                return activityFlag;
+            }
+        }
+        return null;
```
KILLEliteMaste, line 36: Javacord **NEVER** returns `null`. Instead add another enum type `UNKNOWN(-1)` and return this.

javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlags.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * The class constructor.
+     * @param flag The bitmask of the flag.
+     */
+    ActivityFlags(int flag) {
+        this.flag = flag;
+    }
+
+    /**
+     * Gets the flag of the given integer value.
```
KILLEliteMaste, line 26:
```suggestion
     * Gets the activity flag of the given integer value.
```

javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlags.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param flags The integer value of the flag.
+     * @return The flags' id of the given integer value.
```
KILLEliteMaste, line 28:
```suggestion
     * @return The activity flag of the given integer value.
```

javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlags.java
```diff
@@ -0,0 +1,47 @@
...
+            if (activityFlag.flag == flags) {
+                return activityFlag;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Gets the integer value of the flag.
+     * @return Gets the integer value of the flag.
```
KILLEliteMaste, line 41:
```suggestion
     * Gets the integer value of the activity flag.
     * @return Gets the integer value of the activity flag.
```

javacord-api/src/main/java/org/javacord/api/entity/activity/Activity.java
```diff
@@ -81,4 +81,9 @@ public interface Activity extends Nameable {
      */
     Optional<Emoji> getEmoji();
 
+    /**
+     * activity flags OR d together, describes what the payload includes
+     * @return the activity flags.
+     */
+    ActivityFlags getFlags();
```
KILLEliteMaste, line 88: From what I can see this should return a List of ActivityFlag as it says they are OR d together.

javacord-api/src/main/java/org/javacord/api/entity/activity/Activity.java
```diff
@@ -81,4 +81,9 @@ public interface Activity extends Nameable {
      */
     Optional<Emoji> getEmoji();
 
+    /**
+     * activity flags OR d together, describes what the payload includes
+     * @return the activity flags.
```
KILLEliteMaste, line 86:
```suggestion
     * Gets the activity flags of this activity.
     * The flags describe what the payload includes.
     *
     * @return the activity flags.
```
Also don't just copy paste the description straight from the docs. As you can see above we have a structure for the Javadoc and it should also follow this.

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/activity/Activity.java
RealYusufIsmail: Screenshot 2022-05-16 at 19 13 38
Does not show it as list on the docs.

RealYusufIsmail: Should I change it anyway.

KILLEliteMaste: It doesn't make any sense to show any array / list type in the docs. Like it's written there, they are OR d together that would mean that besides `1000` (binary) it would also be possible to have values like `1100` which represent 2 values (in this case  SPECTATE and JOIN_REQUEST)

RealYusufIsmail: oh ok.

KILLEliteMaste: When it comes to how to do this you can have a look at how it's done here https://github.com/Javacord/Javacord/blob/c840cd6aa1f8a6f8e16fbda166123776d8992f29/javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java#L75-L82

RealYusufIsmail: Should I also make this flags an enum set

KILLEliteMaste: Yes

RealYusufIsmail:
Screenshot 2022-05-16 at 20 11 27
Is there any way to avoid this.

javacord-api/src/main/java/org/javacord/api/entity/activity/Activity.java
RealYusufIsmail: Ok done.

javacord-api/src/main/java/org/javacord/api/entity/activity/ActivityFlag.java
```diff
@@ -0,0 +1,48 @@
...
+    ActivityFlag(int flag) {
+        this.flag = flag;
+    }
+
+    /**
+     * Gets the activity flag of the given integer value.
+     * @param flags The integer value of the flag.
+     * @return The activity flag of the given integer value.
+     */
+    public static ActivityFlag getActivityFlagsById(int flags) {
```
KILLEliteMaste, line 31:
```suggestion
    public static ActivityFlag getActivityFlagById(int flags) {
```

javacord-core/src/main/java/org/javacord/core/entity/activity/ActivityImpl.java
```diff
@@ -141,6 +155,11 @@ public Optional<Emoji> getEmoji() {
         return Optional.ofNullable(emoji);
     }
 
+    @Override
+    public EnumSet<ActivityFlag> getFlags() {
+        return flags;
```
KILLEliteMaste, line 160:
```suggestion
        return EnumSet.copyOf(flags);
```

KILLEliteMaste (changes requested):
Only 2 more small things

RealYusufIsmail:
Ok

KILLEliteMaste (approved):
LGTM!

PR #1001, opened 2022-05-18
Add support for creating, updating, deleting and bulk overwriting slash commands without Server instance.

ShindouMihou:
This adds methods that enable creating, updating, and deleting slash commands without the need for a Server instance. The primary motivation for this is to allow the ability to perform those actions regardless of the shard and cluster, in our case, we only needed the first shard to perform the necessary actions to keep the commands up-to-date.

- `bulkOverwriteServerApplicationCommands(long server, List<? extends ApplicationCommandBuilder<?, ?, ?>> applicationCommandBuilderList)`
- `deleteForServer(long server);`
- `createForServer(DiscordApi api, long server)`
- `updateForServer(DiscordApi api, long server)`

This also modifies the behavior of `ApplicationCommand` to store the server id instead of the Server instance but keeps the same Optional behavior for when the command isn't of a server slash command. (I'd be happy to get some notes if this does any bad side effect).

✅ The PR is tested to work for quite some time, although still unsure of the side effects of modifying the behavior of `ApplicationCommand`.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -295,8 +295,25 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
      *                                      which should be used to perform the bulk overwrite.
      * @return A list containing all application commands.
      */
-    CompletableFuture<List<ApplicationCommand>> bulkOverwriteServerApplicationCommands(
+    default CompletableFuture<List<ApplicationCommand>> bulkOverwriteServerApplicationCommands(
             Server server,
+            List<? extends ApplicationCommandBuilder<?, ?, ?>> applicationCommandBuilderList) {
+        return bulkOverwriteServerApplicationCommands(server.getId(), applicationCommandBuilderList);
+    }
+
+    /**
+     * Bulk overwrites the servers application commands.
+     * This should be preferably used when updating and/or creating multiple
+     * application commands at once instead of {@link ApplicationCommandUpdater#updateForServer(Server)}
+     * and {@link ApplicationCommandBuilder#createForServer(Server)}
+     *
+     * @param server                        The server where the bulk overwrite should be performed on
+     *                                      which should be used to perform the bulk overwrite.
+     * @param applicationCommandBuilderList A list containing the ApplicationCommandBuilders.
+     * @return A list containing all application commands.
+     */
+    CompletableFuture<List<ApplicationCommand>> bulkOverwriteServerApplicationCommands(
+            long server,
             List<? extends ApplicationCommandBuilder<?, ?, ?>> applicationCommandBuilderList);
 
     /**
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
@@ -67,6 +67,13 @@ public interface ApplicationCommand extends DiscordEntity, Specializable<Applica
     boolean getDefaultPermission();
 
     /**
+     * Gets the server id of this command if it is not global.
+     *
+     * @return The server of this command.
+     */
+    Optional<Long> getServerId();
+
+    /**
      * Gets the server of this command if it is not global.
      *
      * @return The server of this command.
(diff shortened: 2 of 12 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -299,6 +299,21 @@ CompletableFuture<List<ApplicationCommand>> bulkOverwriteServerApplicationComman
             Server server,
             List<? extends ApplicationCommandBuilder<?, ?, ?>> applicationCommandBuilderList);
 
+    /**
```
KILLEliteMaste, line 302: Can't select it but you can delegate the method here in the interface to your newly introduced method so we only have 1 implementation

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandBuilder.java
```diff
@@ -71,4 +71,18 @@ public CompletableFuture<R> createForServer(Server server) {
         return delegate.createForServer(server);
```
KILLEliteMaste, line 71: Delegate the call to your new method and then you can also remove the methods from the Delegates for this specific method

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
```diff
@@ -84,4 +91,12 @@ public interface ApplicationCommand extends DiscordEntity, Specializable<Applica
      * @return A future to check if the deletion was successful.
      */
     CompletableFuture<Void> deleteForServer(Server server);
+
+    /**
+     * Deletes this application command globally.
```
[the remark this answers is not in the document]
ShindouMihou: This is actually a comment issue from the original method (I copied the method without looking at the method) 😆 
https://github.com/BeemoBot/Javacord/blob/13a44ef17964c82c7a796ce722cebd822053040f/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java#L80

I'll fix it up on this PR though.

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
KILLEliteMaste: Basically instead of how it's like now you do
```java
default CompletableFuture<List<ApplicationCommand>> bulkOverwriteServerApplicationCommands(
            Server server,
            List<? extends ApplicationCommandBuilder<?, ?, ?>> applicationCommandBuilderList){
return bulkOverwriteServerApplicationCommands(server.getId(), applicationCommandBuilderList)
}
```

ShindouMihou:
I should have applied all the following suggestions with the ApplicationCommand's `server` and `serverId` both added in `ApplicationCommand` as a compromise (https://discord.com/channels/151037561152733184/354648262453624833/976895755518353408) which works well for determining the status of the command itself (whether it's a server application command or a global application command).

It's not squashed ahead since I think I might want more checks to see whether all the suggestions were actually applied properly or all the suggestions were really applied. 💯

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandImpl.java
```diff
@@ -28,6 +28,7 @@ public abstract class ApplicationCommandImpl implements ApplicationCommand {
     private final Map<DiscordLocale, String> descriptionLocalizations = new HashMap<>();
 
     private final Server server;
+    private final long serverId;
```
KILLEliteMaste, line 31: Can be null as you are now doing checks if the `id == 0L` when it is not available. Also simplifies the return type of `getServerId()` with `Optional.ofNullable(serverId)`

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandImpl.java
ShindouMihou: I tried setting this as null before but Java doesn't allow primitive types to be null. I could set it to the Long class instead which is nullable.

KILLEliteMaste: Exactly, doesn't really make sense to have it as a primitive type since you return `Long` in the getteranyway

KILLEliteMaste (changes requested):

ShindouMihou:
Applied the remaining suggestions and squashed into a single commit, should be ready to go if there aren't any more suggestions.

KILLEliteMaste (changes requested):
LGTM!

KILLEliteMaste (approved):
LGTM!

PR #1004, opened 2022-05-25
Rework application command permissions for permissions v2

### LINKED ISSUE #992: Rework Application Command Permissions for v2
https://github.com/discord/discord-api-docs/pull/4830

Mysterypotatoguy:
Reworks application commands for permissions v2

Adds the new `default_member_permissions` and `dm_permission` functionality
Removes the old `default_permission` functionality
Adds the `APPLICATION_COMMAND_PERMISSIONS_UPDATE` event
Adds some audit log handling for the new log type

I tried to use some more friendly names for the builder/updater methods, but I'm open to any renaming suggestions as I'm not sure they are 100% clear.

Closes #992


javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
```diff
@@ -62,17 +63,39 @@ public B addDescriptionLocalization(DiscordLocale locale, String localization) {
...
+     * This can later be overridden by server admins.
      *
-     * @param defaultPermission The default permission to set.
+     * @param requiredPermissions The required permissions to use this command.
      * @return The current instance in order to chain call methods.
      */
-    public B setDefaultPermission(boolean defaultPermission) {
-        delegate.setDefaultPermission(defaultPermission);
-        return (B) this;
+    public T setDefaultRequiredPermissions(PermissionType... requiredPermissions) {
```
KILLEliteMaste, line 73: Same here with the comments in the `ApplicationCommandBuilder`

javacord-api/src/main/java/org/javacord/api/event/server/ApplicationCommandPermissionsUpdateEvent.java
```diff
@@ -0,0 +1,34 @@
...
+     * @return The ID of the command updated.
+     */
+    Optional<Long> getCommandId();
+
+    /**
+     * Gets the updated permissions.
+     *
+     * @return The updated permissions.
+     */
+    Collection<ApplicationCommandPermissions> getUpdatedPermissions();
```
KILLEliteMaste, line 33: Should be a `Set` since this will not contain duplicates

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
```diff
@@ -60,11 +62,29 @@ public interface ApplicationCommand extends DiscordEntity, Specializable<Applica
...
+    Optional<EnumSet<PermissionType>> getDefaultRequiredPermissions();
+
+    /**
+     * Gets whether this command is disabled and only usable by server administrators by default.
+     * Note that this is different to {@link #getDefaultRequiredPermissions} returning
+     * {@link PermissionType#ADMINISTRATOR}.
+     *
+     * @return Whether this command is disabled by default.
+     */
+    boolean isDisabledByDefault();
```
KILLEliteMaste, line 79: See `ApplicationCommandBuilder`

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandPermissions.java
```diff
@@ -27,19 +30,31 @@ public interface ApplicationCommandPermissions {
...
-                .build();
-    }
+    Optional<ServerChannel> getChannel();
+
+    /**
+     * Gets whether these permissions affect all the server's channels.
+     *
+     * @return Whether these permissions affect all the server's channels.
+     */
+    boolean affectsAllChannels();
```
KILLEliteMaste, line 58: You forgot to add a method to check if `@everyone` is affected. To simplify this you can also directly add a `getServer` method so you just need to implement them in the interface.

https://discord.com/developers/docs/interactions/application-commands#application-command-permissions-object-application-command-permissions-constants

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandPermissions.java
Mysterypotatoguy: The `@everyone` role is the `guild_id` so it will fetch correctly

KILLEliteMaste: I am not sure what you mean by that. What I asked for was a `affectsEveryone` method which return true if the id is the server id or is this not "needed"?

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandBuilder.java
```diff
@@ -62,14 +63,38 @@ public T addDescriptionLocalization(DiscordLocale locale, String localization) {
     }
 
     /**
-     * Sets the default permission for the message context menu
-     * whether the context menu is enabled by default when the app is added to a server.
+     * Sets the default required permissions for this command.
+     * Passing no arguments will cause the command to be available to all users.
```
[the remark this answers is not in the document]
Mysterypotatoguy: Hmm, I do like this solution because it removes the overwriting behaviour of trying to use both `setDefaultRequiredPermissions` and `setDisabledByDefault`, but the `null` behaviour seems very odd to me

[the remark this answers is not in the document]
KILLEliteMaste: I just think this is confusing. I didn't get what this was doing until I looked at the implementation where I saw both methods change the `defaultMemberPermissions` variable.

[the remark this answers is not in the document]
KILLEliteMaste: After thinking about it I can agree that passing `null` is pretty odd in Javacord. It doesn't fit in the style of everything else. What do you think about renaming the method to `setDefaultRequiredPermissionsDisabled` ?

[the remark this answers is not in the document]
Mysterypotatoguy: If you mean for the `null` argument proposed above I don't think that's clearer than `setDisabledByDefault`. It sounds more like there are _no_ required permissions

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandPermissions.java
Mysterypotatoguy: I see, that would be fine but it's just a shortcut for `getRole() == @everyone`

KILLEliteMaste: Yes but it's a special case of a role, which may be used more often or differently than if it just affects a general role. And nothings wrong about convenience methods :)

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandBuilder.java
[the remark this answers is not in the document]
KILLEliteMaste: That's not what I said. I agree with what you said and therefore proposed to change the name of the method from `setDisabledByDefault` to `setDefaultRequiredPermissionsDisabled` as imo it is not clear that they change the same thing just from looking at the method name.

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandPermissionsImpl.java
```diff
@@ -52,17 +62,30 @@ public boolean getPermission() {
...
+
+    @Override
+    public boolean affectsAllChannels() {
+        return id == server.getId() - 1;
+    }
+
+    @Override
+    public boolean affectsEveryoneRole() {
+        return id == server.getEveryoneRole().getId();
+    }
```
KILLEliteMaste, line 89: Please add a `getServer` method and then you are able to use a default implementation in the interface

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandBuilder.java
```diff
@@ -62,14 +63,48 @@ public T addDescriptionLocalization(DiscordLocale locale, String localization) {
...
+     * Passing no arguments will cause the command to be available to admins only.
+     * This can later be overridden by server admins.
      *
-     * @param defaultPermission The default permission.
+     * @param requiredPermissions The required permissions to use this command.
      * @return The current instance in order to chain call methods.
      */
-    public T setDefaultPermission(Boolean defaultPermission) {
-        delegate.setDefaultPermission(defaultPermission);
+    public T setDefaultEnabledForPermissions(PermissionType... requiredPermissions) {
```
KILLEliteMaste, line 73: Please overload with an `EnumSet` param

javacord-api/src/main/java/org/javacord/api/interaction/internal/ApplicationCommandUpdaterDelegate.java
```diff
@@ -39,11 +40,32 @@
...
     /**
-     * Sets the new application command default permission.
+     * Sets the default required permissions for this command.
+     * This can later be overridden by server admins.
      *
-     * @param defaultPermission The default permission to set.
+     * @param requiredPermissions The required permissions to use this command.
      */
-    void setDefaultPermission(boolean defaultPermission);
+    void setDefaultEnabledForPermissions(PermissionType... requiredPermissions);
```
KILLEliteMaste, line 48: Same param overloading here

KILLEliteMaste (changes requested):
Please also add convenience methods that take this param in, in the `SlashCommand` and the 2 ContextMenu command interfaces. If you are so kind can you also add the missing convenience method for context menus with `name` and `description`? :)

KILLEliteMaste (approved):
Did the final last things that were left, otherwise LGTM!

PR #1009, opened 2022-06-04
Add basic support for forum channels

### LINKED ISSUE #1005: Forum channels in a server breaks `ServerImpl#getChannels()`
image

KILLEliteMaste:
This implementation for forum channels does not contain any functionality for them since they are not released yet. So it is up to a future PR to add forum specific functionality.
This should fix #1005 but I couldn't test this as I do not have access to forum channels.

This also includes a breaking change where `ServerThreadChannel#getParent` now returns a `ServerChannel` instead of a `ServerTextChannel`


javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -96,13 +95,10 @@ public void handle(JsonNode packet) {
...
                 break;
-            case GROUP_CHANNEL:
-                logger.info("Received CHANNEL_UPDATE packet for a group channel. This should be impossible.");
-                break;
-            case CHANNEL_CATEGORY:
+            case SERVER_FORUM_CHANNEL:
                 handleServerChannel(packet);
                 handleRegularServerChannel(packet);
-                handleChannelCategory(packet);
+                //handleServerForumChannel(packet);
```
wasdennnoch, line 101: Commented?

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerForumChannelBuilderDelegateImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public void setCategory(ChannelCategory category) {
+        this.category = category;
+    }
+
+    @Override
+    public CompletableFuture<ServerForumChannel> create() {
+        ObjectNode body = JsonNodeFactory.instance.objectNode();
+        body.put("type", 0);
```
wasdennnoch, line 42: Is this intended to create text channels? Possibly leftover from testing.

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerThreadChannelImpl.java
```diff
@@ -117,8 +117,8 @@ public ServerThreadChannelImpl(final DiscordApiImpl api, final ServerImpl server
     }
 
     @Override
-    public ServerTextChannel getParent() {
-        return getServer().getTextChannelById(parentId)
+    public RegularServerChannel getParent() {
+        return getServer().getRegularChannelById(parentId)
```
wasdennnoch, line 121: Just to make sure: I assume this is done in preparation for storing generic channel types for unknown channels as suggested in https://github.com/Javacord/Javacord/issues/1005#issuecomment-1137880106? Because if that's not the case, this change alone won't solve the issue of Javacord throwing when handling treads in new unsupported channel types.

wasdennnoch (commented):
Don't know if this is ready for review yet, but before I forget I wanted to point out some things I noticed.

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
KILLEliteMaste: As this implementation doesn't contain any forum specific functionality(the method currently doesn't to anything), I've commented this line so in a future PR it just has to be uncommented.

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerForumChannelBuilderDelegateImpl.java
KILLEliteMaste: Thank you for pointing this out :)

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerThreadChannelImpl.java
KILLEliteMaste: >  I assume this is done in preparation for storing generic channel types for unknown channels

It may be a side effect that this paves the way to a `DummyServerChannel` but this was done because RegularServerChannel is just an interface ServerTextChannel and ServerForumChannel extend which fixes the issue for now. 
I think adding Dummy implementations require a bigger refactoring and I am not sure how this would affect Javacord in its current state, that would have to be evaluated a bit more. 
Furthermore imo it will be unlikely that there will be a lot more channels that support threads (or they have to get very creative) besides the new text in voice channels that might be able to have threads but for that this change will also fix the current issue.

wasdennnoch: > Furthermore imo it will be unlikely that there will be a lot more channels that support threads

I don't agree with this sentiment. I highly doubt Discord will just stop adding new text channel types. In the current form, Javacord will throw the `AssertionError` every time Discord adds a new channel type with thread support, meaning a lot of channel-related Javacord functions will break every time, meaning you have to update the library every time to add support for the new channel.
In my view, ideally the library should at least try to be forward-compatible, as in, not break when new channels are added, but rather either drop related events or do sanity checks - anything is better than some functions of the lib randomly breaking (in a way that could've been avoided) as discord tests new channel types. Hence my proposal (including example code) in #1005 to make the return type of `getParent()` an optional, which could be done here too as the PR is a breaking change either way. Or, alternatively, don't even process/store threads originating from unsupported parent channels.

I do realize this library is meant to be easy to use, and adding additional API complexity like this might go against that philosophy. However, as a library user/bot developer of any size I also wouldn't expect the library to throw internal errors all of the sudden and would rather have them handled internally without compromising basic functionality such as grabbing a server's channel list. Right now we have to deal with users being unable to use some essential commands just because their server has a forum channel.
Yes, Discord has in the past pushed truly breaking changes to their API without bumping the version, causing unexpected errors which devs could not have forseen. But new channel types are a thing that absolutely do happen every now and then, and can and should be handled gracefully.

_Yes I just wrote an essay about channel types. I'm sorry lol. But I do think this should be handled better._

KILLEliteMaste: I think the only thing that's breaking Javacord is the `getParent` method of a `ServerThreadChannel`. To address this issue also in case of new channels we have the following options which you also mentioned:
- Return an optional
- Drop all events if the parent cannot be found so the thread will never be added as a channel
- Wrap the `getParent` method (which is used only in `getChannels` internally) in a try block to catch the exception and in that case either don't add the thread to the list or just insert it. In both cases the list is sorted but not complete which would be a good compromise imo that everything still works. The `getParent` method will still not work if you call it in your code and throw an exception but it shouldn't affect anything else in Javacord

Going to discuss this with other contributors to see what they think about that

bassner (approved):
Code looks good to me as it's basically nothing new, just taken from the other channel types. Can't test as I don't have access to forum channels anywhere

bassner (changes requested):
Just one minor suggestion

javacord-core/src/main/java/org/javacord/core/entity/channel/UnknownRegularServerChannelImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getId());
+    }
+
+    @Override
+    public String toString() {
+        return String.format("ServerForumChannel (id: %s, name: %s)", getIdAsString(), getName());
```
wasdennnoch, line 46:
```suggestion
        return String.format("UnknownRegularServerChannel (id: %s, name: %s)", getIdAsString(), getName());
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2465,6 +2476,56 @@ default List<ServerChannel> getChannelsByNameIgnoreCase(String name) {
...
+     * @param name The name of the channels.
+     * @return A sorted list (by position) with all channels with the given name.
+     */
+    default List<ServerChannel> getRegularChannelsByNameIgnoreCase(String name) {
+        return Collections.unmodifiableList(
+                getChannels().stream()
+                        .filter(channel -> channel.getName().equalsIgnoreCase(name))
+                        .collect(Collectors.toList()));
+    }
+
```
bassner, line 2528: Why do these two not return ``RegularServerChannel``?

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelDeleteHandler.java
```diff
@@ -65,12 +63,57 @@ public void handle(JsonNode packet) {
...
+                break;
+            case CHANNEL_CATEGORY:
+                handleCategory(packet);
+                break;
+            case PRIVATE_CHANNEL:
+                handlePrivateChannel(packet);
+                break;
+            default: {
+                handleUnknownRegularServerChannel(packet);
+                handleUnknownServerChannel(packet);
```
wasdennnoch, line 77: Since the create/update handler have custom error handling as well, for consistency:
```suggestion
                try {
                    handleUnknownRegularServerChannel(packet);
                    handleUnknownServerChannel(packet);
                } catch (Exception exception) {
                    logger.warn("An error occurred when trying to delete a fallback channel implementation", exception);
                }
```

wasdennnoch (commented):
From what I see the unknown channel types assume a `guild_id` is always present, which wouldn't be the case for new DM channel, or if discord decides to allow bots back in DM groups. To be fair, the chance of _that_ happening is probably very low, and the packet processing errors would be caught, so it might not matter much.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
KILLEliteMaste: Good catch :)

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelDeleteHandler.java
KILLEliteMaste: It's not only consistency, but like you mentioned in case Discord adds some sort of channel that does not belong to a server  it will throw an exception when trying to get the server id.

PR #1013, opened 2022-06-15
updated dependencies so they are more in sync

feddevanderlist:
I updated the log4j update on the normal build.
Also updated test plugins to the highest version they would go without failing.

### DIFF
--- a/gradle/tests.gradle
+++ b/gradle/tests.gradle
@@ -3,17 +3,15 @@ subprojects {
     apply plugin: 'groovy'
 
     dependencies {
-        testImplementation 'org.spockframework:spock-core:1.2-groovy-2.5'
-        // make the tests compilable on Java 11, can be removed when updating to Spock 1.3
-        testImplementation 'org.codehaus.groovy:groovy:2.5.4'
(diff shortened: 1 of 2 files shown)


gradle/tests.gradle
```diff
@@ -3,29 +3,27 @@ subprojects {
     apply plugin: 'groovy'
 
     dependencies {
-        testImplementation 'org.spockframework:spock-core:1.2-groovy-2.5'
-        // make the tests compilable on Java 11, can be removed when updating to Spock 1.3
-        testImplementation 'org.codehaus.groovy:groovy:2.5.4'
+        testImplementation 'org.spockframework:spock-core:1.3-groovy-2.5'
```
KILLEliteMaste, line 6: Is there a reason why you didn't upgrade this dependency to the latest version?

feddevanderlist: The v2 version doesn't build and I didn't have time to update to find out why.

KILLEliteMaste (approved):
Thank you for updating the dependencies, LGTM!

PR #1014, opened 2022-06-15
Add text in voice channels

KILLEliteMaste:
## Checklist

## Changelog
- Added support for text in Voice
- Added `TextableRegularServerChannel` which contains shared methods for text related functionality of RegularServerChannels
- Fixed a bug where `canWrite` and more methods return always true for Threads

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/TextableRegularServerChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/TextableRegularServerChannel.java
@@ -1,0 +1,123 @@
+package org.javacord.api.entity.channel;
+
+import org.javacord.api.entity.Mentionable;
+import org.javacord.api.entity.permission.PermissionType;
+import org.javacord.api.entity.user.User;
+import org.javacord.api.entity.webhook.WebhookBuilder;
+
+public interface TextableRegularServerChannel extends TextChannel, Mentionable, Categorizable {
+
+    @Override
+    default boolean canWrite(User user) {
+        return asRegularServerChannel().map(regularServerChannel ->
+                regularServerChannel.hasPermission(user, PermissionType.ADMINISTRATOR)
+                        || regularServerChannel.hasPermissions(user, PermissionType.VIEW_CHANNEL,
+                        PermissionType.SEND_MESSAGES)).orElse(false);
+    }
+
+    @Override
+    default boolean canUseExternalEmojis(User user) {
+        if (!canWrite(user)) {
+            return false;
+        }
+        return asRegularServerChannel().map(regularServerChannel ->
+                regularServerChannel.hasAnyPermission(user,
+                        PermissionType.ADMINISTRATOR,
+                        PermissionType.USE_EXTERNAL_EMOJIS)).orElse(false);
+    }
+
+    @Override
+    default boolean canEmbedLinks(User user) {
+        if (!canWrite(user)) {
+            return false;
+        }
+        return asRegularServerChannel().map(regularServerChannel ->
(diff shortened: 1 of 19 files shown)


Mysterypotatoguy:
Perhaps `MessageableRegularServerChannel` to fit with the current `Messageable` interface?

KILLEliteMaste:
> Perhaps `MessageableRegularServerChannel` to fit with the current `Messageable` interface?

I also thought about that but I don't think that fits well as `Messageable` only contains send methods while `TextableRegularServerChannel` is a lot more than that. I would expect with that name some specialized send methods or something similar

Saladoc:
Maybe we can include a `ServerAudioChannel` as a common ancestor for `ServerStageVoiceChannel` and `ServerVoiceChannel`. We can revert that if text chat is ever enabled for stage channels, but we shouldn't offer functionality that is not there.

javacord-api/src/main/java/org/javacord/api/entity/channel/TextableRegularServerChannel.java
```diff
@@ -0,0 +1,123 @@
+package org.javacord.api.entity.channel;
+
+import org.javacord.api.entity.Mentionable;
+import org.javacord.api.entity.permission.PermissionType;
+import org.javacord.api.entity.user.User;
+import org.javacord.api.entity.webhook.WebhookBuilder;
+
+public interface TextableRegularServerChannel extends TextChannel, Mentionable, Categorizable {
```
Saladoc, line 8: From the name I would expect `TextableRegularServerChannel` to extend RegularServerChannel` - is there something preventing that?

Saladoc (commented):
Need to verify [Mysterypotatoguys note](https://github.com/Javacord/Javacord/pull/1014/files#r906203093) and possibly implement that.

javacord-api/src/main/java/org/javacord/api/entity/channel/TextableRegularServerChannel.java
KILLEliteMaste: Yes, this clashes with some methods from the other interfaces but the name might be just chosen badly. The channel structure also gets quiet worse considering that the voice channel methods need to be put into a separate interface so the `ServerStageVoiceChannel` can no longer extend from the `ServerVoiceChannel`. In addition I think this PR should be redone once we have decided what the channel hierarchy should look like.

KILLEliteMaste: This is our current channel hierarchy (without the PR applied)
image

Just taking the `VoiceChannel` as an example, it is no longer possible to have voice in private channels. So we have to decide whether we want to keep it clean for possible future non server voice channels or also move methods into it that are usable only on a server. Depending on what we decide, if we want to keep it clean, we would need a new interface `ServerVoiceChannel` but this would clash with our current existing channel.

RealYusufIsmail: will need changing considering #1162

KILLEliteMaste:
> I'm still not quite happy with the channel hierarchy, but since I can't come up with a better solution right now I think it will be better to accept this for now and revisit the entire structure at a later point.

Yeah, since Discord adds more and more channels which share functionality and are also unique to its kind, a re-visit of the channel structure is necessary at some point.

Any idea what the breaking change is? Is been quiet some time ago since I created this PR and can't see atm what the breaking change is.

Saladoc:
I think it was the changed inheritances which might break some binary compatibility. I doubt it's a serious change, but it might require recompilation, so better safe than sorry in marking it as breaking.

PR #1015, opened 2022-06-16
Adds support for Attachment options in Slash Commands.

### LINKED ISSUE #958: Implement Attachment Option Type for Slash Interactions
See https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-type

ShindouMihou:
This patch adds support for the attachment option type in slash commands and fixes #958. It's a simple change that creates a new `Attachment` class containing shared properties with `MessageAttachment` and `Attachment`. 

✅ Tested with multiple attachments and has been verified to work.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionImpl.java
@@ -1,12 +1,14 @@
 package org.javacord.core.interaction;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import org.javacord.api.entity.Attachment;
 import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.user.User;
 import org.javacord.api.interaction.InteractionType;
 import org.javacord.api.interaction.SlashCommandInteraction;
 import org.javacord.api.interaction.SlashCommandInteractionOption;
 import org.javacord.core.DiscordApiImpl;
+import org.javacord.core.entity.AttachmentImpl;
 import org.javacord.core.entity.server.ServerImpl;
 import org.javacord.core.entity.user.MemberImpl;
 import org.javacord.core.entity.user.UserImpl;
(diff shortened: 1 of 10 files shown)


KILLEliteMaste (changes requested):
The `SlashCommandOption` interface should have the convenience methods for attachments too

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionImpl.java
```diff
@@ -54,11 +57,21 @@ public SlashCommandInteractionImpl(DiscordApiImpl api, TextChannel channel, Json
...
                 });
             }
+
+            if (resolved.has("attachments")) {
+                resolved.get("attachments").fields().forEachRemaining(attachmentNode -> {
+                    JsonNode attachmentData = attachmentNode.getValue();
+                    Long id = Long.parseLong(attachmentNode.getKey());
+                    if (!resolvedAttachments.containsKey(id)) {
+                        resolvedAttachments.put(id, new AttachmentImpl(api, attachmentData));
+                    }
```
montlikadani, line 67:
```suggestion
                    resolvedAttachments.computeIfAbsent(Long.parseLong(attachmentNode.getKey()), id -> new AttachmentImpl(api, attachmentData));
```
With `computeIfAbsent` I think looks better

ShindouMihou: I do agree. That section of the code is simply a copy of what was written for `resolvedUsers` as well.

ShindouMihou:
Applied all suggestions from KILLEliteMaste. I didn't apply the last one since I think it's better for now to keep the same style as the other resolved data such as `resolvedMembers` and `resolvedUsers`.

javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionImpl.java
KILLEliteMaste: In that case it doesn't make sense to copy it from the users. The users need it because of how Javacord works (user == member object). If we did not do this check we would simply overwrite the user which contains the member.
It is not wrong, but also not needed. I would suggest to apply this suggestion for both (user / attachment) or just do a simple `put` for users as it looks cleaner.

ShindouMihou: Actually yeah, that made sense. I think my brain went offline while looking at this suggestion. A simple `put` should just do the trick in this scenario.

ShindouMihou: Applied the following changes, thanks for the input! 🧇

KILLEliteMaste (approved):
LGTM!

PR #1017, opened 2022-06-16
Fix a bug where sending Message without attachments results in 2 different message (Return value & MSGCreateEvent)

KILLEliteMaste:


bassner (changes requested):
Thanks for the fix!
We should fix it here as well
https://github.com/Javacord/Javacord/blob/03af2019719d08d1cd92f0a1a4ddad85f012d385/javacord-core/src/main/java/org/javacord/core/interaction/MessageComponentInteractionImpl.java#L36

bassner:
I guess as a last thing you need to fix the warnings xD

bassner (approved):

PR #1021, opened 2022-06-17
update netty and sprock-core

feddevanderlist:


Vampire (changes requested):
You cannot simply update Spock like that.
1. Don't update toa milestone version
2. Like you did it, no tests are run as Spock v2 is based on JUnit platform and not anymore on JUnit 4, you should at least do minimal testing when opening a PR. ;-)

feddevanderlist:
I'll go and do some research. I'm used to mvn and strongly typed unit tests.
So, it'll take some time. But I want to learn

feddevanderlist:
Should be good now. This is as far as I can safely go before we need to go to junit platform. Or when netty might stop supporting socks. But I'll look into that in a new branch.

Vampire:
Please combine your commits into one. :-)

feddevanderlist:
> Please combine your commits into one. :-)

Is there a way to do that. I normally just use the squash commits button in git.

Vampire:
Sure, just rework the commits how they should be and force-push.
In PRs force-pushes are fine.
Here an interesting read about Git post-production: https://sethrobertson.github.io/GitPostProduction/

Vampire (approved):
LGTM

PR #1045, opened 2022-07-09
Fix event queue and sticker memory leaks

wasdennnoch:
This fixes two memory leak sources I have noticed:

- When leaving a server, the corresponding `Server` instance isn't removed as an event target key from the `queuedListenerTasks` in the event dispatcher, causing it to stay there forever even though no events will ever be dispatched for it anymore. Fx this by processing all remaining events in the queue for that server and then removing the queue.

- The big one: `DiscordApiImpl.stickers` was static, leaking DiscordApi instances. This only happens when you replace DiscordApi instances at runtime, but when you do it's a severe issue: Since `stickers` is static, when an API instance is destroyed, all its stickers will remain saved there - and each sticker has a reference to that old API instance, keeping it as well as all its caches from getting GC'd, leaking ~150MB of memory per stray instance of a larger shard.
I did wonder why it was static in the first place, and it seems it initially wasn't but then was made static based on a review comment in #904 mentioning sharding support. However, `customEmojis` also aren't static, so I'm not sure why stickers should be - if the API instance owning those stickers goes away, so should the stickers since there is no guarantee they're up to date anymore. If you want a list of all stickers, you can just loop through your `DiscordApi` instances and collect them.

Those changes have been running fine in Beemo for the past ~3 months (though I didn't specifically check if the queue fix is working as intended, but it hasn't caused any issues so far, so, I guess). I feel like the dispatcher fix is a little hacky so lmk if I can improve it somehow.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
@@ -208,9 +208,13 @@ public abstract class EventDispatcherBase {
                     } catch (InterruptedException ignored) { }
                 }
             }
-            Queue<Runnable> queue = queuedListenerTasks.computeIfAbsent(
-                    queueSelector, o -> new ConcurrentLinkedQueue<>());
-            listeners.forEach(listener -> queue.add(() -> consumer.accept(listener)));
+            if (!listeners.isEmpty()) {
+                synchronized (queuedListenerTasks) {
+                    Queue<Runnable> queue = queuedListenerTasks.computeIfAbsent(
+                            queueSelector, o -> new ConcurrentLinkedQueue<>());
+                    listeners.forEach(listener -> queue.add(() -> consumer.accept(listener)));
+                }
+            }
             checkRunningListenersAndStartIfPossible(queueSelector);
         });
     }
@@ -233,6 +237,10 @@ public abstract class EventDispatcherBase {
             // - running for object-dependent tasks, but queue is empty or not present
             Queue<Runnable> queue = queueSelector == null ? null : queuedListenerTasks.get(queueSelector);
             if (queue == null || queue.isEmpty()) {
+                if (queueSelector != null) {
+                    // Remove the entry for the queue selector, so that it eventually can be garbage collected
+                    queuedListenerTasks.remove(queueSelector);
+                }
                 // if no object-independent tasks to be processed everything is fine, return
                 if (queuedListenerTasks.get(null).isEmpty()) {
                     return;
@@ -285,8 +293,15 @@ public abstract class EventDispatcherBase {
                     activeListeners.remove(activeListener);
                     alreadyCanceledListeners.remove(activeListener);
                     runningListeners.remove(finalQueueSelector);
-                    // Inform the dispatchEvent method that it maybe can queue new listeners now
                     synchronized (queuedListenerTasks) {
+                        // Remove the entry for the queue selector, so that it eventually can be garbage collected
+                        if (finalQueueSelector != null) {
+                            Queue<Runnable> remainingQueue = queuedListenerTasks.get(finalQueueSelector);
+                            if (remainingQueue != null && remainingQueue.isEmpty()) {
+                                queuedListenerTasks.remove(finalQueueSelector);
+                            }
+                        }
+                        // Inform the dispatchEvent method that it maybe can queue new listeners now
                         queuedListenerTasks.notifyAll();
                     }
                     checkRunningListenersAndStartIfPossible(finalQueueSelector);
--- a/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java
@@ -371,7 +371,7 @@ public class DiscordApiImpl implements DiscordApi, DispatchQueueSelector {
     /**
      * A map with all stickers.
      */
-    private static final ConcurrentHashMap<Long, Sticker> stickers = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<Long, Sticker> stickers = new ConcurrentHashMap<>();
 
     /**
      * A map with all cached messages.


javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
```diff
@@ -171,6 +171,19 @@ public void setExecutionTimeCheckingEnabled(boolean enable) {
...
 
+    /**
+     * Executes the remaining events in the objet queue for {@code queueSelector} and removes
+     * the object and its task queue from the event dispatcher.
+     *
+     * @param queueSelector The object used to determine which queue to flush and remove.
+     */
+    public void flushAndRemoveTaskQueue(DispatchQueueSelector queueSelector) {
+        while (queuedListenerTasks.containsKey(queueSelector) && !queuedListenerTasks.get(queueSelector).isEmpty()) {
+            checkRunningListenersAndStartIfPossible(queueSelector);
```
KILLEliteMaste, line 182: Does it make sense to run all listeners if the server has left? I mean when there are a bunch of events dispatched and queued up, then if the server leaves it wouldn't make sense to execute them as you can not react anyway as you do not have access anymore to this server

javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
```diff
@@ -171,6 +171,19 @@ public void setExecutionTimeCheckingEnabled(boolean enable) {
         executionTimeCheckingEnabled = enable;
     }
 
+    /**
+     * Executes the remaining events in the objet queue for {@code queueSelector} and removes
```
KILLEliteMaste, line 175: `objet` -> `object`

KILLEliteMaste (changes requested):

KILLEliteMaste:
Iirc my thought was changing it to be static would be better when requesting nitro packs to have them accessible from every API instance to avoid unnecessary requests. And since you can use stickers from a server in other servers too I think, I thought that having it as static was the best choice. But it might be better to change it to non static as I think we have nowhere something like this (across shards). If we one day decide to have some things available across chards, there are probably quiet a few things that need to be rethought.

javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
wasdennnoch: Unsure about this one. Such few events would likely not be of much interest to the user, but at the same time I'd like to avoid dropping events if possible.
Though I haven't yet fully understood how the dispatch itself works (i.e. `checkRunningListenersAndStartIfPossible`) and if there is actually a guarantee that the ServerLeave event will be dispatched if we immediately remove the event queue, or if the event will just be stored in the queue that we then remove, causing it to never be emitted. From what I understand the second case occurs, i.e. during a normal dispatch the event is pushed on the `queuedListenerTasks` and then `checkRunningListenersAndStartIfPossible` starts emitting queued events fifo-style, so the leave event may not be emitted if we immediately nuke the queue.

Vampire: The listeners should run for all events, yes.
Maybe the bot does some bookkeeping of things seen and he should see them or whatever,
and of course also the leave event.

But I don't think the solution here is good.
It takes responsibility that should be handled inside the event dispatcher base out to the guild delete handler where it imho is misplaced. It shouldn't be aware that there is some queue that needs to be flushed and cleaned.
Also it has several race conditions I'm too lazy to type out right now, I'm on vacation. :-)

I didn't test it and just quickly threw it together, so I might have missed something, but what do you think about this instead?
```patch
diff --git a/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java b/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
index e91cd1d0..30812724 100644
--- a/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
@@ -208,10 +208,12 @@ public abstract class EventDispatcherBase {
                     } catch (InterruptedException ignored) { }
                 }
             }
-            Queue<Runnable> queue = queuedListenerTasks.computeIfAbsent(
-                    queueSelector, o -> new ConcurrentLinkedQueue<>());
-            listeners.forEach(listener -> queue.add(() -> consumer.accept(listener)));
-            checkRunningListenersAndStartIfPossible(queueSelector);
+            synchronized (queuedListenerTasks) {
+                Queue<Runnable> queue = queuedListenerTasks.computeIfAbsent(
+                        queueSelector, o -> new ConcurrentLinkedQueue<>());
+                listeners.forEach(listener -> queue.add(() -> consumer.accept(listener)));
+                checkRunningListenersAndStartIfPossible(queueSelector);
+            }
         });
     }

@@ -285,8 +287,15 @@ public abstract class EventDispatcherBase {
                     activeListeners.remove(activeListener);
                     alreadyCanceledListeners.remove(activeListener);
                     runningListeners.remove(finalQueueSelector);
-                    // Inform the dispatchEvent method that it maybe can queue new listeners now
                     synchronized (queuedListenerTasks) {
+                        // Remove the entry for the queue selector, so that it eventually can be garbage collected
+                        if (finalQueueSelector != null) {
+                            Queue<Runnable> remainingQueue = queuedListenerTasks.get(finalQueueSelector);
+                            if ((remainingQueue != null) && remainingQueue.isEmpty()) {
+                                queuedListenerTasks.remove(finalQueueSelector);
+                            }
+                        }
+                        // Inform the dispatchEvent method that it maybe can queue new listeners now
                         queuedListenerTasks.notifyAll();
                     }
                     checkRunningListenersAndStartIfPossible(finalQueueSelector);
```

wasdennnoch: That actually sounds like a much better solution and also supports other kinds of queue selectors, should more be added in the future. And I believe this one should be guaranteed to always run after all remaining items have been processed. I've gone ahead and applied that patch (minus some redundant braces).

javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java
```diff
@@ -285,8 +287,15 @@ private void checkRunningListenersAndStartIfPossible(DispatchQueueSelector queue
...
                     runningListeners.remove(finalQueueSelector);
-                    // Inform the dispatchEvent method that it maybe can queue new listeners now
                     synchronized (queuedListenerTasks) {
+                        // Remove the entry for the queue selector, so that it eventually can be garbage collected
+                        if (finalQueueSelector != null) {
+                            Queue<Runnable> remainingQueue = queuedListenerTasks.get(finalQueueSelector);
+                            if (remainingQueue != null && remainingQueue.isEmpty()) {
+                                queuedListenerTasks.remove(finalQueueSelector);
+                            }
+                        }
```
KILLEliteMaste, line 297: When testing, the `queuedListenerTasks` contains always all servers because of the `ServerBecomesAvailableListener` which gets executed once the servers are available. The issue is not the listener directly but you have to remove the `queueSelector` in line 240 too, if there are no listeners(?) to execute

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):
LGTM! Please also rebase on the latest master version so I can merge it :)

PR #1047, opened 2022-07-10
Adds support for `min_length` and `max_length` for `String` in Slash Commands.

ShindouMihou:
This PR adds the newly implemented `min_length` and `max_length` for String options in Slash Commands. The documentations states `integer` for the parameter (same as for `INTEGER`) but seeing that `INTEGER` uses the `long` type, I decided to use the `long` type as well since I remembered that Discord's integer is longer than Java's integer.

☑️ This PR has been tested and works as intended.
image

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
@@ -117,6 +117,20 @@ public interface SlashCommandOption {
     Optional<Double> getDecimalMaxValue();
 
     /**
+     * If the option is an {@link SlashCommandOptionType#STRING} type, the minimum allowed length.
+     *
+     * @return The minimum allowed length.
+     */
+    Optional<Long> getMinLength();
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#STRING} type, the maximum allowed length.
+     *
+     * @return The maximum allowed length.
+     */
+    Optional<Long> getMaxLength();
+
+    /**
      * Create a new slash command option to be used with a slash command builder.
      * This is a convenience method.
      *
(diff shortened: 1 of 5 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
```diff
@@ -116,6 +116,20 @@ public interface SlashCommandOption {
...
+     * @return The minimum allowed length.
+     */
+    Optional<Long> getStringMinLength();
+
+    /**
+     * If the option is an {@link SlashCommandOptionType#STRING} type, the maximum allowed length.
+     *
+     * @return The maximum allowed length.
+     */
+    Optional<Long> getStringMaxLength();
```
KILLEliteMaste, line 131: I think they should be called more generic. We Use `Long` for the LONG type and `Double` for DECIMAL as the docs tell us. Because this is not bound to something, I think this should be a more generic name like `getMinLength` as this option might be used for other command option types in the future or what do you think about this?

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOptionBuilder.java
```diff
@@ -229,6 +229,29 @@ public SlashCommandOptionBuilder setDecimalMaxValue(double decimalMaxValue) {
...
+     * Sets the {@link SlashCommandOptionType#STRING} min length for the slash command option.
+     *
+     * @param stringMinLength The string min length.
+     * @return The current instance in order to chain call methods
+     */
+    public SlashCommandOptionBuilder setStringMinLength(long stringMinLength) {
+        delegate.setStringMinLength(stringMinLength);
+        return this;
+    }
+
```
KILLEliteMaste, line 242: Remove blank line

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOption.java
ShindouMihou: I decided to follow through with the generic name instead. 🌮

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOptionBuilder.java
```diff
@@ -229,6 +229,29 @@ public SlashCommandOptionBuilder setDecimalMaxValue(double decimalMaxValue) {
...
+
+    /**
+     * Sets the {@link SlashCommandOptionType#STRING} max length for the slash command option.
+     *
+     * @param stringMaxLength The string min length.
+     * @return The current instance in order to chain call methods
+     */
+    public SlashCommandOptionBuilder setStringMaxLength(long stringMaxLength) {
```
KILLEliteMaste, line 250: You forgot to rename these

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandOptionBuilder.java
ShindouMihou: Yeah, I didn't notice. Thanks for the heads up! I double-checked just now and it should all be renamed properly. 🌮

KILLEliteMaste (approved):
LGTM!

PR #1050, opened 2022-07-14
Add SlashCommandInteraction#getFullCommandName

KILLEliteMaste:


javacord-core/src/main/java/org/javacord/core/interaction/SlashCommandInteractionImpl.java
```diff
@@ -102,17 +102,30 @@ public String getCommandName() {
...
             return options;
         }
     }
 
+    @Override
+    public String getFullCommandName() {
+        return getCommandName() + getFullCommandName(getOptions());
+    }
+
+    private String getFullCommandName(List<SlashCommandInteractionOption> options) {
```
Mysterypotatoguy, line 123: Even as an internal method name this seems misleading

KILLEliteMaste: So you think it would be better to rename it to `getFullCommandNameRecursive` like it was previously with `getArguments`?

Mysterypotatoguy: Not at the API level, but yes
Alternatively it could be `getSubcommandNamesRecursive`

Mysterypotatoguy (approved):

PR #1051, opened 2022-07-15
MessageUpdateHandler should now fire an event every time it receives an event

KILLEliteMaste:
### Change
MessageUpdateHandler now dispatches an event every time it receives an update

### Breaking change
Removes get(new/old)Content and embeds from the `MessageEditEvent`. Instead `getOldMessage` and `getMessage` should be used to compare the fields you want

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
@@ -5,15 +5,11 @@ import org.apache.logging.log4j.Logger;
 import org.javacord.api.DiscordApi;
 import org.javacord.api.entity.channel.ServerChannel;
 import org.javacord.api.entity.channel.TextChannel;
-import org.javacord.api.entity.message.Message;
-import org.javacord.api.entity.message.embed.Embed;
 import org.javacord.api.entity.server.Server;
 import org.javacord.api.event.message.CachedMessagePinEvent;
 import org.javacord.api.event.message.CachedMessageUnpinEvent;
 import org.javacord.api.event.message.MessageEditEvent;
 import org.javacord.core.entity.message.MessageImpl;
-import org.javacord.core.entity.message.embed.EmbedBuilderDelegateImpl;
-import org.javacord.core.entity.message.embed.EmbedImpl;
 import org.javacord.core.event.message.CachedMessagePinEventImpl;
 import org.javacord.core.event.message.CachedMessageUnpinEventImpl;
 import org.javacord.core.event.message.MessageEditEventImpl;
@@ -22,8 +18,6 @@ import org.javacord.core.util.gateway.PacketHandler;
 import org.javacord.core.util.logging.LoggerUtil;
 
 import java.time.OffsetDateTime;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
@@ -32,6 +26,7 @@ import java.util.concurrent.TimeUnit;
  * Handles the message update packet.
  */
 public class MessageUpdateHandler extends PacketHandler {
+    private static final int MAX_EDIT_TIMEOUT = 30000;
 
     /**
      * The logger of this class.
@@ -54,7 +49,7 @@ public class MessageUpdateHandler extends PacketHandler {
         api.getThreadPool().getScheduler().scheduleAtFixedRate(() -> {
             try {
                 lastKnownEditTimestamps.entrySet().removeIf(
-                        entry -> System.currentTimeMillis() + offset - entry.getValue() > 5000);
+                        entry -> System.currentTimeMillis() + offset - entry.getValue() > MAX_EDIT_TIMEOUT);
             } catch (Throwable t) {
                 logger.error("Failed to clean last known edit timestamps cache!", t);
             }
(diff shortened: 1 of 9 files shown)


javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
```diff
@@ -75,45 +68,11 @@ public void handle(JsonNode packet) {
...
-                            msg.getChannel(),
-                            event);
-                }
-            }
-        });
+        MessageImpl oldMessage = message.map(MessageImpl::copyMessage).orElse(null);
+        MessageImpl newMessage;
 
         MessageEditEvent editEvent = null;
         if (packet.has("edited_timestamp") && !packet.get("edited_timestamp").isNull()) {
```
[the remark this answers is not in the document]
KILLEliteMaste: > There is no guarantee from Discord that message update packets for embedding links don't send this field

I would say it is implicitly documented. Embedding links is not editing the message, rather it's just completing the message so thence there is no edit timestamp. 

> From what I read this code path also assumes that if edited_timestamp exists, all other message fields are present

Yes , if this is not the case it is simply not possible at all to detect what has changed especially when you haven't cached the message previously. 

> It also has to handle that embeds etc may be updated while edited_timestamp exists, for instance because the links were already cached by discord so it immediately embeds them without a second message update.

Not sure what you mean. When the links get immediately embedded, they will still get updated regardless which route it goes though (message present / create new message).

> This also doesn't fully handle components, which can be updated both with and without edited_timestamp in response to interaction updates.

In what case exactly? 

> Is there a good reason for doing it like this, rather than just fetching the old message from cache, cloning and updating it with the fields that are present in the payload; and then check if oldContent != newContent etc for fields

Yes, the goal was to make the message update handler dynamic so only the MessageImpl has to be updated when there are new message fields. Doing it your mentioned way is bad because we always need to remember to not forget to update the message update handler(it has not been updated since 4 years because of that!). It is also more work to maintain the library which I tried to avoid with this PR.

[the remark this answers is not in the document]
wasdennnoch: > I would say it is implicitly documented. Embedding links is not editing the message, rather it's just completing the message so thence there is no edit timestamp.

Is the process of embedding links documented somewhere? In my mind it's just a normal message update, and in a message update event one can expect any combination of fields to be present.

> Yes, if this is not the case it is simply not possible at all to detect what has changed especially when you haven't cached the message previously.

Isn't this already the case? if the message isn't cached the code already simply creates a new message and emits a message update with a `null` old message, without any attempts at determining what changed. If the message is changed, it compares a few of the fields if they are incldued in the update payload. Can't the same logic be applied to my idea?
And still the issue remains that if Discord at some point decides to send the `edited_timestamp` in a partial message update, the current code will probably break somewhere as not all message fields are present.
(do feel free to point me in the right direction if I'm missing something, you wrote the code after all and know it better than me 😄)

> Not sure what you mean.

It does seem to do that indeed already, just wanted to point out that that field needs the special handling; as explained below components will need that too.

> In what case exactly?

"Defer > Edit Content and Components" edits the deferred message without any edit timestamp, similarly to "Defer > Edit Content and Components > Edit Components", since only updating the components does not set `edited_timestamp`.
Contrary to that, "Defer > Edit Content and Components > Edit Content and Components" does set the timestamp as now something other than components got changed as well (in this example the content). In both examples, components got changed, but one has an edited timestamp and one doesn't.

> Doing it your mentioned way is bad because we always need to remember to not forget to update the message update handler

I mean, you can also move the change detection to `MessageImpl#setUpdatableFields` and make it return a boolean indicating whether a commonly human-edited field changed, that way you don't need to update the handler anymore :p
Dumb jokes aside (even though it's a possible implementation), while I'm all-in for not overcomplicating things, I think it should be easy enough to remember to update the message create/update/delete handlers when new message fields get introduced 😄
I'd argue that overall it would make the maintenance easier, because that way you avoid issues like the one I mentioned where some stuff is only handled in one case but not the other, and don't depend on undocumented behavior that might change at any time, potentially causing additional maintenance work down the line to fix things.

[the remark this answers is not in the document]
KILLEliteMaste: > Is the process of embedding links documented somewhere? In my mind it's just a normal message update, and in a message update event one can expect any combination of fields to be present.

No but you can easily verify how this update looks like. The payload contains just the must have fields + the complete embeds

> And still the issue remains that if Discord at some point decides to send the edited_timestamp in a partial message update, the current code will probably break somewhere as not all message fields are present.

Sure, at some point discord may decide to change something but currently this is not the issue and has probably not been in the past years since Discord exists. We shouldn't overcomplicate this and try to guess what they may be doing if they are never going to do this. If they decide to change something, we can still adjust it.

> "Defer > Edit Content and Components" edits the deferred message without any edit timestamp...

>  I think it should be easy enough to remember to update the message create/update/delete handlers when new message fields get introduced 😄
I'd argue that overall it would make the maintenance easier, because that way you avoid issues like the one I mentioned where some stuff is only handled in one case but not the other, and don't depend on undocumented behavior that might change at any time, potentially causing additional maintenance work down the line to fix things.

I see what you mean. Yeah, it's hard to match all cases. As a workaround to this we could check if the message contains a type of `19` or I think what's even better is to check if the message contains the `content` field as this has always to be present(except in the special cases where only partial fields are present) and then we know that it has to be complete message. This would only break if Discord decides to dispatch events that only contain the content but I think that's unlikely and even then we can change it to something else.
But I don't think that your approach is especially better to maintain. Yes, the current approach introduces *possibilities* where it can break but at the same time is forgetting to adjust the message update handler again and the additional workload of adding the methods and checks is also not optimal.

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
```diff
@@ -127,50 +86,62 @@ public void handle(JsonNode packet) {
...
-
-            if (oldEmbeds != null && newEmbeds != null) {
-                if (newEmbeds.size() != oldEmbeds.size()) {
-                    isMostLikelyAnEdit = true;
+        } else {
+            //Handle special cases where the edited_timestamp is not present or null like embedding links
+            if (message.isPresent()) {
+                newMessage = message.get();
+                if (!newMessage.setSpecialUpdatableFields(packet)) {
+                    logger.warn("Received a not handled special message update packet {}", packet.toString());
```
bassner, line 134: Shouldn't we return here? If no "special fields" have been updated here, nothing has been updated, and we would dispatch an event where the old and new message are exactly identical ... I know that the title of the PR is "dispatch every event", but 5 lines later we do return as well for "not meaningful cases", and I'd argue that this is not meaningful as well

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
```diff
@@ -127,50 +86,62 @@ public void handle(JsonNode packet) {
...
+                }
+            } else {
+                if (packet.path("edited_timestamp").isNull()) {
```
bassner, line 137: Why do we check this here? 
As far as I got it we'd reach this point if
- the message is not in the cache, and
- the edited timestamp is missing or null

so the condition would be always true, right?

By the way, which case this does entire ``else`` block cover? Is there any case where the edited timestamp is null AND we're getting a full message object (as opposed to a partial update for embedded links, for example)?

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
KILLEliteMaste: You are right. The title should probably be "dispatch an event for every message field and known special cases"

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
KILLEliteMaste: I just wrote a long text about that you are correct and that we should never receive a full message in a special case when I suddenly got a full message again of course... Which was the reason why I did this. As you can see the `edit_timestamp` is null and the message has not been cached before as it seems

```json
{
    "type": 19,
    "tts": false,
    "timestamp": "2022-08-06T18:26:09.303000+00:00",
    "pinned": true,
    "message_reference": {
      "message_id": "ID",
      "guild_id": "ID",
      "channel_id": "ID"
    },
    "mentions": [
      {
        "username": "NAME",
        "public_flags": 64,
        "member": {
          "roles": [
            "ID",
            "ID",
            "ID",
            "ID",
            "ID"
          ],
          "premium_since": null,
          "pending": false,
          "nick": "?????????",
          "mute": false,
          "joined_at": "2022-01-05T10:20:24.570000+00:00",
          "flags": 0,
          "deaf": false,
          "communication_disabled_until": null,
          "avatar": null
        },
        "id": "ID",
        "discriminator": "2825",
        "avatar_decoration": null,
        "avatar": "AVATAR"
      }
    ],
    "mention_roles": [],
    "mention_everyone": false,
    "member": {
      "roles": [
        "ID",
        "ID"
      ],
      "premium_since": null,
      "pending": false,
      "nick": "NAME",
      "mute": false,
      "joined_at": "2022-02-08T22:53:42.203000+00:00",
      "flags": 0,
      "deaf": false,
      "communication_disabled_until": null,
      "avatar": null
    },
    "id": "ID",
    "flags": 0,
    "embeds": [],
    "edited_timestamp": null,
    "content": "",
    "components": [],
    "channel_id": "ID",
    "author": {
      "username": "NAME",
      "public_flags": 0,
      "id": "ID",
      "discriminator": "5185",
      "avatar_decoration": null,
      "avatar": "AVATAR"
    },
    "attachments": [
      {
        "width": 512,
        "url": "https://cdn.discordapp.com/attachments/filename.mov",
        "size": 2488103,
        "proxy_url": "https://media.discordapp.net/attachments/filename.mov",
        "id": "ID",
        "height": 640,
        "filename": "filename.mov",
        "content_type": "video/quicktime"
      }
    ],
    "guild_id": "ID"
  }
```

RealYusufIsmail:
Just adding this here, just incase it relates to this issue. When I do
`logger.debug("Current attachments: {}", this.attachments.size());` in MessageBuilderBaseDelegateImpl returns as 0 were it is expected to return as 2 as there is 2 attachments already in the message.

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageUpdateHandler.java
bassner: What was the action on Discord that led to this event? :D

Anyways, Javacord would with the current code here behave very differently just based on the fact wether or not the message was cached before:
- If cached: Update special fields only, ignore other fields
- If not cached: Construct new message (effectively using all fields from the packet)
and I don't see how we could justify this difference in behavior.
It should more or less happen the same - also, what happens if we try to construct a new message object from a very sparse event?

Can't we just update all fields of the message that we receive, regardless of wether or not the timestamp is present, if it is cached? 
Maybe similar to this logic:
```
- Packet comes in
- If message cached: Update all fields using all data that is in the packet; leave data fields not contained in the packet untouched
- If message not cached:
  - If packet contains full message: Construct new message, add it to cache
  - If packet does not contain full message: Fetch message from API, add to cache, use that to create new update event
 
 In all cases, keep the approach to check for and compare the edit timestamp to tell if this is a real edit or somekind of special case
 ```
 Not 100% sure about the "fetch from API" part but I don't see a viable alternative if we want to dispatch an event in this last case with a fully populated message (which a user can reasonably expect if they get passed a ``Message`` object).
In any case we can't just blindly try to construct a message from a potentially partial packet in this last ``else`` branch, we need to first check if that can at all work and yield a useful, complete result

KILLEliteMaste: > What was the action on Discord that led to this event?

As the message has a message reference and is pinned, I tested to pin a message and it looks like its the same behavior. The message is exactly the same except for the fields shown. Left side is the MSG create and right is the update event when pinning the message. As the message reference can be null or not present, it would result in the same value when we construct the message. Only the nonce would be missing but this is not used anywhere in Javacord and its only usability seems to be for validating messages, not sure how important this is for library users.
image

> Anyways, Javacord would with the current code here behave very differently just based on the fact wether or not the message was cached before:

Not really. The message should be complete, otherwise you will never know what exactly has changed in a message if you haven't cached it before. I think I mentioned something like this in the discussing we had in our Discord.

> It should more or less happen the same - also, what happens if we try to construct a new message object from a very sparse event?

Wouldn't work with the current implementation as fields like the type are mandatory when you create a message and this field is not included in i.e. the embedding links update.

> Can't we just update all fields of the message that we receive, regardless of wether or not the edit timestamp is present, if it is cached?

As it is currently assumed that the message is complete as previously mentioned, the `setUpdateableFields` method is different to the `setSpecialUpdatableFields` because i.e. it always gets the content as it always has to be present on a message. So I think we could rewrite it to that, but we would still need to check if it is a complete message with checking the `edited_timestamp`.

> Not 100% sure about the "fetch from API" part

Not an option at all imo. Larger bots might end up requesting a lot of messages in that case.

> In any case we can't just blindly try to construct a message from a potentially partial packet in this last else branch, we need to first check if that can at all work and yield a useful, complete result

Not sure how we would achieve a check. Atm I would assume the packet is complete and I haven't seen anything different than that. Used my bot with a couple million users for probably 24h (not in one session but in total) and I haven't received any exceptions that the construction of a message object failed

PR #1052, opened 2022-07-20
Handle user update gateway packets

wasdennnoch:
`USER_UPDATE` is dispatched when the current user (i.e. the bot) is updated, e.g. to update the username or avatar. Previously the packet would just be ignored, meaning `DiscordApi#getYourself()` would return stale data, as the underlying field was only ever set once, in the `READY` handler (with the same line of code I used here).

Now, the question remains - should this also emit a fully fledged Javacord event? You can already receive `GUILD_MEMBER_UPDATE` events when a user changes, so technically you can aready listen to self-updates; just `getYourself()` would be outdated.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/user/UserUpdateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/user/UserUpdateHandler.java
@@ -2,6 +2,8 @@ package org.javacord.core.util.handler.user;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import org.javacord.api.DiscordApi;
+import org.javacord.core.entity.user.MemberImpl;
+import org.javacord.core.entity.user.UserImpl;
 import org.javacord.core.util.gateway.PacketHandler;
 
 /**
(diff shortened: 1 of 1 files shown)


wasdennnoch:
From [my testing](https://im-a-dolphin.de/s/e6a5370.png), I always receive `GUILD_MEMBER_UPDATE` events in all guilds if the bot user is updated. This also aligns with documentation, i.e. https://discord.com/developers/docs/topics/gateway#caveats:

> Guild Member Update is sent for **current-user updates** regardless of whether the `GUILD_MEMBERS` intent is set. [emphasis mine]

And https://discord.com/developers/docs/topics/gateway#guild-member-update:

> Sent when a guild member is updated. This will also fire when the **user object of a guild member changes**. [emphasis mine]

I suppose it could be possible to not receive any GMEs if your bot is not in any guilds - but if your bot is in 0 guilds it's a pretty useless bot since you can't interact with it at all :D Can't even DM it as you don't share any servers and the bot would be uncached in the client.

Saladoc (approved):
That's exactly what I got from the docs. The "Bot in no servers" example is niche enough that it won't do any harm if the Bot doesn't receive an event for it. If the need arises, we can always add another event in the future.

PR #1061, opened 2022-07-20
Use correct user object in user update event

wasdennnoch:
The inner payload already is a full user object 🥲  tfw the change is so small you think to yourself "it's fine".

I swear I actually tested this one!

_sorry_


Saladoc (approved):

Saladoc:
@all-contributors please add @wasdennnoch for code

PR #1067, opened 2022-07-21
Required workflows should now really skip ci for all contributors

KILLEliteMaste:


KILLEliteMaste:
@Bastian the CI should now skip PRs and commits by the all contributors bot. 
I followed [this](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks) and used the [paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths) to determine whether the workflow should run. 
Now I am not sure if you have to setup anything in the settings that the new workflow runs in case the normal `main` does not

Bastian:
> Now I am not sure if you have to setup anything in the settings that the new workflow runs in case the normal main does not

I don't think so. This should work without any changes in the settings.

Bastian (approved):

PR #1070, opened 2022-07-31
Expose `getServerBoostingSinceTimestamp` and `getServerAvatarHash`

wasdennnoch:
Previously, one could only get a full avatar URL from a member, instead of just the hash like you can already do with a normal user avatar. The boosting timestamp wasn't even exposed at all, it was just internal to `Member`, which was probably an oversight (it was added in the member intent refactor and never touched again).
Since I don't see a reason to keep them private, let's expose them! From what I see these were the only missing member fields, so all Discord member fields should be exposed now.


AhmedA1559 (approved):

KILLEliteMaste (approved):
Thank you, LGTM!

KILLEliteMaste:
Please rebase on the latest version because I can not do it by myself for some reason

Issue #1069, opened 2022-07-27
Support descriptions of message attachments


Saladoc:
[Message Attachments](https://discord.com/developers/docs/resources/channel#attachment-object) may have a description attribute.

This should be supported in our `MessageAttachment`s.

shubhrjain:
Hello @Saladoc Can I pick this issue as I am first time open source contributor ?

Saladoc:
Go ahead, open a PR. If you need help, find us on the linked discord server.

PR #1072, opened 2022-08-04
Add Support for Description in Attachments

shubhwip:
Issue Link : https://github.com/Javacord/Javacord/issues/1069 
[Message Attachments](https://discord.com/developers/docs/resources/channel#attachment-object) may have a description attribute.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
@@ -40,6 +40,11 @@ public class AttachmentImpl implements Attachment {
     private final String fileName;
 
     /**
+     * The description of the attachment.
+     */
+    private final String description;
+
+    /**
      * The size of the attachment in bytes.
      */
     private final int size;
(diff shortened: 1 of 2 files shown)


javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -102,6 +108,11 @@ public String getFileName() {
         return fileName;
     }
 
+    @Override
+    public String getDescription() {
+        return description;
```
montlikadani, line 113: Make this method as `Optional` as the `description` might return `null`.

shubhwip: Nice suggestion !!

KILLEliteMaste (approved):
Thank you for implementing the Message Attachment Description! :) LGTM!
Just a small note: Next time please squash your commits into a single commit. This helps us to have maintain a cleaner commit history

KILLEliteMaste:
@all-contributors please add @shubhwip for code

PR #1073, opened 2022-08-05
Updated discord api to v10

### LINKED ISSUE #960: Upgrade to gateway version `10`

RealYusufIsmail:
I have updated the discord api version to 10

closes #960

TODO:
- [x] tweak identifying payload
- [x] add all the changes for the new version.

attachments: 
- [x] Update the message and keep existing attachments (default behaviour to keep consistent with how it currently works)
- [x] Remove (all/partial) attachments -same as 4
- [x] Add new attachments to existing attachments
- [x] Remove and add new attachments at the same time
- [x] Choose which attachments to keep 
- [x] Fix the bug were it does not detect the current attachments
- [x] Interaction support

I have tested most of it and it works. 

Would appircate if someone double checks the interaction bit please

### Changelog
- Add `MESSAGE_CONTENT` intent
- Upgraded to Gateway v10
- MessageUpdater can now remove specific attachments (with `#removeExistingAttachments`) and add new attachments at the same time


javacord-core/src/main/java/org/javacord/core/entity/message/InteractionMessageBuilderDelegateImpl.java
```diff
@@ -82,21 +87,43 @@ public CompletableFuture<Message> sendFollowupMessage(InteractionBase interactio
...
+            for (Attachment attachment : attachmentsToKeep) {
+
+                InputStream stream = null;
+                try {
+                    stream = attachment.asInputStream();
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+
+                tempAttachments.remove(new FileContainer(stream, attachment.getFileName()));
```
RealYusufIsmail, line 119: Removes the attachments that are wanted. Change this

RealYusufIsmail:
Apart from a few changes I am pretty much done. @KILLEliteMaste can you review this please.

KILLEliteMaste:
Why did you close #994 and opened a new PR?

RealYusufIsmail:
> Member

it was messy and some stuff was messed up.

KILLEliteMaste:
Not a valid reason imo. It's not like the PR is deleted. Anyway continue in this PR for now and please do not open unnecessary PRs

javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
```diff
@@ -159,7 +159,30 @@ public enum Intent {
      *     <li>TYPING_START</li>
      * </ul>
      */
-    DIRECT_MESSAGE_TYPING(14, false);
+    DIRECT_MESSAGE_TYPING(14, false),
+
+    /**
+     * This is required to receive non-empty values for content fields (content, attachments, embeds, and components).
```
KILLEliteMaste, line 165:
```suggestion
     * This is required to receive non-empty values for content fields (content, attachments, embeds and components).
```

javacord-api/src/main/java/org/javacord/api/entity/intent/Intent.java
```diff
@@ -159,7 +159,30 @@ public enum Intent {
...
+     * The following events are received.
+     * <ul>
+     *     <li>GUILD_SCHEDULED_EVENT_CREATE</li>
+     *     <li>GUILD_SCHEDULED_EVENT_UPDATE</li>
+     *     <li>GUILD_SCHEDULED_EVENT_DELETE</li>
+     *     <li>GUILD_SCHEDULED_EVENT_USER_ADD</li>
+     *     <li>GUILD_SCHEDULED_EVENT_USER_REMOVE</li>
+     * </ul>
+     */
+    GUILD_SCHEDULED_EVENTS(16, false);
```
KILLEliteMaste, line 185: Remove it. Should be added in the PR which adds support for guild scheduled events

RealYusufIsmail:
> Not a valid reason imo. It's not like the PR is deleted. Anyway continue in this PR for now and please do not open unnecessary PRs

ok

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
```diff
@@ -1568,11 +1568,16 @@ public CompletableFuture<Void> kickUser(User user, String reason) {
     @Override
     public CompletableFuture<Void> banUser(String userId, int deleteMessageDays, String reason) {
         RestRequest<Void> request = new RestRequest<Void>(getApi(), RestMethod.PUT, RestEndpoint.BAN)
-                .setUrlParameters(getIdAsString(), userId)
-                .addQueryParameter("delete_message_days", String.valueOf(deleteMessageDays));
+                .setUrlParameters(getIdAsString(), userId);
+
+        if (deleteMessageDays > 0) {
+            request.addQueryParameter("delete_message_days", String.valueOf(deleteMessageDays));
+        }
```
KILLEliteMaste, line 1575: Why is this change necessary?

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
```diff
@@ -43,25 +45,54 @@ public interface InteractionMessageBuilderDelegate extends MessageBuilderBaseDel
...
+     * @return The completable future to determine if the message was updated.
+     */
+    default CompletableFuture<Void> updateOriginalMessage(InteractionBase interaction, List<Attachment> attachments,
+                                                          boolean newOrKeepAttachments) {
+        if (newOrKeepAttachments) {
+            return updateOriginalMessage(interaction, attachments, null);
+        } else {
+            return updateOriginalMessage(interaction, null, attachments);
+        }
+    }
```
KILLEliteMaste, line 88: Same here

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
```diff
@@ -43,25 +45,54 @@ public interface InteractionMessageBuilderDelegate extends MessageBuilderBaseDel
...
+     * Edit the message the component was attached to.
+     *
+     * @param interaction The interaction.
+     * @param attachmentsToKeep used to keep specific attachments, and remove the others.
+     * @param addNewAttachments used to add new attachments.
+     * @return The completable future to determine if the message was updated.
+     */
+    CompletableFuture<Void> updateOriginalMessage(InteractionBase interaction, List<Attachment> attachmentsToKeep,
+                                                  List<Attachment> addNewAttachments);
+
```
KILLEliteMaste, line 62: What is this for? It is not usable from anywhere except in the delegate

KILLEliteMaste (changes requested):
The attachment changes in you current implementation are only available in interactions. You are missing updating attachments for normal messages.

Did you really test your changes with actual code if they meet the requirements discussed in our Discord?

javacord-core/src/main/java/org/javacord/core/entity/server/ServerImpl.java
RealYusufIsmail: Screenshot 2022-08-06 at 16 03 12
no longer need to add deleteMessageDays unless it is above 0

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
RealYusufIsmail: This is used to update the message. were the list attachmentsToKeep is used to set the attachments you want to keep and addNewAttachments is used to add new attachments.

If both are null, the message will stay the same and have the same attachments

KILLEliteMaste: You don't understand me. Why are there 3 methods in the delegate if only 1 is used at all in the InteractionMessageUpdater or what it was called where the delegate is used?

RealYusufIsmail: ah i see

RealYusufIsmail (commented):
Ignore these ico

RealYusufIsmail:
Blocked until #1095 is merged

RealYusufIsmail:
https://user-images.githubusercontent.com/67903886/183638127-762186a0-9d75-4d11-8ef9-f04d3297933e.mov


https://user-images.githubusercontent.com/67903886/183638145-60eb4fbb-131f-4505-bfac-d8f685c3c5c0.mov

RealYusufIsmail:
almost done

RealYusufIsmail:
Screenshot 2022-08-09 at 18 43 58
first do not keep attachments
seconds keep attachments

work

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -398,6 +401,42 @@ default CompletableFuture<Message> edit(String content) {
...
+     * }</pre>
+     *
+     * @return A future to check if the update was successful.
+     */
+    default CompletableFuture<Message> edit(String content, Collection<Attachment> keepAttachments,
+                                            Collection<File> newAttachments) {
+        return new MessageUpdater(this).setContent(content, keepAttachments, newAttachments).applyChanges();
+    }
+
+
```
KILLEliteMaste, line 439: I would not add these methods, because then you would have to add it for every other too which clutters the interface a lot. I think it's okay to redirect users to use the MessageUpdater if they want to perform more complex updates. Instead you should add methods for only changing the attachments.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
```diff
@@ -169,6 +171,36 @@ public T setContent(String content) {
...
+     *                        the attachments in the message.
+     * @param newAttachments New Attachments you want to add to the message.
+     *
+     * @return The current instance in order to chain call methods.
+     */
+    public T setContent(String content, Collection<Attachment> keepAttachments, Collection<File> newAttachments) {
+        delegate.setContent(content, keepAttachments, newAttachments);
+        return myClass.cast(this);
+    }
+
```
KILLEliteMaste, line 203: This seems extremely inflexible as you can not change the attachments without changing or setting the content again. Why didn't you add new methods like `keepAllAttachments` and `keepAttachments(Attachment...)` / `keepAttachments(Collection<Attachment>`?

javacord-api/src/main/java/org/javacord/api/entity/message/internal/MessageBuilderBaseDelegate.java
```diff
@@ -155,6 +177,11 @@ public interface MessageBuilderBaseDelegate {
      */
     void removeAllComponents();
 
+    /**
+     * Removes all the attachments from the message.
```
KILLEliteMaste, line 181:
```suggestion
     * Removes all attachments from the message.
```

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -199,6 +201,66 @@ public void setContent(String content) {
...
+                //remove everything else apart from fileContainer
+                this.attachments.removeIf(fileContainer1 -> !fileContainer1.equals(fileContainer));
+
+                this.attachments.add(fileContainer);
+                attachmentsChanged = true;
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
```
KILLEliteMaste, line 262: With this approach you are reuploading all files instead of keeping them I think. As the docs state, you need to set the attachments you want to keep in the `attachments` array. For that you will most likely need a new variable `List<Attachments>` and then add the attachments you want to keep to it and then finally in the JSON body when sending the request
https://discord.com/developers/docs/resources/channel#edit-message-jsonform-params

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
RealYusufIsmail: Thats not true. I tested it and it keep the file that i stated to keep and remove the others

RealYusufIsmail: The part of setting that attachments you want to keep is done by the programmer

KILLEliteMaste: You may re-read what I wrote... But to be more exact:
You are **re-uploading** the files instead of keeping them. This is a significant difference.
You can easily check this by opening the image in the browser so you get the CDN link. After editing the message, the link is not the same

RealYusufIsmail: https://user-images.githubusercontent.com/67903886/184222722-cf4f2c31-5b98-45ee-83e3-098f4f753b01.mov

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -94,7 +97,17 @@ public class MessageBuilderBaseDelegateImpl implements MessageBuilderBaseDelegat
     /**
      * All attachments which should be added to the message.
      */
-    protected final List<FileContainer> attachments = new ArrayList<>();
+    protected final List<FileContainer> newAttachments = new ArrayList<>();
+
+    /**
+     * The current attachments of the message.
+     */
+    protected final List<Attachment> attachments = new ArrayList<>();
```
KILLEliteMaste, line 105: This is not needed, because you get the message as an argument in the update function. This should most likely be a boolean for keeping all attachments / remove all attachments. 
By default it's `true`. If `false`, remove all and if it's `true` and `attachmentsToKeep` contains something, just keep the attachments in the list

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -196,9 +209,37 @@ public void appendNewLine() {
...
     }
 
+    private void keepAllAttachments() {
+        attachmentsToKeep.addAll(new ArrayList<>(this.attachments));
+    }
+
+    @Override
+    public void keepAttachments(Attachment... attachments) {
+        keepAttachments(Arrays.asList(attachments));
+    }
```
KILLEliteMaste, line 223: Move the logic to the updater instead so no implementation is necessary

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -196,9 +209,37 @@ public void appendNewLine() {
...
+
+    @Override
+    public void keepAttachments(Collection<Attachment> keepAttachments) {
```
KILLEliteMaste, line 226: The methods should probably be renamed to `addKeepAttachments`. It is not really clear what this methods does with the current implementation(you are clearing the list) and then you could also add `setKeepAttachments` so someone can overwrite the list

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -196,9 +209,37 @@ public void appendNewLine() {
...
+        for (Attachment keepAttachment : keepAttachments) {
+            attachments.forEach(a -> {
+                if (Objects.equals(a.getFileName(), keepAttachment.getFileName())) {
+                    attachmentsList.add(a);
+                }
+            });
+        }
+
+        attachmentsToKeep.addAll(attachmentsList);
+    }
```
KILLEliteMaste, line 241: You can not check if the name is the same because you can set description for new files as the docs state
https://discord.com/developers/docs/resources/webhook#edit-webhook-message-jsonform-params
You will also not be able to do this once you remove the attachments variable because you get the attachments already via the message in the edit method.

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
RealYusufIsmail: i changed a bit so no need for this

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
```diff
@@ -169,6 +171,17 @@ public T setContent(String content) {
...
+    public T setAttachmentsToKeep(Attachment... attachments) {
+        setAttachmentsToKeep(Arrays.asList(attachments));
+        return myClass.cast(this);
+    }
+
+    public T setAttachmentsToKeep(Collection<Attachment> attachments) {
+        delegate.setAttachmentsToKeep(attachments);
+        return myClass.cast(this);
+    }
+
```
KILLEliteMaste, line 184: These methods are still here... And they are also usable for i.e. the `InteractionImmediateResponseBuilder` where they are useless

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -197,6 +209,19 @@ public void setContent(String content) {
         strBuilder.setLength(0);
         strBuilder.append(content);
         contentChanged = true;
+        keepAttachments = true;
```
KILLEliteMaste, line 212: Not needed

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
```diff
@@ -169,6 +171,17 @@ public T setContent(String content) {
         return myClass.cast(this);
     }
 
+
+    public T setAttachmentsToKeep(Attachment... attachments) {
+        setAttachmentsToKeep(Arrays.asList(attachments));
+        return myClass.cast(this);
+    }
+
+    public T setAttachmentsToKeep(Collection<Attachment> attachments) {
```
KILLEliteMaste, line 180: In the implementation you do not set the attachments, you are just adding them to the list. 
Should be renamed to `add*`

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -574,6 +599,12 @@ public CompletableFuture<Message> edit(Message message, boolean updateAll) {
 
         prepareComponents(body, updateAll || componentsChanged);
 
+        if (keepAttachments) {
+            prepareAttachments(message.getAttachments(), body);
+        } else {
+            body.set("attachments", JsonNodeFactory.instance.objectNode().arrayNode());
+        }
```
KILLEliteMaste, line 606: What happens if you do not set the attachments array empty? I always read the documentation that the attachments will still be removed

Mysterypotatoguy:
Just my two cents, I'm not sure if I missed a conversation somewhere, but having to always specify keeping attachments seems like an odd design. I would expect by default a message edit keeps all previous attachments.
My suggestion is just to have `addAttachment`/`removeAttachment`/`removeAllAttachments` and a `getAttachments` to be able to access attachments to remove.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -644,6 +675,26 @@ private void prepareAllowedMentions(ObjectNode body) {
...
+            }
+        } else if (!attachmentsList.isEmpty()) {
+            for (MessageAttachment attachment : attachmentsList) {
+                attachments.add(((AttachmentImpl) attachment).toJsonNode());
+            }
+        } else if (!attachmentsToKeep.isEmpty()) {
+            for (Attachment attachment : attachmentsToKeep) {
+                attachments.add(((AttachmentImpl) attachment).toJsonNode());
+            }
+        }
```
KILLEliteMaste, line 695: Did you test your new changes? This should not work as you check if the list of attachments contains the id of the attachment in the first if statement.

Anyway, the logic seem wrong to me. 
1. The method gets executed when we want to keep the attachments. 
2. `if(!attachmentToKepp.isEmpty())` then add only the attachments from the collection
3. Or else the list is empty, then add all attachments from the message

RealYusufIsmail:
> Just my two cents, I'm not sure if I missed a conversation somewhere, but having to always specify keeping attachments seems like an odd design. I would expect by default a message edit keeps all previous attachments. My suggestion is just to have `addAttachment`/`removeAttachment`/`removeAllAttachments` and a `getAttachments` to be able to access attachments to remove.

By deafult attachments are kept

RealYusufIsmail:
Screenshot 2022-08-12 at 17 28 05
The add part is done through message builder and KILLEliteMaste said no to this. To keep specific attachments I added a method for that @Mysterypotatoguy

KILLEliteMaste (changes requested):
Please do not ignore requested changes and mark them as resolved if you did no apply them. You did multiple times and now again with the methods in the `MessageBuilderBase`. 
Reviews are a really time consuming task and it is quiet exhausting if I have to mark the same issues over and over

Mysterypotatoguy:
Okay, I've re-read and I think I understand better now. In this case I think `setAttachmentsToKeep` is a misleading name and implies you need to always call it when editing. What about `removeAttachmentsExcept`? 
Or better yet if we can somehow get access to the message object (might involve splitting edit logic out) then there can just be a single internal list of attachments that can be edited instead of `newAttachments` and `attachmentsToKeep`, then what I said before is possible

RealYusufIsmail:
> Please do not ignore requested changes and mark them as resolved if you did no apply them. You did multiple times and now again with the methods in the `MessageBuilderBase`. Reviews are a really time consuming task and it is quiet exhausting if I have to mark the same issues over and over

I apologise and will not do this next time

KILLEliteMaste:
> Okay, I've re-read and I think I understand better now. In this case I think `setAttachmentsToKeep` is a misleading name and implies you need to always call it when editing. What about `removeAttachmentsExcept`? Or better yet if we can somehow get access to the message object (might involve splitting edit logic out) then there can just be a single internal list of attachments that can be edited instead of `newAttachments` and `attachmentsToKeep`, then what I said before is possible

Currently the message is always passed into the MessageUpdater so we get the attachments from there. 

I think it would be nice to have both methods. But we need to clear method names to not confuse them with `addAttachment` which adds a new attachment to the message. Maybe we should also rename the method `addAttachmentToKepp` to just `retainAttachment(s)` and it's quiet shorter. 
So maybe we have:
- `retainAttachment(s)`
- `retainAllAttachments`
- `removeAttachmentsExcept`
- `removeAllAttachments`

>  then there can just be a single internal list of attachments

I don't think this will work out as new files need to be handled [differently ](https://discord.com/developers/docs/resources/channel#edit-message) than just being added to the `attachments` array to keep files

Mysterypotatoguy:
`retainAttachments` sounds fine, `retainAllAttachments` is confusing again. To someone just reading the name does it need to be called every edit to retain existing attachments?
I'm also not sure what `removeRetainingAttachment` and `removeAllRetainingAttachments` should do, remove the *existing* attachments?

KILLEliteMaste:
> I'm also not sure what removeRetainingAttachment and removeAllRetainingAttachments should do, remove the existing attachments?

Basically just the opposite of the other method. 
`retainAttachments` = only keep the provided attachments
`removeRetainingAttachments` = remove the provided attachments, instead of only keeping them

> retainAttachments sounds fine, retainAllAttachments is confusing again

Maybe a bit, yes. But should there be no way to revert the retaining attachments? So then you are screwed once you use any retainAttachment method. We can still mention in the Javadocs that this is the default behavior.

Mysterypotatoguy:
If you do make a mistake like that, you can just throw away the edit without executing it. I don't see why in static code you would need to be able to revert something like that, unless I am misunderstanding what you are saying.
For example in MessageBuilder there is no option to remove content once it has been added because it makes no sense, you can just throw away the MessageBuilder (or correct the code)

IMO the API should be
`addAttachment`
`addAttachments`
`removeExistingAttachment`
`removeExistingAttachments`
`removeExistingAttachmentsExcept`

KILLEliteMaste:
> If you do make a mistake like that, you can just throw away the edit without executing it. I don't see why in static code you would need to be able to revert something like that, unless I am misunderstanding what you are saying. For example in MessageBuilder there is no option to remove content once it has been added because it makes no sense, you can just throw away the MessageBuilder (or correct the code)
> 
> IMO the API should be `addAttachment` `addAttachments` `removeExistingAttachment` `removeExistingAttachments` `removeExistingAttachmentsExcept`

The names don't go well with adding attachment descriptions for new files 
image

Then we are basically stuck with only being able to remove attachments but not adding them to this array. Or should there be another method for adding these description? Then I think we need another variable to store specifically the attachments that should be newly added and that's also not that nice

Mysterypotatoguy:
Probably an object in itself, `AttachmentBuilder`?

KILLEliteMaste (changes requested):
We have a few more needed changes for the API
The methods should be renamed and added as follows:
`removeExistingAttachment(Attachment)`
`removeExistingAttachments()` (remove all attachments)
`removeExistingAttachments(Attachment...)`
`removeExistingAttachments(Collection<Attachment>)`

Overload the `addAttachment` methods with a `String` description parameter i.e. `addAttachment(File file)` will be overlaoded with`addAttachment(File file, String description)`. In the implementation we can only have the implementation with the description and the currently existing methods will call these methods with a `null` descrption. You most likely have to adjust the `Filecontainer` to also include an optional description. Then when also a description is set you have to add them to the `attachment` JSON where you specify the attachments that should be kept. To include new files with a description as previously mentioned the format has to look like 
```json
"id": n,
"description": "desription"
```
where n = the index number when adding a file in the `addMultipartBodyToRequest` method

The priority of this PR is getting higher due to the message content intent that will be enforced very soon. Do you have an estimate when you can finish this PR with the current new and older requested changes?

RealYusufIsmail:
> We have a few more needed changes for the API The methods should be renamed and added as follows: `removeExistingAttachment(Attachment)` `removeExistingAttachments()` (remove all attachments) `removeExistingAttachments(Attachment...)` `removeExistingAttachments(Collection<Attachment>)`
> 
> Overload the `addAttachment` methods with a `String` description parameter i.e. `addAttachment(File file)` will be overlaoded with`addAttachment(File file, String description)`. In the implementation we can only have the implementation with the description and the currently existing methods will call these methods with a `null` descrption. You most likely have to adjust the `Filecontainer` to also include an optional description. Then when also a description is set you have to add them to the `attachment` JSON where you specify the attachments that should be kept. To include new files with a description as previously mentioned the format has to look like
> 
> ```json
> "id": n,
> "description": "desription"
> ```
> 
> where n = the index number when adding a file in the `addMultipartBodyToRequest` method
> 
> The priority of this PR is getting higher due to the message content intent that will be enforced very soon. Do you have an estimate when you can finish this PR with the current new and older requested changes?

I can try to get this pr done in 2 days, but might take longer if something comes up. I will make it my main focus

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilderBase.java
```diff
@@ -388,6 +541,49 @@ public T removeAllComponents() {
...
+     * Removes multiple attachments from the message.
+     *
+     * @param attachments The attachments to remove.
+     * @return The current instance in order to chain call methods.
+     */
+    public T removeExistingAttachments(Collection<Attachment> attachments) {
+        delegate.removeExistingAttachments(attachments);
+        return myClass.cast(this);
+    }
+
```
KILLEliteMaste, line 586: I think this is now the 3rd time I am telling you that these methods can not be in this class because they are available in i.e. The MessageBuilder where they are useless. So they should only be available when **updating** a message
Just for example the following code does not have these methods but currently do. 
```java
new InteractionImmediateResponseBuilderImpl(null).removeExistingAttachments(Arrays.asList());
new MessageBuilder().removeExistingAttachments(Arrays.asList());
```

javacord-api/src/main/java/org/javacord/api/entity/message/internal/MessageBuilderBaseDelegate.java
```diff
@@ -162,6 +197,7 @@ public interface MessageBuilderBaseDelegate {
      */
     void setTts(boolean tts);
 
+
```
KILLEliteMaste, line 200: Remove line

javacord-api/src/main/java/org/javacord/api/entity/message/internal/MessageBuilderBaseDelegate.java
```diff
@@ -170,27 +206,61 @@ public interface MessageBuilderBaseDelegate {
...
+    void addAttachment(BufferedImage image, String fileName, String description);
+
     /**
      * Adds an attachment to the message.
      *
      * @param file The file to add as an attachment.
      */
     void addAttachment(File file);
 
+
```
KILLEliteMaste, line 225: Remove line

javacord-api/src/main/java/org/javacord/api/interaction/callback/InteractionMessageBuilder.java
```diff
@@ -278,69 +317,146 @@ public InteractionMessageBuilder addAttachment(BufferedImage image, String fileN
         return this;
     }
 
+
```
KILLEliteMaste, line 320: Remove line

javacord-api/src/main/java/org/javacord/api/interaction/callback/InteractionMessageBuilder.java
```diff
@@ -278,69 +317,146 @@ public InteractionMessageBuilder addAttachment(BufferedImage image, String fileN
...
+        return this;
+    }
+
     @Override
     public InteractionMessageBuilder addAttachment(Icon icon) {
         delegate.addAttachment(icon);
         return this;
     }
 
+
```
KILLEliteMaste, line 345: Remove line

KILLEliteMaste, line 345: Remove line

javacord-api/src/main/java/org/javacord/api/interaction/callback/InteractionMessageBuilder.java
```diff
@@ -278,69 +317,146 @@ public InteractionMessageBuilder addAttachment(BufferedImage image, String fileN
...
         return this;
     }
 
+    @Override
+    public InteractionMessageBuilder addAttachment(InputStream stream, String fileName, String description) {
+        delegate.addAttachment(stream, fileName, description);
+        return this;
+    }
+
+
```
KILLEliteMaste, line 388: Remove lines

javacord-api/src/main/java/org/javacord/api/interaction/callback/InteractionMessageBuilderBase.java
```diff
@@ -71,6 +74,54 @@
...
+     */
+    T removeExistingAttachments(Attachment... attachments);
+
+    /**
+     * Removes multiple attachments from the message.
+     *
+     * @param attachments The attachments to remove.
+     * @return The current instance in order to chain call methods.
+     */
+    T removeExistingAttachments(Collection<Attachment> attachments);
```
KILLEliteMaste, line 105: Should be the same as mentioned in the MessageBuilderBase

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -199,6 +216,36 @@ public void setContent(String content) {
...
+    public void removeExistingAttachments(Collection<Attachment> attachments) {
+        attachmentsToRemove.addAll(attachments);
+    }
+
+    @Override
+    public void addAttachmentsToKeep(Collection<Attachment> keepAttachmentsCollection) {
+        //as in setContent we keep all attachments, were as here we only keep the ones that are in the list
+        attachmentsToKeep.addAll(keepAttachmentsCollection);
+        keepAttachments = true;
+    }
```
KILLEliteMaste, line 247: We do not have this method anymore as this clashes with `attachmentsToRemove`. So we specify only the attachments we want to remove. So then you can remove the `attachmentsToKeep` variable as it is no longer needed

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -360,7 +484,16 @@ public void addAttachmentAsSpoiler(URL url) {
...
+        newAttachments.add(new FileContainer(url, true));
+        attachmentsChanged = true;
+    }
+
+    @Override
+    public void addAttachmentAsSpoiler(URL url, String description) {
+        if (url == null) {
+            throw new IllegalArgumentException("url cannot be null!");
+        }
+        newAttachments.add(new FileContainer(url, true, description));
```
KILLEliteMaste, line 496: I explicitly told you that the methods **without the description parameter** in the implementation are no longer needed. When using `addAttachmentAsSpoiler(URL)` in a MessageBuilder you can already set `null` to the description there. The same is for the other addAttachment methods above

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -199,6 +216,36 @@ public void setContent(String content) {
...
+    public void removeExistingAttachments() {
+        attachmentsToKeep.clear();
+        keepAttachments = false;
+    }
+
+    @Override
+    public void removeExistingAttachments(Attachment... attachments) {
+        removeExistingAttachments(Arrays.asList(attachments));
+        keepAttachments = false;
+    }
```
KILLEliteMaste, line 235: Method not needed. You can move the logic to the updater classes so we have a method less in the implementation

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -220,7 +267,11 @@ public void copy(Message message) {
 
         for (MessageAttachment attachment : message.getAttachments()) {
             // Since spoiler status is encoded in the file name, it is copied automatically.
-            this.addAttachment(attachment.getUrl());
+            if (attachment.getDescription().isPresent()) {
+                this.addAttachment(attachment.getUrl(), attachment.getDescription().get());
+            } else {
+                this.addAttachment(attachment.getUrl(), null);
+            }
```
KILLEliteMaste, line 274:
```suggestion
           this.addAttachment(attachment.getUrl(), attachment.getDescription().orElse(null));
```

javacord-core/src/main/java/org/javacord/core/interaction/ComponentInteractionOriginalMessageUpdaterImpl.java
```diff
@@ -5,7 +5,7 @@
 
 import java.util.concurrent.CompletableFuture;
 
-public class ComponentInteractionOriginalMessageUpdaterImpl
+public  class ComponentInteractionOriginalMessageUpdaterImpl
```
KILLEliteMaste, line 8: Remove space

javacord-core/src/main/java/org/javacord/core/util/FileContainer.java
```diff
@@ -90,23 +107,46 @@ public FileContainer(BufferedImage file, String type) {
      * @param file The file as a file.
      */
     public FileContainer(File file) {
-        this(file, false);
+        this(file, false, null);
     }
 
+
```
KILLEliteMaste, line 113: Remove line

javacord-core/src/main/java/org/javacord/core/util/FileContainer.java
```diff
@@ -143,20 +205,43 @@ public FileContainer(URL file) {
...
+     * Creates a new file container with an url.
+     *
+     * @param file The file as an url.
+     * @param description The description of the url.
+     */
+    public FileContainer(URL file, String description) {
+        this(file, false, description);
+    }
+
+
```
KILLEliteMaste, line 218: Remove line

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
RealYusufIsmail: sorry i think i mis read what you said

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -644,6 +681,20 @@ private void prepareAllowedMentions(ObjectNode body) {
...
+
+        if (removeAttachments) {
+            if (!attachmentsToRemove.isEmpty()) {
+                for (Attachment attachment : attachmentsList) {
+                    if (!attachmentsToRemove.contains(attachment)) {
+                        attachments.add(((AttachmentImpl) attachment).toJsonNode());
+                    }
+                }
+            }
+        }
```
KILLEliteMaste, line 695: If there are not attachments to remove, it means all attachments should be kept. You are missing the else case where all attachments are added that are kept.

javacord-core/src/main/java/org/javacord/core/interaction/ExtendedInteractionMessageBuilderBaseImpl.java
```diff
@@ -49,73 +49,145 @@ public T copy(InteractionBase interaction) {
...
+    @Override
+    public T addAttachmentAsSpoiler(URL url, String description) {
+        delegate.addAttachmentAsSpoiler(url, description);
         return myClass.cast(this);
     }
 
     @Override
     public T addAttachmentAsSpoiler(byte[] bytes, String fileName) {
-        delegate.addAttachment(bytes, "SPOILER_" + fileName);
+        addAttachment(bytes, "SPOILER_" + fileName);
```
KILLEliteMaste, line 172:
```suggestion
        addAttachment(bytes, "SPOILER_" + fileName, null);
```

javacord-core/src/main/java/org/javacord/core/interaction/ExtendedInteractionMessageBuilderBaseImpl.java
```diff
@@ -49,73 +49,145 @@ public T copy(InteractionBase interaction) {
...
+    @Override
+    public T addAttachmentAsSpoiler(byte[] bytes, String fileName, String description) {
+        delegate.addAttachment(bytes, "SPOILER_" + fileName, description);
         return myClass.cast(this);
     }
 
     @Override
     public T addAttachmentAsSpoiler(InputStream stream, String fileName) {
-        delegate.addAttachment(stream, "SPOILER_" + fileName);
+        addAttachment(stream, "SPOILER_" + fileName);
```
KILLEliteMaste, line 184:
```suggestion
        addAttachment(stream, "SPOILER_" + fileName, null);
```

javacord-core/src/main/java/org/javacord/core/interaction/InteractionOriginalResponseUpdaterImpl.java
```diff
@@ -167,73 +167,149 @@ public InteractionOriginalResponseUpdater setTts(boolean tts) {
...
+    @Override
+    public InteractionOriginalResponseUpdater addAttachment(InputStream stream, String fileName, String description) {
+        delegate.addAttachment(stream, fileName, description);
         return this;
     }
 
     @Override
     public InteractionOriginalResponseUpdater addAttachmentAsSpoiler(BufferedImage image, String fileName) {
-        delegate.addAttachment(image, "SPOILER_" + fileName);
+        addAttachment(image, "SPOILER_" + fileName);
```
KILLEliteMaste, line 242:
```suggestion
        addAttachment(image, "SPOILER_" + fileName, null);
```

javacord-core/src/main/java/org/javacord/core/interaction/InteractionOriginalResponseUpdaterImpl.java
```diff
@@ -167,73 +167,149 @@ public InteractionOriginalResponseUpdater setTts(boolean tts) {
...
+    @Override
+    public InteractionOriginalResponseUpdater addAttachmentAsSpoiler(URL url, String description) {
+        delegate.addAttachmentAsSpoiler(url, description);
         return this;
     }
 
     @Override
     public InteractionOriginalResponseUpdater addAttachmentAsSpoiler(byte[] bytes, String fileName) {
-        delegate.addAttachment(bytes, "SPOILER_" + fileName);
+        addAttachment(bytes, "SPOILER_" + fileName);
```
KILLEliteMaste, line 291:
```suggestion
        addAttachment(bytes, "SPOILER_" + fileName, null);
```

javacord-core/src/main/java/org/javacord/core/interaction/InteractionOriginalResponseUpdaterImpl.java
```diff
@@ -167,73 +167,149 @@ public InteractionOriginalResponseUpdater setTts(boolean tts) {
...
+    public InteractionOriginalResponseUpdater addAttachmentAsSpoiler(byte[] bytes, String fileName,
+                                                                     String description) {
+        delegate.addAttachment(bytes, "SPOILER_" + fileName, description);
         return this;
     }
 
     @Override
     public InteractionOriginalResponseUpdater addAttachmentAsSpoiler(InputStream stream, String fileName) {
-        delegate.addAttachment(stream, "SPOILER_" + fileName);
+        addAttachment(stream, "SPOILER_" + fileName);
```
KILLEliteMaste, line 305:
```suggestion
        addAttachment(stream, "SPOILER_" + fileName, null);
```

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -38,4 +42,39 @@ public CompletableFuture<Message> applyChanges() {
     public CompletableFuture<Message> replaceMessage() {
         return delegate.edit(message, true);
     }
+
+
```
KILLEliteMaste, line 46:
```suggestion
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -38,4 +42,39 @@ public CompletableFuture<Message> applyChanges() {
...
+    /**
+     * Removes an attachment from the message.
+     *
+     * @param attachment The attachment to remove.
+     */
+    public void removeExistingAttachment(Attachment attachment) {
+        delegate.removeExistingAttachment(attachment);
+    }
```
KILLEliteMaste, line 54: The same for the methods below
```suggestion
    public MessageUpdater removeExistingAttachment(Attachment attachment) {
        delegate.removeExistingAttachment(attachment);
        return this;
    }
```

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
```diff
@@ -46,7 +46,7 @@ public static WebhookMessageBuilder fromMessage(Message message) {
         }
         for (MessageAttachment attachment : message.getAttachments()) {
             // Since spoiler status is encoded in the file name, it is copied automatically.
-            builder.addAttachment(attachment.getUrl());
+            builder.addAttachment(attachment.getUrl(), attachment.getDescription().get());
```
KILLEliteMaste, line 49:
```suggestion
            builder.addAttachment(attachment.getUrl(), attachment.getDescription().orElse(null));
```

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
```diff
@@ -248,29 +261,67 @@ public WebhookMessageBuilder addAttachment(BufferedImage image, String fileName)
...
+     *
+     * @param file The file to add as an attachment.
+     * @param description The description of the file.
+     * @return The current instance in order to chain call methods.
+     */
+    public WebhookMessageBuilder addAttachment(File file, String description) {
+        delegate.addAttachment(file, description);
+        return this;
+    }
+
```
KILLEliteMaste, line 279: You marked this as resolved again without removing the line

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -594,18 +623,18 @@ private CompletableFuture<Message> checkForAttachmentsAndExecuteRequest(TextChan
                                                                             ObjectNode body,
                                                                             RestRequest<Message> request,
                                                                             boolean clearAttachmentsIfAppropriate) {
-        if (attachments.isEmpty() && embeds.stream().noneMatch(EmbedBuilder::requiresAttachments)) {
-            if (clearAttachmentsIfAppropriate) {
+        if (newAttachments.isEmpty() && embeds.stream().noneMatch(EmbedBuilder::requiresAttachments)) {
+            if (attachmentsToRemove.isEmpty() && clearAttachmentsIfAppropriate) {
```
KILLEliteMaste, line 627: Correct me if I am wrong, but shouldn't this check removed? It can only be true if the message is being edited. And if  `attachmentsToRemove` is empty we add all attachments to the array previously to keep them. With this method we would override the array and remove all attachments? I think this should be removed

KILLEliteMaste (changes requested):

Mysterypotatoguy (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -168,6 +170,40 @@ public CompletableFuture<BufferedImage> asImage() {
         return new FileContainer(getUrl()).asBufferedImage(getApi());
     }
 
+    /**
+     * Adds the json data to the given object node.
```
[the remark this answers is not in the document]
KILLEliteMaste: The return text also needs to be changed. Seems to be copied from the `EmbedBuilderDelegateImpl`

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -654,23 +699,28 @@ private void prepareAllowedMentions(ObjectNode body) {
...
             String mediaType = URLConnection
-                    .guessContentTypeFromName(attachments.get(i).getFileTypeOrName());
+                    .guessContentTypeFromName(fileContainer.getFileTypeOrName());
             if (mediaType == null) {
                 mediaType = "application/octet-stream";
             }
-            multipartBodyBuilder.addFormDataPart("file" + i, attachments.get(i).getFileTypeOrName(),
-                    RequestBody.create(MediaType.parse(mediaType), bytes));
+            multipartBodyBuilder.addFormDataPart("file" + i, fileContainer.getFileTypeOrName(),
+                    RequestBody.create(bytes, MediaType.parse(mediaType)));
```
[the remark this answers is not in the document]
KILLEliteMaste: > According to the [docs](https://discord.com/developers/docs/reference#editing-message-attachments-example-request-bodies-multipartformdata) this needs to be in the format file[n]

It actually has to be `files[n]` (missing `s` in your new example)

```
--boundary
Content-Disposition: form-data; name="files[0]"; filename="myfilename.png"
Content-Type: image/png

[image bytes]
--boundary
Content-Disposition: form-data; name="files[1]"; filename="mygif.gif"
Content-Type: image/gif

[image bytes]
--boundary--
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -38,4 +42,47 @@ public CompletableFuture<Message> applyChanges() {
     public CompletableFuture<Message> replaceMessage() {
         return delegate.edit(message, true);
     }
+
+    /**
+     * Removes an attachment from the message.
+     *
+     * @param attachment The attachment to remove.
+     * @return The updated message.
```
Mysterypotatoguy, line 50:
```suggestion
     * @return The current instance in order to chain call methods.
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -38,4 +42,47 @@ public CompletableFuture<Message> applyChanges() {
...
+     */
+    public MessageUpdater removeExistingAttachment(Attachment attachment) {
+        delegate.removeExistingAttachment(attachment);
+        return this;
+    }
+
+    /**
+     * Removes all the attachments from the message.
+     *
+     * @return The updated message.
```
Mysterypotatoguy, line 60:
```suggestion
     * @return The current instance in order to chain call methods.
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -38,4 +42,47 @@ public CompletableFuture<Message> applyChanges() {
...
+    public MessageUpdater removeExistingAttachments() {
+        delegate.removeExistingAttachments();
+        return this;
+    }
+
+    /**
+     * Removes multiple attachments from the message.
+     *
+     * @param attachments The attachments to remove.
+     * @return The updated message.
```
Mysterypotatoguy, line 71:
```suggestion
     * @return The current instance in order to chain call methods.
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageUpdater.java
```diff
@@ -38,4 +42,47 @@ public CompletableFuture<Message> applyChanges() {
...
+    public MessageUpdater removeExistingAttachments(Attachment... attachments) {
+        removeExistingAttachments(Arrays.asList(attachments));
+        return this;
+    }
+
+    /**
+     * Removes multiple attachments from the message.
+     *
+     * @param attachments The attachments to remove.
+     * @return The updated message.
```
Mysterypotatoguy, line 82:
```suggestion
     * @return The current instance in order to chain call methods.
```

Mysterypotatoguy (changes requested):

KILLEliteMaste (approved):
LGTM!

Mysterypotatoguy (approved):
LGTM if everything has been tested

KILLEliteMaste:
@all-contributors please add @RealYusufIsmail for code

Issue #1076, opened 2022-08-06
Add `app_permissions` field to interactions


KILLEliteMaste:
https://discord.com/developers/docs/change-log#calculated-permissions-in-interaction-payloads
https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object

RealYusufIsmail:
shall i add this in api 10 update

KILLEliteMaste:
Non, iIt is not required to upgrade to gateway v10, so it belongs to it's own PR

PR #1077, opened 2022-08-06
Add app_permissions field to Interactions class

AhmedA1559:
#1076.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
@@ -145,9 +145,10 @@ public interface InteractionBase extends DiscordEntity {
     Optional<DiscordLocale> getServerLocale();
 
     /**
-     * Gets the set of permissions of the app in context to the interaction.
+     * Gets the permissions the bot has within the channel the interaction was sent from.
+     * Not present if the interaction has been invoked in a direct message.
      *
-     * @return The app's permissions, in context of the interaction.
+     * @return The bots permissions within the channel the interaction was sent from.
      */
-    Optional<EnumSet<PermissionType>> getAppPermissions();
+    Optional<EnumSet<PermissionType>> getBotPermissions();
 }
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
```diff
@@ -142,4 +144,10 @@ default CompletableFuture<Void> respondWithModal(String customId, String title,
      */
     Optional<DiscordLocale> getServerLocale();
 
+    /**
+     * Gets the set of permissions of the app in context to the interaction.
```
KILLEliteMaste, line 148:
```suggestion
     * Gets the permissions the bot has within the channel the interaction was sent from.
     * Not present if the interaction has been invoked in a direct message.
```

javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
```diff
@@ -142,4 +144,10 @@ default CompletableFuture<Void> respondWithModal(String customId, String title,
...
+     *
+     * @return The app's permissions, in context of the interaction.
```
KILLEliteMaste, line 150:
```suggestion
     * @return The bots permissions within the channel the interaction was sent from.
```

javacord-api/src/main/java/org/javacord/api/interaction/InteractionBase.java
```diff
@@ -142,4 +144,10 @@ default CompletableFuture<Void> respondWithModal(String customId, String title,
...
+     */
+    Optional<EnumSet<PermissionType>> getAppPermissions();
```
KILLEliteMaste, line 152: I think this should be renamed as we do not have anything else than a bot like an only REST-Client
```suggestion
    Optional<EnumSet<PermissionType>> getBotPermissions();
```

KILLEliteMaste (changes requested):
Just the renaming and some Javadoc changes, otherwise it looks good to me.

KILLEliteMaste (approved):
Thank you for the PR. LGTM!

KILLEliteMaste:
@all-contributors please add @AhmedA1559 for code

Issue #1075, opened 2022-08-06
Add message interaction objects


KILLEliteMaste:
It seems like we do not currently support message interaction objects 
> This is sent on the [message object](https://discord.com/developers/docs/resources/channel#message-object) when the message is a response to an Interaction without an existing message.

https://discord.com/developers/docs/change-log#add-subcommand-groups-and-subcommands-to-message-interaction-objects
https://discord.com/developers/docs/interactions/receiving-and-responding#message-interaction-object

AhmedA1559:
Can I be assigned this?

KILLEliteMaste:
Sure :)

PR #1078, opened 2022-08-06
Add support for Message Interactions Object

AhmedA1559:
#1075.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -22,6 +22,7 @@ import org.javacord.api.entity.permission.Role;
 import org.javacord.api.entity.server.Server;
 import org.javacord.api.entity.sticker.StickerItem;
 import org.javacord.api.entity.user.User;
+import org.javacord.api.interaction.MessageInteraction;
 import org.javacord.api.listener.message.MessageAttachableListenerManager;
 import org.javacord.api.util.DiscordRegexPattern;
 
@@ -914,6 +915,13 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
      * @return All reactions of the message.
      */
     List<Reaction> getReactions();
+
+    /**
+     * Gets the Message Interaction Object if the message is a response to an Interaction without an existing message.
+     *
+     * @return The Message Interaction Object.
+     */
+    Optional<MessageInteraction> getMessageInteraction();
 
     /**
      * Gets all components of the message.
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
@@ -19,6 +19,7 @@ import org.javacord.api.entity.message.embed.Embed;
 import org.javacord.api.entity.permission.Role;
 import org.javacord.api.entity.sticker.StickerItem;
 import org.javacord.api.entity.user.User;
+import org.javacord.api.interaction.MessageInteraction;
 import org.javacord.api.util.DiscordRegexPattern;
 import org.javacord.core.DiscordApiImpl;
 import org.javacord.core.entity.emoji.UnicodeEmojiImpl;
@@ -28,6 +29,7 @@ import org.javacord.core.entity.server.ServerImpl;
 import org.javacord.core.entity.sticker.StickerItemImpl;
 import org.javacord.core.entity.user.MemberImpl;
 import org.javacord.core.entity.user.UserImpl;
+import org.javacord.core.interaction.MessageInteractionImpl;
 import org.javacord.core.listener.message.InternalMessageAttachableListenerManager;
 import org.javacord.core.util.cache.MessageCacheImpl;
 
@@ -73,6 +75,11 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
      * The components of the message.
      */
     private final List<HighLevelComponent> components = new ArrayList<>();
+
+    /**
+     * The Message Interaction Object of the message.
+     */
+    private final MessageInteraction messageInteraction;
 
     /**
      * The type of the message.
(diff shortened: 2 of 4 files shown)


javacord-core/src/main/java/org/javacord/core/interaction/MessageInteractionImpl.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param message  The message that the interaction originated from.
+     * @param jsonData The json data of the interaction context.
+     */
+    public MessageInteractionImpl(Message message, JsonNode jsonData) {
+        this.message = message;
+
+        id = jsonData.get("id").asLong();
+
+        int typeId = jsonData.get("type").asInt();
+        type = InteractionType.fromValue(typeId);
```
KILLEliteMaste, line 36: Please remove the `typeId` variable and inline the statement

javacord-core/src/main/java/org/javacord/core/interaction/MessageInteractionImpl.java
```diff
@@ -0,0 +1,82 @@
...
+
+        name = jsonData.get("name").asText();
+
+        if (jsonData.hasNonNull("member")) {
+            MemberImpl member = new MemberImpl(
+                    (DiscordApiImpl) message.getApi(),
+                    (ServerImpl) message.getServer().orElseThrow(AssertionError::new),
+                    jsonData.get("member"),
+                    null
```
KILLEliteMaste, line 45: Does the member packet include the user? I wouldn't say so because it is already present in a separate field `user`. Then you would need to create a user object instead of passing null

javacord-core/src/main/java/org/javacord/core/interaction/MessageInteractionImpl.java
```diff
@@ -0,0 +1,82 @@
...
+            user = (UserImpl) member.getUser();
+        } else {
+            user = new UserImpl((DiscordApiImpl) message.getApi(), jsonData.get("user"), (MemberImpl) null, null);
+        }
+    }
+
+    @Override
+    public DiscordApi getApi() {
+        return message.getApi();
+    }
```
KILLEliteMaste, line 56: You can override the method in the interface and use a default implementation
```java
@Override
default DiscordApi getApi() {
    return getMessage().getApi();
}
```

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/interaction/MessageInteractionImpl.java
AhmedA1559: > Does the member packet include the user? I wouldn't say so because it is already present in a separate field `user`. Then you would need to create a user object instead of passing null

The documentation lists the type of the member field as a "partial member object" so I think it does not then.

KILLEliteMaste: Don't you have the JSON packet from your testing so you could confirm if this is really the case? :)

AhmedA1559:
Weirdly enough, I can't seem to get Discord to return a member field in the json. I looked around and it also seems that [Discord.js](https://cs.github.com/discordjs/discord.js/blob/a7deb8f89830ead6185c5fb46a49688b6d209ed1/packages/discord.js/src/structures/Message.js#L343) omits the member object, opting only to include the user field only.

KILLEliteMaste:
> Weirdly enough, I can't seem to get Discord to return a member field in the json. I looked around and it also seems that [Discord.js](https://cs.github.com/discordjs/discord.js/blob/a7deb8f89830ead6185c5fb46a49688b6d209ed1/packages/discord.js/src/structures/Message.js#L343) omits the member object, opting only to include the user field only.

Have you tried enabling the `GUILD_MEMBERS` intent?

AhmedA1559:
> > Weirdly enough, I can't seem to get Discord to return a member field in the json. I looked around and it also seems that [Discord.js](https://cs.github.com/discordjs/discord.js/blob/a7deb8f89830ead6185c5fb46a49688b6d209ed1/packages/discord.js/src/structures/Message.js#L343) omits the member object, opting only to include the user field only.
> 
> Have you tried enabling the `GUILD_MEMBERS` intent?

Yeah: `DiscordApi api = new DiscordApiBuilder().setToken(token).setAllIntents().login().join();`

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -907,6 +908,13 @@ default Optional<CompletableFuture<Message>> requestReferencedMessage() {
      */
     List<Reaction> getReactions();
 
+    /**
+     * Gets the Message Interaction Object if the message responds to an interaction.
```
KILLEliteMaste, line 912:
```suggestion
     * Gets the Message Interaction Object if the message is a response to an Interaction without an existing message.
```

KILLEliteMaste (changes requested):

KILLEliteMaste:
Not sure what you did, but in my setup with also having the guild members intent enabled I get a member in the message interaction when invoking a UserContextInteraction. But the logic looks right to me for handling the member JSON

KILLEliteMaste (approved):
LGTM!
Can you please rebase to the latest version, resolve the merge conflicts and squash the PR into a single commit?

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -365,6 +372,12 @@ public void setUpdatableFields(JsonNode data) {
             });
         }
 
+        if (data.hasNonNull("interaction")) {
+            messageInteraction = new MessageInteractionImpl(this, data.get("interaction"));
+        } else {
+            messageInteraction = null;
+        }
```
KILLEliteMaste, line 379: Move to the the constructor as I don't think this will ever change

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):

Issue #1084, opened 2022-08-07
Add missing field `position` to the message object


KILLEliteMaste:
https://discord.com/developers/docs/resources/channel#message-object

PR #1088, opened 2022-08-07
Add missing field position to the message object

shubhwip:
Issue Link : https://github.com/Javacord/Javacord/issues/1084
Issue Description : https://discord.com/developers/docs/resources/channel#message-object

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -959,6 +959,16 @@ public interface Message extends DiscordEntity, Comparable<Message>, UpdatableFr
     Set<StickerItem> getStickerItems();
 
     /**
+     * A generally increasing integer (there may be gaps or duplicates)
+     * that represents the approximate position of the message in a thread,
+     * it can be used to estimate the relative position of the message in a thread
+     * in company with {@link ServerThreadChannel#getTotalNumberOfMessagesSent()} on parent thread.
+     *
+     * @return The approximate position of the message in a thread.
+     */
+    Optional<Integer> getPosition();
+
+    /**
      * Gets all channels mentioned in this message.
      *
      * @return All channels mentioned in this message.
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
@@ -170,6 +170,11 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
      * The sticker items in this message.
      */
     private final Set<StickerItem> stickerItems = new HashSet<>();
+        
+    /**
+     * The approximate position of the message in a thread.
+     */
+    private final Integer position;
 
     /**
      * Creates a new message object.
@@ -208,6 +213,8 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
         messageInteraction = data.hasNonNull("interaction")
             ? new MessageInteractionImpl(this, data.get("interaction"))
             : null;
+        
+        position = data.hasNonNull("position") ? data.get("position").asInt() : null;
 
         setUpdatableFields(data);
 
@@ -240,6 +247,7 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
      * @param mentions          The users mentioned in this message.
      * @param roleMentions      The roles mentioned in this message.
      * @param stickerItems      The sticker items in this message.
+     * @param position          The position in this message
      */
     private MessageImpl(DiscordApiImpl api, TextChannel channel, long id, MessageType type,
                         MessageAuthor author, MessageActivityImpl activity, String content,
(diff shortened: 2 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -942,6 +942,13 @@ default Optional<CompletableFuture<Message>> requestReferencedMessage() {
      */
     Set<StickerItem> getStickerItems();
 
+    /**
+     * Gets the approximate position of the message in a thread
```
RealYusufIsmail, line 946:
```suggestion
     * Gets the approximate position of the message in a thread.
```

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -157,6 +157,14 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
      */
     private final Set<StickerItem> stickerItems = new HashSet<>();
 
+    /**
+     * A generally increasing integer (there may be gaps or duplicates)
+     * that represents the approximate position of the message in a thread,
+     * it can be used to estimate the relative position of the messsage in a thread
+     * in company with total_message_sent on parent thread
```
RealYusufIsmail, line 164: No need to get into this much detail in an impl.

Just says the approximate position of the message in a thread

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -157,6 +157,14 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
...
+     */
```
RealYusufIsmail, line 165:
```suggestion
    /**
     * The position of the message in a channel.
     */
```

RealYusufIsmail (changes requested):

RealYusufIsmail (approved):

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
KILLEliteMaste: I would say you should include the Javadoc you had in the implementation as this is not very informative. Someone might think why is it approximate or in general what you would do with this (its used in companion with total_message_sent). Including the Javadoc should clear it up a bit

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -942,6 +942,16 @@ default Optional<CompletableFuture<Message>> requestReferencedMessage() {
      */
     Set<StickerItem> getStickerItems();
 
+    /**
+     * A generally increasing integer (there may be gaps or duplicates)
+     * Represents the approximate position of the message in a thread
+     * It can be used to estimate the relative position of the messsage
+     * in a thread in company with total_message_sent on parent thread.
```
KILLEliteMaste, line 949: Even when placing dots after the sentences it reads very choppy so lets just change it to what's written in the docs
```suggestion
     * A generally increasing integer (there may be gaps or duplicates)
     * that represents the approximate position of the message in a thread,
     * it can be used to estimate the relative position of the message in a thread
     * in company with {@link ServerThreadChannel#getTotalNumberOfMessagesSent()} on parent thread.
```
This will probably not work right away because the mentioned method in the Javadoc does no yet exists. So we have to wait for #1089 to be merged first.

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
shubhwip: Thanks for the putting suggested change :) !! Added the change but for now build will fail until #1089 is merged.

RealYusufIsmail: will try to get as fast as possible

KILLEliteMaste:
@shubhwip the PR has been merged. Can you please rebase the branch with the latest version?

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -208,6 +213,10 @@ public MessageImpl(DiscordApiImpl api, TextChannel channel, JsonNode data) {
         messageInteraction = data.hasNonNull("interaction")
             ? new MessageInteractionImpl(this, data.get("interaction"))
             : null;
+        
+        position = data.hasNonNull("position") 
+            ? data.get("position").asInt() 
+            : null;
```
RealYusufIsmail, line 219:
```suggestion
        
        position = data.hasNonNull("position") ? data.get("position").asInt() 
            : null;
```

RealYusufIsmail: Unsure if this will cause the build to fail

RealYusufIsmail (approved):

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
shubhwip: no it was due to missing constructor param.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
```diff
@@ -170,6 +170,11 @@ public class MessageImpl implements Message, InternalMessageAttachableListenerMa
      * The sticker items in this message.
      */
     private final Set<StickerItem> stickerItems = new HashSet<>();
+        
+    /**
+     * The approximate position of the message in a thread.
+     */
+    private final int position;
```
KILLEliteMaste, line 177: Assigning `null` to this will not work. Change it to `Integer`

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):
Thank you, LGTM!

PR #1090, opened 2022-08-07
Added  missing field fail_if_not_exists to the Message Reference Object

### LINKED ISSUE #1085: Add missing field `fail_if_not_exists` to the Message Reference Object
https://discord.com/developers/docs/resources/channel#message-reference-object

RealYusufIsmail:
closes #1085


javacord-core/src/main/java/org/javacord/core/entity/message/MessageReferenceImpl.java
```diff
@@ -58,6 +63,7 @@ public MessageReferenceImpl(DiscordApiImpl api, Message message, JsonNode data)
         }
 
         this.message = message;
+        this.failIfNotExists = data.get("fail_if_not_exists").asBoolean();
```
shubhwip, line 66: Can we specify here default value as `true` as mentioned in the docs ?

javacord-core/src/main/java/org/javacord/core/entity/message/MessageReferenceImpl.java
```diff
@@ -84,4 +90,9 @@ public Optional<Long> getMessageId() {
     public Optional<Message> getMessage() {
         return Optional.ofNullable(message);
```
shubhwip, line 91: Would there be a need of `Optional` if we have default value as `true` always :) ?

wasdennnoch:
`fail_if_not_exists` is used when _sending_ messages, not when receiving them, so it has to be added to the corresponding message builders etc. It determines if, when sending a reply, discord should return you an error when you're replying to an unknown message, or if Discord should ignore the message reference in that case and treat it as a normal message send without a reply.

RealYusufIsmail:
> `fail_if_not_exists` is used when _sending_ messages, not when receiving them, so it has to be added to the corresponding message builders etc. It determines if, when sending a reply, discord should return you an error when you're replying to an unknown message, or if Discord should ignore the message reference in that case and treat it as a normal message send without a reply.

will do so

RealYusufIsmail:
> #1085

When replying is there a way to check if the message id is null like here,

```java
 if (replyingTo != null) {
            body.putObject("message_reference").put("message_id", replyingTo);
        }
```

RealYusufIsmail:
or would i check here for example ```java
private CompletableFuture<Message> executeRequestWithoutAttachments(TextChannel channel,
                                                                        ObjectNode body,
                                                                        RestRequest<Message> request) {
        request.setBody(body);
        //check if fail_if_not_exists is false and message doesn't exist, trough exception
        return request.execute(result -> ((DiscordApiImpl) channel.getApi())
                .getOrCreateMessage(channel, result.getJsonBody()))
                .whenComplete((newMessage, throwable) -> {
                    if (throwable != null) {
                        throwable.printStackTrace();
                    }
                });
    }

```

**Comment by wasdennnoch** [comment]
> When replying is there a way to check if the message id is null like here,

Not sure where/why you're trying to check that. The code you posted already checks if the message should be a reply, and populates allowed_mentions. Now you just have to add a variable in the replied_mentions payload that tells _discord_ if it should check if the message exists or not. So something like

```java
if (replyingTo != null) {
    body.putObject("message_reference")
        .put("message_id", replyingTo)
        .put("fail_if_not_exists", failIfNotExists);
}
```

where the boolean `failIfNotExists` (or whatever you'll call the variable) will be configurable by the user somehow.

RealYusufIsmail:
ok

RealYusufIsmail:
> > When replying is there a way to check if the message id is null like here,
> 
> Not sure where/why you're trying to check that. The code you posted already checks if the message should be a reply, and populates allowed_mentions. Now you just have to add a variable in the replied_mentions payload that tells _discord_ if it should check if the message exists or not. So something like
> 
> ```java
> if (replyingTo != null) {
>     body.putObject("message_reference")
>         .put("message_id", replyingTo)
>         .put("fail_if_not_exists", failIfNotExists);
> }
> ```
> 
> where the boolean `failIfNotExists` (or whatever you'll call the variable) will be configurable by the user somehow.

done

javacord-core/src/main/java/org/javacord/core/entity/message/MessageReferenceImpl.java
```diff
@@ -84,4 +91,9 @@ public Optional<Long> getMessageId() {
...
     }
+
+    @Override
+    public Optional<Boolean> doesMessageReferenceExist() {
+        return Optional.ofNullable(failIfNotExists);
```
shubhwip, line 97: Do we need `Optional` here as value of `failIfNotExists ` will be true by default always.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageReferenceImpl.java
```diff
@@ -58,6 +63,8 @@ public MessageReferenceImpl(DiscordApiImpl api, Message message, JsonNode data)
         }
 
         this.message = message;
+        this.failIfNotExists = data.hasNonNull("fail_if_not_exists") ? data.get("fail_if_not_exists").asBoolean() :
```
shubhwip, line 66: This could have been simplified with `this.failIfNotExists = data.get("fail_if_not_exists").asBoolean(true);`

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -1517,20 +1517,23 @@ default boolean canDelete(User user) {
      * Replies to this message with the given text.
      *
      * @param messageContent The text to reply with.
-     * @return The sent message.
+     * @param checkIfMessageExists Used to tell discord if you want to check if the message exists.
+     * @return The message that was sent.
      */
-    default CompletableFuture<Message> reply(String messageContent) {
-        return new MessageBuilder().replyTo(getId()).setContent(messageContent).send(getChannel());
+    default CompletableFuture<Message> reply(String messageContent, boolean checkIfMessageExists) {
```
AhmedA1559, line 1523: Wouldn't this break pre-existing code? I think you should make a method with signature `CompletableFuture<Message> reply(String messageContent)` that calls this method with a default parameter for `checkIfMessageExists`.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -116,21 +116,23 @@ public MessageBuilder addStickers(Collection<Sticker> stickers) {
      * Sets the message to reply to.
      *
      * @param message The message to reply to.
+     * @param checkIfMessageExists Used to tell discord if you want to check if the message exists.
      * @return The current instance in order to chain call methods.
      */
-    public MessageBuilder replyTo(Message message) {
-        delegate.replyTo(message.getId());
+    public MessageBuilder replyTo(Message message, boolean checkIfMessageExists) {
```
AhmedA1559, line 122: The same here.

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
```diff
@@ -121,6 +121,11 @@ public class MessageBuilderBaseDelegateImpl implements MessageBuilderBaseDelegat
      */
     protected Long replyingTo = null;
 
+    /**
+     * Used to tell discord to check if the reference message exists.
+     */
+    private boolean checkIfMessageExists = false;
```
AhmedA1559, line 127: [The discord docs](https://discord.com/developers/docs/resources/channel#message-reference-object-message-reference-structure) report that this is, by default, true.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageReference.java
```diff
@@ -75,4 +75,12 @@ default Optional<CompletableFuture<Message>> requestMessage() {
 
         return getMessageId().flatMap(messageId -> getChannel().map(channel -> channel.getMessageById(messageId)));
     }
+
+    /**
+     * Gets weather the message reference is valid.
```
AhmedA1559, line 80:
```suggestion
     * Gets whether the message reference is valid.
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageReference.java
```diff
@@ -75,4 +75,12 @@ default Optional<CompletableFuture<Message>> requestMessage() {
...
+     *
+     * @return {@code true} if the message reference is valid, {@code false} otherwise.
+     */
+    //
```
AhmedA1559, line 84:
```suggestion
```

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -1517,20 +1517,23 @@ default boolean canDelete(User user) {
...
+        return new MessageBuilder().replyTo(getId(), checkIfMessageExists).setContent(messageContent)
+                .send(getChannel());
     }
 
     /**
      * Replies to this message with the given embed.
      *
      * @param embed The EmbedBuilder to reply with.
-     * @return The sent message.
+     * @param checkIfMessageExists Used to tell discord if you want to check if the message exists.
```
AhmedA1559, line 1532:
```suggestion
     * @param checkIfMessageExists If true, tells discord to throw an error if the message reference does not exist, if false, then leaves out the message reference if it does not exist but still sends the message.
```

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -1517,20 +1517,23 @@ default boolean canDelete(User user) {
      * Replies to this message with the given text.
      *
      * @param messageContent The text to reply with.
-     * @return The sent message.
+     * @param checkIfMessageExists Used to tell discord if you want to check if the message exists.
```
AhmedA1559, line 1520:
```suggestion
     * @param checkIfMessageExists If true, tells discord to throw an error if the message reference does not exist, if false, then leaves out the message reference if it does not exist but still sends the message.
```

AhmedA1559 (changes requested):

javacord-core/src/main/java/org/javacord/core/entity/message/MessageReferenceImpl.java
RealYusufIsmail: yeah, will change that

javacord-core/src/main/java/org/javacord/core/entity/message/MessageBuilderBaseDelegateImpl.java
RealYusufIsmail: Would just doing
` private boolean checkIfMessageExists;`
work better

RealYusufIsmail:
@AhmedA1559 @wasdennnoch can you do re review please

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -116,21 +116,23 @@ public MessageBuilder addStickers(Collection<Sticker> stickers) {
...
     /**
      * Sets the message to reply to.
      *
      * @param messageId The id of the message to reply to.
+     * @param checkIfMessageExists Used to tell discord if you want to check if the message exists.
      * @return The current instance in order to chain call methods.
      */
-    public MessageBuilder replyTo(long messageId) {
-        delegate.replyTo(messageId);
+    public MessageBuilder replyTo(long messageId, boolean checkIfMessageExists) {
```
AhmedA1559, line 134: Needs to be fixed to include `MessageBuilder replyTo(long messageId)` and `MessageBuilder replyTo(Message message)` signatures.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -1517,20 +1517,47 @@ default boolean canDelete(User user) {
...
+    /**
+     * Replies to this message with the given embed.
+     *
+     * @param embed The EmbedBuilder to reply with.
+     * @param checkIfMessageExists If true, tells discord to throw an error if the message reference does not exist,
+     *                             if false, then leaves out the message reference,
+     *                             if it does not exist but still sends the message.
+     * @return The message that was sent.
+     */
+    default CompletableFuture<Message> reply(EmbedBuilder embed, boolean checkIfMessageExists) {
```
AhmedA1559, line 1559: I think the name `checkIfMessageExists` is a misnomer: setting this to false still has the behavior of checking the message reference, but instead does not assert an error. It would probably be better to have a name like `assertReferenceExists` instead. This should be applied to the rest of the PR.

AhmedA1559 (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
RealYusufIsmail: will do

RealYusufIsmail:
@AhmedA1559 re-review please

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -119,7 +119,7 @@ public MessageBuilder addStickers(Collection<Sticker> stickers) {
      * @return The current instance in order to chain call methods.
      */
     public MessageBuilder replyTo(Message message) {
-        delegate.replyTo(message.getId());
+        replyTo(message, false);
```
AhmedA1559, line 122: The default value is supposed to be true.

AhmedA1559 (approved):
Other than that, LGTM. Please make sure to test though to check behavior.

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
RealYusufIsmail: ok

RealYusufIsmail:
done

RealYusufIsmail:
@KILLEliteMaste can you have a look at this pr please

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -119,7 +119,7 @@ public MessageBuilder addStickers(Collection<Sticker> stickers) {
      * @return The current instance in order to chain call methods.
      */
     public MessageBuilder replyTo(Message message) {
-        delegate.replyTo(message.getId());
+        replyTo(message, true);
         return this;
```
KILLEliteMaste, line 123:
```suggestion
        return replyTo(message, true);
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -130,7 +130,31 @@ public MessageBuilder replyTo(Message message) {
      * @return The current instance in order to chain call methods.
      */
     public MessageBuilder replyTo(long messageId) {
-        delegate.replyTo(messageId);
+        replyTo(messageId, true);
+        return this;
```
KILLEliteMaste, line 134:
```suggestion
        return replyTo(messageId, true);
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageBuilder.java
```diff
@@ -130,7 +130,31 @@ public MessageBuilder replyTo(Message message) {
...
+    /**
+     * Sets the message to reply to.
+     *
+     * @param message The message to reply to.
+     * @param assertReferenceExists Used to tell discord if you want to check if the message exists.
+     * @return The current instance in order to chain call methods.
+     */
+    public MessageBuilder replyTo(Message message, boolean assertReferenceExists) {
+        delegate.replyTo(message.getId(), assertReferenceExists);
+        return this;
```
KILLEliteMaste, line 146:
```suggestion
        return replyTo(message.getId(), assertReferenceExists);
```

javacord-api/src/main/java/org/javacord/api/entity/message/MessageReference.java
```diff
@@ -75,4 +75,11 @@ default Optional<CompletableFuture<Message>> requestMessage() {
 
         return getMessageId().flatMap(messageId -> getChannel().map(channel -> channel.getMessageById(messageId)));
     }
+
+    /**
+     * Gets whether the message reference is valid.
+     *
+     * @return {@code true} if the message reference is valid, {@code false} otherwise.
+     */
+    boolean doesMessageReferenceExist();
```
KILLEliteMaste, line 84: As @wasdennnoch pointed out, this field is only used when sending the message.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -1544,20 +1544,47 @@ default boolean canDelete(User user) {
...
+        return reply(messageContent, true);
+    }
+
+    /**
+     * Replies to this message with the given text.
+     *
+     * @param messageContent The text to reply with.
+     * @param assertReferenceExists If true, tells discord to throw an error if the message reference does not exist,
+     *                             if false, then leaves out the message reference,
+     *                             if it does not exist but still sends the message.
```
KILLEliteMaste, line 1559:
```suggestion
     * @param assertReferenceExists If true, throw an error if the message you are replying to does not exists otherwise, 
     *                                                   if false send the message regardless without a reference to a message.
```
The same for the other Javadoc

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):
Thank you, LGTM!

PR #1091, opened 2022-08-07
Support allowed mentions `replied_user`

wasdennnoch:
https://discord.com/developers/docs/resources/channel#allowed-mentions-object-allowed-mentions-structure

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/mention/internal/AllowedMentionsBuilderDelegate.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/mention/internal/AllowedMentionsBuilderDelegate.java
@@ -27,6 +27,13 @@ public interface AllowedMentionsBuilderDelegate {
     void setMentionUsers(boolean value);
 
     /**
+     * Sets whether the replied user should be mentioned if the message is a reply.
+     *
+     * @param value Whether the message mentions the replied user.
+     */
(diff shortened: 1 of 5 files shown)


javacord-api/src/main/java/org/javacord/api/entity/message/mention/internal/AllowedMentionsBuilderDelegate.java
```diff
@@ -26,6 +26,13 @@ public interface AllowedMentionsBuilderDelegate {
      */
     void setMentionUsers(boolean value);
 
+    /**
+     * Sets the general mentions of the message.
```
KILLEliteMaste, line 30: What is `general mentions`? Shouldn't it just be something like `Sets whether the author should be mentioned if the message is a reply` ?

wasdennnoch: I just copied that from the above 3 methods 😅 
I think it refers to [this](https://github.com/BeemoBot/Javacord/blob/adrian/allowed-mentions-replied-user/javacord-core/src/main/java/org/javacord/core/entity/message/mention/AllowedMentionsBuilderDelegateImpl.java#L19), but I agree it's not very descriptive so I'll change it.

KILLEliteMaste (approved):
LGTM!

PR #1093, opened 2022-08-07
Reinstate javadoc options lost in rework

Mysterypotatoguy:
Re-adds some of the missing javadoc build options, in particular the `-no-module-directories` option which fixes the `/unknown/` redirect

I'm not too familiar with Gradle/Kotlin DSL so any input on code styling welcome.
I have missed out the section which split the docs into "Public API" and "Internal helpers" as I'm not sure if we wanted to keep that
And lastly there are a few function calls in the previous javadoc task which I'm not sure of their use and I cannot find documentation for, in particular `delegate.use()`

As a side note our compatibility in the `build.gradle.kts` still seems to be at 1.8, should this be upgraded?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -152,16 +152,32 @@ publishing {
 }
 
 subprojects {
-    tasks.register("configureJavadoc") {
-        doLast {
-            tasks.javadoc {
-                options.windowTitle = "Javacord $version (${project.ext.get("shortName")})"
-                title = options.windowTitle
+    tasks.withType<Javadoc>().configureEach {
+        options {
+            this as StandardJavadocDocletOptions
+            locale = "en"
+            encoding = "UTF-8"
+            docTitle = "Javacord ${project.version} (${project.property("shortName")})"
+            windowTitle = "$docTitle Documentation"
+            links("https://docs.oracle.com/javase/8/docs/api/")
+            isUse = true
+            isVersion = true
+            isAuthor = true
+            isSplitIndex = true
+
+            val toolchain = javadocTool
+                .map { JavaVersion.toVersion(it.metadata.languageVersion) }
+                .orElse(provider { JavaVersion.current() })
+                .get()
+            if (toolchain.isJava9Compatible) {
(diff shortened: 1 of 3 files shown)


Vampire:
Let's swap roles.
I pushed the proper changes.
The new build is full of bad practices, some added by me, some added by the rework.
I at least wanted to reduce it some and it was faster to just do it right than to tell you in detail what to change. :-)

Your changes also missed other things, like for example the proper linking from core docs to api docs.

> I have missed out the section which split the docs into "Public API" and "Internal helpers" as I'm not sure if we wanted to keep that

I added also this and strongly vote to keep it.
I find it ultimately useful, being able to just look at the official public api and have the internal helpers filtered out.
Anyone not wanting to use these tabs does not have to use them.
But it also makes it a step clearer which part of the api is under the shifted-semver guarantees and which are not.

> And lastly there are a few function calls in the previous javadoc task which I'm not sure of their use and I cannot find documentation for, in particular delegate.use()

The `delegate.` was to overcome a Groovy specific, where `use` is a method provided by Groovy that would be used instead of the method on the standard javadoc doclet options.
The methods are documented in the JavaDoc of the standard javadoc doclet options and also in the help of the Javadoc tool, as they just enable the same-named options for the JavaDoc tool.
`use` for example is an option that imho each and every JavaDoc should enable. I'm not aware of any reason not to enable it and always wondered why it is not on by default and I miss it regularly in any JavaDoc that does not have it enabled.
It generates the `Use` section where you can look up where types are used in the api, so for example which methods take class X as arguments, which return it as result, ....

> As a side note our compatibility in the build.gradle.kts still seems to be at 1.8, should this be upgraded?

No, why should it?
1.8 is still what we support.
It was just pre-announced that somewhen next year we will change that.
But who knows when next version will be released.
I don't think this should be changed already.

Mysterypotatoguy:
I can't approve my own PR but LGTM, tested locally

Vampire:
Yeah, no problem, technically, I'll do it.

Vampire (approved):

PR #1101, opened 2022-08-09
Fix typo in BotInviteScope.java

eltociear:
paramater -> parameter


KILLEliteMaste (approved):
Thank you for the fix :)

PR #1102, opened 2022-08-09
Added Session-specific Gateway Resume URLs

### LINKED ISSUE #1074: Session-specific Gateway Resume URLs
https://discord.com/developers/docs/change-log#sessionspecific-gateway-resume-urls

RealYusufIsmail:
closes : #1074 

## TODO

- [x] invalidate resume url & session on fail
- [x] cache resume url and session if session id is null
- [x] If resume url is null then use normal url
- [x] Resume with resume url if it is not null


RealYusufIsmail:
blocked until #1073

KILLEliteMaste:
Docs got a bit [clearer ](https://github.com/discord/discord-api-docs/commit/4ca01c593d7bc7c3f329cb3882fc03d495bc35fa)and we should not use the resume url when the session is invalid

RealYusufIsmail:
> Docs got a bit [clearer ](https://github.com/discord/discord-api-docs/commit/4ca01c593d7bc7c3f329cb3882fc03d495bc35fa)and we should not use the resume url when the session is invalid

will check it out

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -372,7 +373,15 @@ public void disconnect() {
     private void connect() {
         try {
             WebSocketFactory factory = new WebSocketFactory();
-            String webSocketUri = getGateway(api) + "?encoding=json&v=" + Javacord.DISCORD_GATEWAY_VERSION;
+            String webSocketUri;
+
+            if (sessionId == null) {
```
KILLEliteMaste, line 378: Why do you check if the session id is null? And you have duplicated code when the resumeUrl is null. This can be simplified by only checking whether resumeUrl is null

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -459,6 +468,7 @@ private void connect() {
             }
             websocket.connect();
         } catch (Throwable t) {
+            resumeUrl = null;
```
KILLEliteMaste, line 471: Shouldn't the session id be set to null too in this case?

KILLEliteMaste (changes requested):
I got this from the Discord Devs server how to handle the resumeUrl properly maybe this helps you as some things are missing I think:
```
Cache resume url just like you do session id
- do I have session id?
 - no, ready, cache resume url and session
 - yes
  - do i have resume url?
  - no, resume to default
  - yes, resume to url

- did my resume fail?
 - no, all is well
 - yes, invalidate resume url & session
```

I am wondering, don't we have to set the session id and now resumeUrl to null in case we receive an `INVALID_SESSION`? Unfortunately I have no experience with the Discord Websocket so far so i can't really tell.

RealYusufIsmail:
Will check it out

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -141,6 +141,7 @@ public class DiscordWebSocketAdapter extends WebSocketAdapter {
 
     private volatile int lastSeq = -1;
     private volatile String sessionId = null;
+    private String resumeUrl = null;
```
KILLEliteMaste, line 144:
```suggestion
    private volatile String resumeUrl = null;
```

KILLEliteMaste (changes requested):

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -645,8 +652,13 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
...
                 }
-                api.getGatewayIdentifyRatelimiter().requestQuota();
-                sendIdentify(websocket);
+                if (packet.get("d").asBoolean()) {
+                    api.getGatewayIdentifyRatelimiter().requestQuota();
+                    sendIdentify(websocket);
+                } else {
+                    sessionId = null;
+                    resumeUrl = null;
+                }
```
bassner, line 661: I don't get these changes. Any explanations on why you added it?
This looks to me as if the flag is false, you're not doing anything and just let the bot die.

RealYusufIsmail: The Boolean d identifies if the api can be resume or not. JDA does something similar

bassner: I know. But how would that be relevant here? We are not resuming, we are sending an identify(``sendIdentify``). So why would it matter if it is resumable?

Mysterypotatoguy: In the case that `d` is false the session is *not* resumable and from this point I presume we wait for the disconnect and start a full reconnect

bassner: As I said, we are not resuming. We are identifying...

Mysterypotatoguy: Oh, in that case yes that is correct https://discord.com/developers/docs/topics/gateway#resuming

bassner: Before this code block, we wait the 1-5 random seconds and then send an identify. I do not get why we would base that on the flag. This worked perfectly fine so far and I see no justification to change this in a PR that just aims to dynamically change the url used for resuming

Mysterypotatoguy: I agree with doing this before the 1-5s wait but this is currently nonconformant to the gateway docs. Leaving it like this will just cause longer reconnect delays if we try to resume an un-resumable invalid session

bassner: How exactly is this nonconformant?

```
1. Connection lost
2. Reconnect to the gateway
3. Send Opcode 6 Gateway Resume packet with session id
4. Gateway replies with opcode 9 invalid session (e.g. because the network loss was just too long)
5. We wait 1-5 seconds
6. We send an identify
```
We're doing steps 4 to 6 right here. And I don't get how the "d" flag would have any influence on wether or not we should send a fresh identify as documented in the link you sent above

bassner: Given [this PR](https://github.com/discord/discord-api-docs/pull/5323) (which is apparently not yet included in the docs website), the actual strategy should be to set the resume url to null and close the connection using ``sendCloseFrame``, and then let Javacord automatically create a new connection. Right?

Mysterypotatoguy: > the gateway has invalidated an active session and is requesting client action

We may also be able to receive INVALID_SESSION *before* a disconnect

bassner: Hmh. So, given that we should do the following:
```
Received Opcode 9
- Clear resume URL and session ID
- Did we just try to resume?
  Yes: Wait 1-5 seconds
  No: Continue without delay
- send close frame
```

RealYusufIsmail: how do i check if we should resume

RealYusufIsmail: would i use the boolean d

RealYusufIsmail:
I think I am done though I am unsure if I have done this right. @KILLEliteMaste any ideas

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -633,20 +641,28 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
...
+                        } catch (InterruptedException e) {
+                            logger.error("Interrupted while delaying reconnect!");
+                            return;
+                        }
                     }
+                    api.getGatewayIdentifyRatelimiter().requestQuota();
+                    sendIdentify(websocket);
+                    sendCloseFrame(websocket,
+                            WebSocketCloseReason.COMMANDED_RECONNECT.getNumericCloseCode(),
+                            WebSocketCloseReason.COMMANDED_RECONNECT.getCloseReason());
```
bassner, line 664: I don't think COMMANDED_REDIRECT is there correct code to use here. Discord has not commanded a reconnect, we just have an invalid session. We could just use the normal DISCONNECT close code here I guess ... Discord doesn't care afaik but using COMMANDED_REDIRECT is definitely incorrect

bassner (changes requested):

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -151,7 +152,8 @@ public class DiscordWebSocketAdapter extends WebSocketAdapter {
             new AtomicMarkableReference<>(null, false);
     private final AtomicReference<WebSocketFrame> nextHeartbeatFrame = new AtomicReference<>(null);
     private final List<WebSocketListener> identifyFrameListeners = Collections.synchronizedList(new ArrayList<>());
-
+    private final AtomicMarkableReference<WebSocketFrame> triedToResume =
+            new AtomicMarkableReference<>(null, false);
```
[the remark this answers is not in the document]
RealYusufIsmail: Yeah sounds reasonable, will implement it tomorrow or the latest the weekend

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -633,20 +641,28 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
...
+                                () -> oneToFiveSeconds / 100 % 10);
+                        try {
+                            Thread.sleep(oneToFiveSeconds);
+                        } catch (InterruptedException e) {
+                            logger.error("Interrupted while delaying reconnect!");
+                            return;
+                        }
                     }
+                    api.getGatewayIdentifyRatelimiter().requestQuota();
+                    sendIdentify(websocket);
```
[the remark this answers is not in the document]
RealYusufIsmail: can you check if i have done this right in my lates commit

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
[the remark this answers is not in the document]
RealYusufIsmail: can you check if i have done this right in my lates commit

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -656,12 +656,14 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
                             logger.error("Interrupted while delaying reconnect!");
                             return;
                         }
+                    } else {
+                        sendIdentify(websocket);
+
                     }
                     api.getGatewayIdentifyRatelimiter().requestQuota();
```
bassner, line 663: Huh? We don't want to send an identify here. You're still disconnecting right after this, right? Please remove the sendIdentify and the requestQuota

bassner (changes requested):

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
RealYusufIsmail: ok

RealYusufIsmail: why was sendIdentify used in the past.

bassner: because we didn't reconnect and used the existing connection instead
And now we do reconnect

RealYusufIsmail: ok, so does everything look good now, like as in done

bassner (changes requested):

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -630,23 +639,28 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
...
-                    int oneToFiveSeconds = 1000 + (int) (Math.random() * 4000);
-                    logger.info("Could not resume session. Reconnecting in {}.{} seconds...",
-                            () -> oneToFiveSeconds / 1000,
-                            () -> oneToFiveSeconds / 100 % 10);
-                    try {
-                        Thread.sleep(oneToFiveSeconds);
-                    } catch (InterruptedException e) {
-                        logger.error("Interrupted while delaying reconnect!");
-                        return;
+                    if (triedToResume) {
```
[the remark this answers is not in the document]
RealYusufIsmail: what do you mean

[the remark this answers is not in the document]
RealYusufIsmail: ah i see

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -630,9 +646,11 @@ public void onTextMessage(WebSocket websocket, String text) throws Exception {
                                WebSocketCloseReason.COMMANDED_RECONNECT.getCloseReason());
                 break;
             case INVALID_SESSION:
+                sessionId = null;
+                resumeUrl = null;
                 if (lastSentFrameWasIdentify.isMarked()) {
-                    logger.info("Hit identifying rate limit. Retrying in 5 seconds...");
-                } else {
+                    logger.info("Hit identifying rate limit. Reconnecting ...");
```
KILLEliteMaste, line 652:
```suggestion
                    logger.info("Hit identifying rate limit. Reconnecting...");
```

KILLEliteMaste (changes requested):
Other than that it LGTM!

KILLEliteMaste (approved):

bassner:
I'll test this branch in a staging environment to see if it's stable

bassner:
This seems to work properly on my system 👍 
grafik

bassner (approved):

javacord-core/src/main/java/org/javacord/core/util/gateway/WebSocketCloseReason.java
```diff
@@ -7,6 +7,8 @@ public enum WebSocketCloseReason {
 
     DISCONNECT(WebSocketCloseCode.NORMAL),
     HEARTBEAT_NOT_PROPERLY_ANSWERED(WebSocketCloseCode.UNKNOWN_ERROR, "Heartbeat was not answered properly"),
+    INVALID_SESSION_RECONNECT(WebSocketCloseCode.INVALID_SESSION_RECONNECT, "Received an INVALID_SESSION - " +
+            "reconnecting."),
```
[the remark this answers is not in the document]
RealYusufIsmail: yeah thats better

RealYusufIsmail:
@bassner if is all ok, can you merge it.

bassner (approved):

PR #1106, opened 2022-08-14
Improve release workflow

### LINKED ISSUE #990: Add CI task to release a new version
With https://github.com/Javacord/Javacord/pull/769, a release is automatically triggered when the `-SNAPSHOT` suffix is removed from the version in the `gradle.properties` file.

However, the README is not automatically updated, so one must remember to run `./gradlew updateReadme` after removing the `-SNAPSHOT` suffix. This is error-prone.

A simple solution would be a CI task that removes the `-SNAPSHOT` suffix, updates the README, and then pushes the new commit.

### LINKED ISSUE #1033: Improve Release Workflow
There are some small issues in the new release workflow:

- The release draft has the title `Release <x>.<y>.<z>` instead of `v<x>.<y>.<z>`
- The release workflow fails to push a commit to master that appends the `-SNAPSHOT` suffix because the branch is protected. => Let the GitHub Action create a PR instead (and maybe auto-merge it)
- Gradle should automatically close and release the staging artifact in Nexus (so I don't have to do it manually everytime)

shubhwip:
resolves #1033 
resolves #990


shubhwip:
Shall we address this one #990 in the last step of this workflow after auto merging PR ?

Bastian:
Sorry for the late response. If you want to, feel free to also do #990 :)

.github/workflows/release.yml
```diff
@@ -26,17 +27,68 @@ jobs:
...
+      - name: Enable Release Pull Request Automerge
+        if: steps.cpr.outputs.pull-request-operation == 'created'
+        uses: peter-evans/enable-pull-request-automerge@v2
+        with:
+          token: ${{ secrets.PAT }}
+          pull-request-number: ${{ steps.cpr.outputs.pull-request-number }}
+          merge-method: squash
+          
+      # Reference : https://github.com/gradle/gradle-build-action    
+      - name: Setup Gradle
```
Bastian, line 91: What's the reason for the Gradle Build as the last step?

.github/workflows/release.yml
```diff
@@ -26,17 +27,68 @@ jobs:
...
+            [1]: https://github.com/peter-evans/create-pull-request
+          labels: |
+            release
+            automated pr
+          assignees: Bastian
+          reviewers: Bastian
+          team-reviewers: |
+            owners
+            maintainers
+          milestone: 1
```
Bastian, line 72: This is the wrong milestone.
The correct milestone would be dynamic (the one for the version, e.g. 17 -> https://github.com/Javacord/Javacord/milestone/17). Since it's probably pretty complicated to fine the correct milestone, it's fine to just skip this and not assign the PR to any milestone.

.github/workflows/release.yml
```diff
@@ -26,17 +27,68 @@ jobs:
...
           git config --global user.name 'github-actions[bot]'
           git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          git checkout -b release-v${{ steps.read-version.outputs.version }}
           git add gradle.properties
           git commit -am "Increment patch version and append -SNAPSHOT suffix"
-          git push
\ No newline at end of file
+          git push
+
+      # Reference : https://github.com/marketplace/actions/create-pull-request
```
Bastian, line 46: There's no need to have these "Reference" comments. It's pretty easy to find the documentation for an action if you need it + we don't have these comments on the other actions.

Bastian (changes requested):
Thanks for the PR! 🎉 
Looks good overall, just some minor problems, see the code comments :)

Bastian (approved):

Bastian:
Thank you very much! 🎉

Issue #1112, opened 2022-08-31
Update banning system


RealYusufIsmail:
https://github.com/discord/discord-api-docs/pull/5219

RealYusufIsmail:
I will be working on this

PR #1113, opened 2022-08-31
[Breaking change] Updated Ban system

RealYusufIsmail:
Removed reason as it has been removed.

Changed from delete_message_days to delete_message_seconds

Closes : #1112

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -2055,6 +2055,17 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      *
      * @param user     The user to ban.
      * @param duration The number of messages to delete within the duration. (Between 0 and 604800 seconds (7 days))
+     * @return A future to check if the ban was successful.
+     */
+    default CompletableFuture<Void> banUser(User user, Duration duration) {
+        return banUser(user.getId(), duration, null);
+    }
+
+    /**
+     * Bans the given user from the server.
+     *
+     * @param user     The user to ban.
+     * @param duration The number of messages to delete within the duration. (Between 0 and 604800 seconds (7 days))
      * @param reason   The reason for the ban.
      * @return A future to check if the ban was successful.
      */
@@ -2166,7 +2177,9 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      * @deprecated Use {@link #banUser(String, long, TimeUnit)} instead.
      */
     @Deprecated
-    CompletableFuture<Void> banUser(String userId, int deleteMessageDays, String reason);
+    default CompletableFuture<Void> banUser(String userId, int deleteMessageDays, String reason) {
+        return banUser(userId, deleteMessageDays, TimeUnit.DAYS, reason);
+    }
 
     /**
      * Bans the given user from the server.
@@ -2206,17 +2219,6 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      */
     default CompletableFuture<Void> banUser(long userId, long deleteMessageDuration, TimeUnit unit, String reason) {
         return banUser(Long.toUnsignedString(userId), deleteMessageDuration, unit, reason);
-    }
-
-    /**
-     * Bans the given user from the server.
-     *
-     * @param user     The user to ban.
-     * @param duration The number of messages to delete within the duration. (Between 0 and 604800 seconds (7 days))
-     * @return A future to check if the ban was successful.
-     */
-    default CompletableFuture<Void> banUser(User user, Duration duration) {
-        return banUser(user.getId(), duration, null);
     }
 
     /**
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -40,6 +40,7 @@ import org.javacord.api.entity.webhook.Webhook;
 import org.javacord.api.event.server.member.ServerMembersChunkEvent;
 import org.javacord.api.interaction.SlashCommand;
 import org.javacord.api.listener.server.ServerAttachableListenerManager;
+
 import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.File;
@@ -182,7 +183,7 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      *
      * @param user The user to check.
      * @return The timestamp of when this user will no longer be timed out. Empty or a timestamp in the past,
-     *         if user is currently not timed out.
+     * if user is currently not timed out.
      */
     Optional<Instant> getTimeout(User user);
 
@@ -1572,7 +1573,7 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      * <p>If you want to update several settings at once, it's recommended to use the
      * {@link ServerUpdater} from {@link #createUpdater()} which provides a better performance!
      *
-     * @param user    The user.
+     * @param user     The user.
      * @param duration The duration of the timeout.
      * @return A future to check if the update was successful.
      */
@@ -1586,9 +1587,9 @@ public interface Server extends DiscordEntity, Nameable, UpdatableFromCache<Serv
      * <p>If you want to update several settings at once, it's recommended to use the
      * {@link ServerUpdater} from {@link #createUpdater()} which provides a better performance!
      *
-     * @param user    The user.
+     * @param user     The user.
      * @param duration The duration of the timeout.
-     * @param reason  The audit log reason for this update.
+     * @param reason   The audit log reason for this update.
      * @return A future to check if the update was successful.
      */
     default CompletableFuture<Void> timeoutUser(User user, Duration duration, String reason) {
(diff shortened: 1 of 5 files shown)


javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2009,12 +2010,12 @@ default CompletableFuture<Void> banUser(User user, int deleteMessageDays) {
...
      * @param user              The user to ban.
-     * @param deleteMessageDays The number of days to delete messages for (0-7).
-     * @param reason            The reason for the ban.
+     * @param deleteMessageDuration The number of messages to delete within the duration. (0-7)
+     * @param unit                  The unit of time for the duration.
      * @return A future to check if the ban was successful.
      */
-    default CompletableFuture<Void> banUser(User user, int deleteMessageDays, String reason) {
-        return banUser(user.getId(), deleteMessageDays, reason);
+    default CompletableFuture<Void> banUser(User user, int deleteMessageDuration, TimeUnit unit) {
```
KILLEliteMaste, line 2017: Please overload with 
`banUser(User user, int deleteMessageDurationInSeconds)` 
`banUser(User user, Duration durationToDeleteMessages)`

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2009,12 +2010,12 @@ default CompletableFuture<Void> banUser(User user, int deleteMessageDays) {
      * Bans the given user from the server.
      *
      * @param user              The user to ban.
-     * @param deleteMessageDays The number of days to delete messages for (0-7).
-     * @param reason            The reason for the ban.
+     * @param deleteMessageDuration The number of messages to delete within the duration. (0-7)
```
KILLEliteMaste, line 2013: Here and the methods below please change the allowed duration in the Javadoc to `Between 0 and 604800 seconds (7 days) `

KILLEliteMaste (changes requested):
- Please do not remove the old methods and instead deprecate them so users have some time to migrate. 
- You can already switch to setting the AuditLogReason on the request instead as a parameter but **the reason itself has not been removed from the ban object**. Just how you set it has been (will be) changed.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
RealYusufIsmail: No need for the fist one as you can do that with unit

RealYusufIsmail: basically ```java
ban(user, 10, TimeUnit.Second, "")
```

Comment (RealYusufIsmail) [reply javacord-api/src/main/java/org/javacord/api/entity/server/Server.java:2034]: I cant even do the first because I would have to remove 
```java
default CompletableFuture<Void> banUser(User user, int deleteMessageDays, String reason) {
        return banUser(user.getId(), deleteMessageDays, reason);
    }
```
due to duplication

RealYusufIsmail:
@KILLEliteMaste any updates on the pr?

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2076,11 +2188,62 @@ default CompletableFuture<Void> banUser(long userId, int deleteMessageDays) {
...
+
+    /**
+     * Bans the given user from the server.
+     *
+     * @param userId   The id of the user to ban.
+     * @param duration The number of messages to delete within the duration. (Between 0 and 604800 seconds (7 days))
+     * @param reason   The reason for the ban.
+     * @return A future to check if the ban was successful.
+     */
+    CompletableFuture<Void> banUser(String userId, Duration duration, String reason);
```
KILLEliteMaste, line 2233:
```suggestion
    default CompletableFuture<Void> banUser(String userId, Duration duration, String reason){
        return banUser(userId, duration.getSeconds(), TimeUnit.SECONDS, reason)
    }
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2001,8 +2003,9 @@ default CompletableFuture<Void> banUser(User user) {
      * @param deleteMessageDays The number of days to delete messages for (0-7).
      * @return A future to check if the ban was successful.
```
KILLEliteMaste, line 2004: Missing Javadoc `deprecated` tag

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2012,19 +2015,71 @@ default CompletableFuture<Void> banUser(User user, int deleteMessageDays) {
...
+    /**
+     * Bans the given user from the server.
+     *
+     * @param user     The user to ban.
+     * @param duration The number of messages to delete within the duration. (Between 0 and 604800 seconds (7 days))
+     * @return A future to check if the ban was successful.
+     */
+    default CompletableFuture<Void> banUser(User user, Duration duration) {
+        return banUser(user.getId(), duration, null);
+    }
```
KILLEliteMaste, line 2073: This should be above the method above

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2043,9 +2098,12 @@ default CompletableFuture<Void> banUser(long userId) {
      * @param userId            The id of the user to ban.
      * @param deleteMessageDays The number of days to delete messages for (0-7).
      * @return A future to check if the ban was successful.
+     * @deprecated Use {@link #banUser(String, int, TimeUnit)} instead as
+     *             deleteMessageDays is deprecated by discord.
```
KILLEliteMaste, line 2102:
```suggestion
     * @deprecated Use {@link #banUser(String, int, TimeUnit)} instead.
```
The same for the other deprecations

KILLEliteMaste (changes requested):
> the reason itself has not been removed from the ban object. Just how you set it has been (will be) changed.

You ignored the comment. Add the reason back to everything

RealYusufIsmail:
> > the reason itself has not been removed from the ban object. Just how you set it has been (will be) changed.
> 
> You ignored the comment. Add the reason back to everything

as in the old discord reason, if so that has been removed by discord so it wont work

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2054,19 +2101,71 @@ default CompletableFuture<Void> banUser(String userId, int deleteMessageDays) {
...
      * Bans the given user from the server.
      *
      * @param userId            The id of the user to ban.
      * @param deleteMessageDays The number of days to delete messages for (0-7).
      * @param reason            The reason for the ban.
      * @return A future to check if the ban was successful.
+     * @deprecated Use {@link #banUser(String, long, TimeUnit)} instead.
      */
+    @Deprecated
     CompletableFuture<Void> banUser(String userId, int deleteMessageDays, String reason);
```
KILLEliteMaste, line 2169:
```suggestion
    default CompletableFuture<Void> banUser(String userId, int deleteMessageDays, String reason) {
        return banUser(userId, deleteMessageDays, TimeUnit.DAYS, reason);
    }
```
With this we can reduce the implementation to just 1 method

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2076,11 +2175,74 @@ default CompletableFuture<Void> banUser(long userId, int deleteMessageDays) {
...
+     * Bans the given user from the server.
+     *
+     * @param user     The user to ban.
+     * @param duration The number of messages to delete within the duration. (Between 0 and 604800 seconds (7 days))
+     * @return A future to check if the ban was successful.
+     */
+    default CompletableFuture<Void> banUser(User user, Duration duration) {
+        return banUser(user.getId(), duration, null);
+    }
+
```
KILLEliteMaste, line 2221: Why did you move the method here? You should have moved it just before `default CompletableFuture<Void> banUser(User user, Duration duration, String reason) {` line 2052

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
RealYusufIsmail: Sorry, I thought you meant to put above the method for ban which has duration

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2012,19 +2016,71 @@ default CompletableFuture<Void> banUser(User user, int deleteMessageDays) {
...
+    /**
+     * Bans the given user from the server.
+     *
+     * @param user                  The user to ban.
+     * @param deleteMessageDuration The number of messages to delete within the duration.
+     *                              (Between 0 and 604800 seconds (7 days))
+     * @param unit                  The unit of time for the duration.
+     * @return A future to check if the ban was successful.
+     */
+    default CompletableFuture<Void> banUser(User user, long deleteMessageDuration, TimeUnit unit) {
```
KILLEliteMaste, line 2035: Is there any specific reason why you choose `long` instead of `int` for the `deleteMessageDuration` ?

RealYusufIsmail: yeah, 
Screenshot 2022-09-10 at 12 19 51

KILLEliteMaste: I think it's better to change it to an `int`. Long exceeds the max value that can be used here so it's overkill and unnecessary. You can still cast the int to a long without any issues

RealYusufIsmail: ok

RealYusufIsmail: done

KILLEliteMaste (approved):
LGTM!

PR #1120, opened 2022-09-05
Add Component Only sendMessage method

AhmedA1559:
https://github.com/discord/discord-api-docs/pull/5435


KILLEliteMaste (changes requested):
You should also add other methods like we did for the EmbedBuilder
`sendMessage(HighLevelComponent)`
`sendMessage(List<HighLevelComponent>)`

AhmedA1559:
> You should also add other methods like we did for the EmbedBuilder `sendMessage(HighLevelComponent)` `sendMessage(List<HighLevelComponent>)`

There aren't any pre-existing parameter signatures for ``HighLevelComponent`` in those forms, should I add that to the other methods too like ``sendMessage(String content, List<EmbedBuilder> embeds, List<HighLevelComponent> components)``?

KILLEliteMaste (approved):
Thank you for the PR, LGTM!

PR #1131, opened 2022-09-21
Completed forum support

RealYusufIsmail:
## Checklist
- [x] I have tested this PR.

This pr adds the remaining parameters for the forum channel which discord has added recently.

A json example
```json
{
  "version" : 1688397270779,
  "type" : 15,
  "topic" : "siuuu",
  "template" : "",
  "rate_limit_per_user" : 0,
  "position" : 0,
  "permission_overwrites" : [ ],
  "parent_id" : "938122131949097053",
  "name" : "test",
  "last_message_id" : "1141797369030123531",
  "id" : "1031971612238561390",
  "icon_emoji" : {
    "name" : "🧪",
    "id" : null
  },
  "flags" : 16,
  "default_thread_rate_limit_per_user" : 0,
  "default_reaction_emoji" : {
    "emoji_name" : "👍",
    "emoji_id" : null
  },
  "available_tags" : [ {
    "name" : "3",
    "moderated" : false,
    "id" : "1032173703833854022",
    "emoji_name" : null,
    "emoji_id" : null
  } ]
}
```

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerForumChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerForumChannel.java
@@ -1,7 +1,15 @@
 package org.javacord.api.entity.channel;
 
 import org.javacord.api.entity.Mentionable;
+import org.javacord.api.entity.channel.forum.DefaultReaction;
+import org.javacord.api.entity.channel.forum.ForumLayoutType;
+import org.javacord.api.entity.channel.forum.ForumTag;
+import org.javacord.api.entity.channel.forum.SortOrderType;
 import org.javacord.api.listener.channel.server.forum.ServerForumChannelAttachableListenerManager;
+
+import java.util.EnumSet;
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -9,6 +17,92 @@ import java.util.concurrent.CompletableFuture;
  */
 public interface ServerForumChannel extends RegularServerChannel, Mentionable, Categorizable,
         ServerForumChannelAttachableListenerManager {
+
+    /**
+     * Gets the version of the forum channel.
+     *
+     * @return The version of the forum channel.
+     */
+    long getVersion();
+
+    /**
+     * Gets the topic of the forum channel.
+     *
+     * @return The topic of the forum channel.
+     */
+    Optional<String> getTopic();
+
+    /**
+     * Gets the default thread rate limit per user of the forum channel.
+     *
+     * @return The default thread rate limit per user of the forum channel.
+     */
+    int getDefaultThreadRateLimitPerUser();
+
+    /**
+     * Gets the amount of seconds a user has to wait before sending another message (0-21600).
+     *
+     * @return The amount of seconds a user has to wait before sending another message.
+     */
+    int getRateLimitPerUser();
+
+    /**
+     * Gets whether the forum channel is nsfw.
+     *
+     * @return Whether the forum channel is nsfw.
+     */
+    boolean isNsfw();
+
+    /**
+     * Gets the last message id of the forum channel.
+     *
+     * @return The last message id of the forum channel.
+     */
+    Optional<Long> getLastMessageId();
+
+    /**
+     * Gets the last message id of the forum channel as a string.
+     *
+     * @return The last message id of the forum channel as a string.
+     */
+    default Optional<String> getLastMessageIdAsString() {
+        return getLastMessageId().map(Long::toUnsignedString);
+    }
+
+    /**
+     * Gets the forum channel's flags.
+     *
+     * @return The forum channel's flags.
+     */
+    EnumSet<ChannelFlag> getFlags();
+
+    /**
+     * Gets the set of tags that can be used in the forum channel.
+     *
+     * @return The set of tags that can be used in the forum channel.
+     */
+    Set<ForumTag> getForumTags();
+
+    /**
+     * Gets the default sort type used to order the posts in the forum channel.
+     *
+     * @return The default sort type used to order the posts in the forum channel.
+     */
+    Optional<SortOrderType> getDefaultSortType();
+
+    /**
+     * Gets the default emoji shown in the add reaction button.
+     *
+     * @return The default emoji shown in the add reaction button.
+     */
+    Optional<DefaultReaction> getDefaultReaction();
+
+    /**
+     * Gets the layout type of the forum channel.
+     *
+     * @return The layout type of the forum channel.
+     */
+    Optional<ForumLayoutType> getForumLayoutType();
 
     /**
      * Creates an updater for this channel.
@@ -43,5 +137,4 @@ public interface ServerForumChannel extends RegularServerChannel, Mentionable, C
     default CompletableFuture<Void> removeCategory() {
         return createUpdater().removeCategory().update();
     }
-
 }
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/forum/DefaultReaction.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/forum/DefaultReaction.java
@@ -1,0 +1,31 @@
+package org.javacord.api.entity.channel.forum;
+
+import java.util.Optional;
+
+/**
+ * This class represents a default reaction.
+ */
+public interface DefaultReaction {
+    /**
+     * Gets the id of a guild's custom emoji.
+     *
+     * @return The id of a guild's custom emoji.
+     */
+    Optional<Long> getEmojiId();
+
+    /**
+     * Gets the id of a guild's custom emoji as a string.
+     *
+     * @return The id of a guild's custom emoji as a string.
+     */
+    default Optional<String> getEmojiIdAsString() {
+        return getEmojiId().map(Long::toUnsignedString);
+    }
+
+    /**
+     * Gets the name of a guild's custom emoji.
+     *
+     * @return The name of a guild's custom emoji.
+     */
+    Optional<String> getEmojiName();
+}
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/forum/ForumTag.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/forum/ForumTag.java
@@ -1,0 +1,41 @@
+package org.javacord.api.entity.channel.forum;
+
+import org.javacord.api.entity.DiscordEntity;
+import org.javacord.api.entity.Nameable;
+
+import java.util.Optional;
+
+/**
+ * This class represents a forum tag.
+ */
+public interface ForumTag extends DiscordEntity, Nameable {
+    /**
+     * Gets whether this tag can only be added to or removed from a thread by a moderator.
+     *
+     * @return Whether this tag is locked to MANAGE_THREADS members
+     */
+    boolean isModerated();
+
+    /**
+     * Gets the id of a guild's custom emoji.
+     *
+     * @return The id of a guild's custom emoji.
+     */
+    Optional<Long> getEmojiId();
+
+    /**
+     * Gets the id of a guild's custom emoji as a string.
+     *
+     * @return The id of a guild's custom emoji as a string.
+     */
+    default Optional<String> getEmojiIdAsString() {
+        return getEmojiId().map(Long::toUnsignedString);
+    }
+
+    /**
+     * Gets the name of a guild's custom emoji.
+     *
+     * @return The name of a guild's custom emoji.
+     */
+    Optional<String> getEmojiName();
+}
--- a/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerForumChannelImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerForumChannelImpl.java
@@ -3,12 +3,23 @@ package org.javacord.core.entity.channel;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.javacord.api.entity.DiscordEntity;
 import org.javacord.api.entity.channel.ChannelCategory;
+import org.javacord.api.entity.channel.ChannelFlag;
 import org.javacord.api.entity.channel.ServerForumChannel;
+import org.javacord.api.entity.channel.forum.DefaultReaction;
+import org.javacord.api.entity.channel.forum.ForumLayoutType;
+import org.javacord.api.entity.channel.forum.ForumTag;
+import org.javacord.api.entity.channel.forum.SortOrderType;
 import org.javacord.core.DiscordApiImpl;
+import org.javacord.core.entity.channel.forum.DefaultReactionImpl;
+import org.javacord.core.entity.channel.forum.ForumTagImpl;
 import org.javacord.core.entity.server.ServerImpl;
 import org.javacord.core.listener.channel.server.forum.InternalServerForumChannelAttachableListenerManager;
+
+import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -17,11 +28,65 @@ import java.util.concurrent.CompletableFuture;
 public class ServerForumChannelImpl extends RegularServerChannelImpl
         implements ServerForumChannel, InternalServerForumChannelAttachableListenerManager {
 
-
     /**
      * The parent id of the channel.
      */
-    private volatile long parentId;
+    private long parentId;
+
+    /**
+     * The version of the channel.
+     */
+    private long version;
+
+    /**
+     * The topic of the channel.
+     */
+    private String topic;
+
+    /**
+     * The rate limit per user of the channel.
+     */
+    private int rateLimitPerUser;
+
+    /**
+     * Whether the channel is nsfw.
+     */
+    private boolean nsfw;
+
+    /**
+     * The last message id of the channel.
+     */
+    private Long lastMessageId;
+
+    /**
+     * The flags of the channel.
+     */
+    private EnumSet<ChannelFlag> flags = EnumSet.noneOf(ChannelFlag.class);
+
+    /**
+     * The available tags of the channel.
+     */
+    private Set<ForumTag> forumTags = new HashSet<>();
+
+    /**
+     * The default sort order of the channel.
+     */
+    private SortOrderType defaultSortOrder;
+
+    /**
+     * The default reaction of the channel.
+     */
+    private DefaultReaction defaultReaction;
+
+    /**
+     * The default thread rate limit per user of the channel.
+     */
+    private int defaultThreadRateLimitPerUser;
+
+    /**
+     * The forum layout type of the channel.
+     */
+    private ForumLayoutType forumLayoutType;
 
     /**
      * Creates a new server forum channel object.
@@ -33,6 +98,29 @@ public class ServerForumChannelImpl extends RegularServerChannelImpl
     public ServerForumChannelImpl(DiscordApiImpl api, ServerImpl server, JsonNode data) {
         super(api, server, data);
         parentId = Long.parseLong(data.has("parent_id") ? data.get("parent_id").asText("-1") : "-1");
+        version = data.has("version") ? data.get("version").asLong() : -1;
+        topic = data.hasNonNull("topic") ? data.get("topic").asText() : null;
+        defaultThreadRateLimitPerUser = data.hasNonNull("default_thread_rate_limit_per_user")
+                ? data.get("default_thread_rate_limit_per_user").asInt() : 0;
+        rateLimitPerUser = data.get("rate_limit_per_user").asInt();
+        nsfw = data.has("nsfw") && data.get("nsfw").asBoolean();
+        lastMessageId = data.has("last_message_id") ? data.get("last_message_id").asLong() : null;
+        if (data.has("flags")) {
+            for (JsonNode flag : data.get("flags")) {
+                flags.add(ChannelFlag.getByValue(flag.asInt()));
+            }
+        }
+        if (data.has("available_tags")) {
+            for (JsonNode tag : data.get("available_tags")) {
+                forumTags.add(new ForumTagImpl(api, tag));
+            }
+        }
+        defaultSortOrder = data.has("default_sort_order")
+                ? SortOrderType.getByValue(data.get("default_sort_order").asInt()) : null;
+        defaultReaction = data.hasNonNull("default_reaction_emoji")
+                ? new DefaultReactionImpl(data.get("default_reaction_emoji")) : null;
+        forumLayoutType = data.has("default_forum_layout")
+                ? ForumLayoutType.getByValue(data.get("default_forum_layout").asInt()) : null;
     }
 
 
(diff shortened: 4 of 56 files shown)


RealYusufIsmail:
@KILLEliteMaste

Saladoc (changes requested):
Thanks for your contribution, but "Updated Forum" is in no way a proper description for a changeset of this size. Please update your pull request with a more in-depth description (take the new issue templates as inspiration).

Saladoc (changes requested):
I've marked several questions regarding the API parts. I've held off on reviewing the implementation because there might yet be changes following the API reviews.

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerForumChannel.java
```diff
@@ -1,15 +1,128 @@
...
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 /**
  * This class represents a server forum channel.
  */
 public interface ServerForumChannel extends RegularServerChannel, Mentionable, Categorizable,
         ServerForumChannelAttachableListenerManager {
```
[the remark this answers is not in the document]
KILLEliteMaste: A Server Forum channel only contains posts (thread channels). You can not send messages or do anything like that besides creating a (public?) thread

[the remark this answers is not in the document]
Saladoc: It's ambiguous to me whether this is a forum channel or a forum thread. Because other methods certainly make it seem like a singular thread.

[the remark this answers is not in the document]
RealYusufIsmail: Will have a look. I need a json example

javacord-api/src/main/java/org/javacord/api/event/channel/server/forum/ServerForumChannelChangeAppliedTagsEvent.java
```diff
@@ -0,0 +1,24 @@
...
+     * @return The old applied tags.
+     */
+    List<Long> getOldAppliedTags();
+
+    /**
+     * Gets the new applied tags.
+     *
+     * @return The new applied tags.
+     */
+    List<Long> getNewAppliedTags();
```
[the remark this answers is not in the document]
RealYusufIsmail: ForumTag is not cached and the applied event returns a long

javacord-api/src/main/java/org/javacord/api/event/channel/server/forum/ServerForumChannelChangeTagsEvent.java
```diff
@@ -0,0 +1,26 @@
+package org.javacord.api.event.channel.server.forum;
+
+import org.javacord.api.entity.channel.forum.ForumTag;
+
+import java.util.List;
+
+/**
+ * A server forum channel change tags event.
+ */
+public interface ServerForumChannelChangeTagsEvent extends ServerForumChannelEvent {
```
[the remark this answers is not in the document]
RealYusufIsmail: A tag was changed in this one. A new tag was added ServerForumChannelChangeAppliedTagsEvent completely different things.

javacord-api/src/main/java/org/javacord/api/event/channel/server/forum/ServerForumChannelChangeAppliedTagsEvent.java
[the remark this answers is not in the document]
KILLEliteMaste: It is, the `ServerForumChannel` has a list of `available_tags` while the threads have a new field `applied_tags` with the id of the tag. Then you just need to get the parent of the thread and get the tag by it's id. 
Btw. in your current implementation the `ServerForumChannel` has the `applied_tags` field which is wrong and a field `tags` which does not exist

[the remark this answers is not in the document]
RealYusufIsmail: This is a json example ```json
{
  id: "1019659259706888292",
  last_message_id: "1019915240034549811",
  default_thread_rate_limit_per_user: 0,
  type: 15,
  name: "support-deno",
  position: 13,
  flags: 16,
  parent_id: "904410404690739241",
  topic: "1. Every question must be related to MythCord and its modules.\n2. Every question must be related to ...",
  guild_id: "902572770469216376",
  permission_overwrites: [],
  rate_limit_per_user: 3600,
  nsfw: false,
  available_tags: [
    {
      id: "1019661812779728916",
      name: "REST",
      emoji_id: "1006325149726937139",
      emoji_name: null,
      moderated: false
    },
    {
      id: "1019661851086307389",
      name: "Gateway",
      emoji_id: "1019661320678821898",
      emoji_name: null,
      moderated: false
    },
    {
      id: "1019663161349111919",
      name: "Update discussion",
      emoji_id: "998690910097387650",
      emoji_name: null,
      moderated: true
    }
  ],
  template: "",
  default_reaction_emoji: { emoji_id: "1004874786964512798", emoji_name: null },
  default_sort_order: null
}
```

File: javacord-api/src/main/java/org/javacord/api/entity/channel/forum/DefaultReaction.java [hunk 4]
```diff
@@ -0,0 +1,29 @@
+package org.javacord.api.entity.channel.forum;
+
+/**
+ * This class represents a default reaction.
+ */
+public interface DefaultReaction {
```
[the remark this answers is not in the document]
RealYusufIsmail: Yes as the api clearly states that it is separate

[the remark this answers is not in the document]
RealYusufIsmail: I assume it is different will have a look

javacord-api/src/main/java/org/javacord/api/entity/channel/forum/ForumTag.java
```diff
@@ -0,0 +1,39 @@
...
+    default String getEmojiIdAsString() {
+        return Long.toUnsignedString(getEmojiId());
+    }
+
+    /**
+     * Gets the name of a guild's custom emoji.
+     *
+     * @return The name of a guild's custom emoji.
+     */
+    String getEmojiName();
```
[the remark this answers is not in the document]
RealYusufIsmail: This is different as in the docs it says the emoji to show in the add reaction button on a thread in a GUILD_FORUM channel

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerForumChannel.java
```diff
@@ -1,15 +1,128 @@
...
+     * @return The count of users in this forum thread channel.
+     */
+    int getMemberCount();
+
+    /**
+     * Gets the id of the last message sent in this forum thread channel.
+     *
+     * @return The id of the last message sent in this forum thread channel.
+     */
+    long getLastMessageId();
```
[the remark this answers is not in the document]
RealYusufIsmail: I dont really see an issue in exposing it.

RealYusufIsmail:
Unable to work on. Most is done just testing and some minor changes are needed so if someone wants to take over let me know.

QuadrixYT:
Please add the function

RealYusufIsmail:
Screenshot 2023-04-17 at 21 39 13
Note to self

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerForumChannelImpl.java
```diff
@@ -33,6 +100,32 @@ public class ServerForumChannelImpl extends RegularServerChannelImpl
...
+        }
+        if (data.has("available_tags")) {
+            for (JsonNode tag : data.get("available_tags")) {
+                availableTags.add(new AvailableTagImpl(api, tag));
+            }
+        }
+        template = data.has("template") ? data.get("template").asText() : null;
+        defaultSortOrder = data.has("default_sort_order")
+                ? SortOrderType.getByValue(data.get("default_sort_order").asInt()) : null;
+        defaultReaction = data.has("default_reaction_emoji")
```
Zeitheron, line 127: Should be `data.hasNonNull` as the packets may have a `"default_reaction_emoji": null` value.

RealYusufIsmail: Will be sure to check it out once done with my exams

Zeitheron:
Another nice-to-have feature would be creating a forum post itself.
The current implementation (as far as I'm aware) does not allow that.
It uses `RestEndpoint.START_THREAD_WITHOUT_MESSAGE`, it seems.
Would also be really nice to have something like `ServerForumChannel.createPost`

RealYusufIsmail:
image

RealYusufIsmail:
image
```java
class ForumListener implements ServerForumChannelChangeAppliedTagsListener, ServerForumChannelChangeAvailableTagsListener, ServerForumChannelChangeDefaultReactionListener, ServerForumChannelChangeDefaultThreadRateLimitPerUserListener, ServerForumChannelChangeFlagsListener, ServerForumChannelChangeLastMessageIdListener, ServerForumChannelChangePermissionOverwritesListener, ServerForumChannelChangePositionListener, ServerForumChannelChangeRateLimitPerUserListener, ServerForumChannelChangeTemplateListener, ServerForumChannelChangeTopicListener, ServerForumChannelChangeVersionListener, ServerForumChannelChangeIconEmojiListener {
    @Override
    public void onServerForumChannelChangeAppliedTags(ServerForumChannelChangeAppliedTagsEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old applied tags: " + event.getOldAppliedTags() + " new applied tags: " + event.getNewAppliedTags());
    }

    @Override
    public void onServerForumChannelChangeAvailableTags(ServerForumChannelChangeAvailableTagsEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old tags: " + event.getOldTags() + " new tags: " + event.getNewTags());
    }

    @Override
    public void onServerForumChannelChangeDefaultReaction(ServerForumChannelChangeDefaultReactionEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old default reaction: " + event.getOldDefaultReaction() + " new default reaction: " + event.getNewDefaultReaction());
    }

    @Override
    public void onServerForumChannelChangeDefaultThreadRateLimitPerUser(ServerForumChannelChangeDefaultThreadRateLimitPerUserEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old default thread rate limit per user: " + event.getOldDefaultThreadRateLimitPerUser() + " new default thread rate limit per user: " + event.getNewDefaultThreadRateLimitPerUser());
    }

    @Override
    public void onServerForumChannelChangeFlags(ServerForumChannelChangeFlagsEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old flags: " + event.getOldFlags() + " new flags: " + event.getNewFlags());
    }

    @Override
    public void onServerForumChannelChangeLastMessageId(ServerForumChannelChangeLastMessageIdEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old last message id: " + event.getOldLastMessageId() + " new last message id: " + event.getNewLastMessageId());
    }

    @Override
    public void onServerForumChannelChangePermissionOverwrites(ServerForumChannelChangePermissionOverwritesEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old permission overwrites: " + event.getOldPermissionOverwrites() + " new permission overwrites: " + event.getNewPermissionOverwrites());
    }

    @Override
    public void onServerForumChannelChangePosition(ServerForumChannelChangePositionEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old position: " + event.getOldPosition() + " new position: " + event.getNewPosition());
    }

    @Override
    public void onServerForumChannelChangeRateLimitPerUser(ServerForumChannelChangeRateLimitPerUserEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old rate limit per user: " + event.getOldRateLimitPerUser() + " new rate limit per user: " + event.getNewRateLimitPerUser());
    }

    @Override
    public void onServerForumChannelChangeTemplate(ServerForumChannelChangeTemplateEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old template: " + event.getOldTemplate() + " new template: " + event.getNewTemplate());
    }

    @Override
    public void onServerForumChannelChangeTopic(ServerForumChannelChangeTopicEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old topic: " + event.getOldTopic() + " new topic: " + event.getNewTopic());
    }

    @Override
    public void onServerForumChannelChangeIconEmoji(ServerForumChannelChangeIconEmojiEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old icon emoji: " + event.getOldIconEmoji() + " new icon emoji: " + event.getNewIconEmoji());
    }

    @Override
    public void onServerForumChannelChangeVersion(ServerForumChannelChangeVersionEvent event) {
        event.getApi().getChannelById("1008788690065166496").flatMap(Channel::asTextChannel).get().sendMessage("Old version: " + event.getOldVersion() + " new version: " + event.getNewVersion());
    }
}

```

RealYusufIsmail:
> Another nice-to-have feature would be creating a forum post itself. The current implementation (as far as I'm aware) does not allow that. It uses `RestEndpoint.START_THREAD_WITHOUT_MESSAGE`, it seems. Would be something like `ServerForumChannel.createPost`

might be for another pr

javacord-api/src/main/java/org/javacord/api/entity/channel/forum/AvailableTag.java
```diff
@@ -0,0 +1,41 @@
...
+     * @return Whether this tag is locked to MANAGE_THREADS members
+     */
+    boolean isModerated();
+
+    /**
+     * Gets the id of a guild's custom emoji.
+     *
+     * @return The id of a guild's custom emoji.
+     */
+    long getEmojiId();
```
felldo, line 24: image

So this should probably return an Optional

javacord-core/src/main/java/org/javacord/core/entity/channel/forum/IconEmojiImpl.java
```diff
@@ -0,0 +1,43 @@
...
+     */
+    private final Long id;
+
+    /**
+     * Creates a new icon emoji.
+     *
+     * @param data The json data of the icon emoji.
+     */
+    public IconEmojiImpl(JsonNode data) {
+        this.name = data.get("name").asText();
```
felldo, line 25: Here I get a NPE when running with my bot.
The channel json seems to contains either `"icon_emoji": null,` or 
```json
"icon_emoji": {
      "name": "💬",
      "id": null
    }
```
Though in my examples I never got an ID if it was an object.

But This doesn't seem to be released yet? Haven't found anything in the docs. So might be worth asking in the DDevs server or DAPI server

javacord-api/src/main/java/org/javacord/api/entity/channel/forum/DefaultReaction.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Gets the id of a guild's custom emoji.
+     *
+     * @return The id of a guild's custom emoji.
+     */
+    long getEmojiId();
```
felldo, line 14: Can be null
image

felldo (changes requested):
- You are missing the forum layout types https://discord.com/developers/docs/resources/channel#channel-object-forum-layout-types
- Remove everything related to the position events/listener/fields because the ServerChannel already takes care of a channels position. See `ServerChannelChangePositionListener`
- The `ServerForumChannel` extends `RegularServerChannel` channel which already takes care of the overwritten permissions. So remove everything permission overwrite related you have added in this PR
- The NSFW listener also already exists. So remove everything in that regard (except the instance field from the forum channel, because previously only text channels were nsfw) and move the check/dispatcher logic in the `ChannelUpdateHandler` from the `handleServerTextChannel` method to `handleRegularServerChannel`. Also adjust the `ServerChannelChangeNsfwFlagListener` interface to be able to attach it to the `ServerForumChannel`
- You did not add the `applied_tags` field to threads. It is also not possible to set them atm.
image

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerForumChannelImpl.java
```diff
@@ -33,6 +112,34 @@ public class ServerForumChannelImpl extends RegularServerChannelImpl
...
+            for (JsonNode flag : data.get("flags")) {
+                flags.add(ChannelFlag.getByValue(flag.asInt()));
+            }
+        }
+        if (data.has("available_tags")) {
+            for (JsonNode tag : data.get("available_tags")) {
+                availableTags.add(new AvailableTagImpl(api, tag));
+            }
+        }
+        template = data.has("template") ? data.get("template").asText() : null;
```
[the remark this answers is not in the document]
wasdennnoch: Forum templates are not documented and are only available to a limited subset of servers, so I wouldn't rely on this field always existing or being non-null. Since it's not documented, they can change or remove it at any time without it being considered a breaking change.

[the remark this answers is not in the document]
felldo: Actually, just remove it 
image

javacord-core/src/main/java/org/javacord/core/entity/channel/forum/IconEmojiImpl.java
RealYusufIsmail: Will have a look. But those emojis from my assumption are the new ones on the new mobile version UI. So could be saved there and present in the json.

RealYusufIsmail: image

As I thoght

RealYusufIsmail: They are for all channels shall I just remove and implement it in a separate pr.

wasdennnoch: Not a maintainer, but I'd say remove for now, given they are undocumented and have, afaik, already breaking changed the format a few weeks ago.

felldo: I agree. At least it's not our fault if then something breaks

felldo:
A quick look on the `applied_tags` field shows that you added it to the forum channel but it is a field on the thread channel
image

Also instead of saving the ids, get the tags from the forum channel. You should be able to use `getParent` to check for the forum channel and get it's tags which you can add to the list of applied tags in the thread.
In that regard it should probably renamed from `AppliedTag` to `ForumTag` as it is in the docs

RealYusufIsmail:
> A quick look on the `applied_tags` field shows that you added it to the forum channel but it is a field on the thread channel image
> 
> Also instead of saving the ids, get the tags from the forum channel. You should be able to use `getParent` to check for the forum channel and get it's tags which you can add to the list of applied tags in the thread. In that regard it should probably renamed from `AppliedTag` to `ForumTag` as it is in the docs

It says only for forum channels and Media channels. Not all threads

felldo:
> It says only for forum channels and Media channels. Not all threads

Doesn't change the fact, that you put this field in the forum channel and will never contain something. So for now just add it to the thread channel and get the tags from the parent channel if this channel has tags. Otherwise we would need a new thread channel, which will be returned from the forum and media channel but theres currently an internal discussion about the channel structure, so just for now it is fine imo.

RealYusufIsmail:
@felldo made the changes can you check on them please.

javacord-core/src/main/java/org/javacord/core/entity/channel/ServerThreadChannelUpdaterDelegateImpl.java
```diff
@@ -70,6 +78,19 @@ public void setSlowmodeDelayInSeconds(int delay) {
...
+    public void setForumTags(Set<ForumTag> forumTags) {
+        channel.asServerForumChannel().ifPresent(serverForumChannel -> {
+            for (ForumTag tag : forumTags) {
+                if (!serverForumChannel.getForumTags().contains(tag)) {
+                    throw new IllegalArgumentException("The tag " + tag.getName() + " is not a valid forum tag");
+                }
+            }
+
+            this.forumTags = forumTags;
+        });
```
felldo, line 91: This will not work. You get the Thread (`channel`) as a forum channel. 
And the error message is wrong. It should say something like "this tag does not belong to the forum channel of this thread"

javacord-core/src/main/java/org/javacord/core/entity/channel/forum/DefaultReactionImpl.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Used to set the name of a guild's custom emoji.
+     *
+     * @param emojiName The name of a guild's custom emoji.
+     * @return The current instance in order to chain call methods.
+     */
+    public DefaultReactionImpl setEmojiName(String emojiName) {
+        this.emojiName = emojiName;
+        return this;
+    }
```
felldo, line 39:
```suggestion
```
This is probably not used anywhere, as we always create a new object in the `ChannelUpdater`?

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -421,6 +455,165 @@ private void handleServerForumChannel(JsonNode jsonChannel) {
...
+                    new ServerForumChannelChangeVersionEventImpl(channel, newVersion, oldVersion);
+
+            api.getEventDispatcher().dispatchServerForumChannelChangeVersionEvent(
+                    (DispatchQueueSelector) channel.getServer(), channel.getServer(), channel, event);
+        }
+
+
+        String oldTopic = channel.getTopic().orElse(null);
+        String newTopic = jsonChannel.has("topic") && !jsonChannel.get("topic").isNull()
+                ? jsonChannel.get("topic").asText() : null;
```
felldo, line 474:
```suggestion
        String newTopic = jsonChannel.hasNonNull("topic") ? jsonChannel.get("topic").asText() : null;
```

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -421,6 +455,165 @@ private void handleServerForumChannel(JsonNode jsonChannel) {
...
+                    (DispatchQueueSelector) channel.getServer(), channel.getServer(), channel, event);
+        }
+
+        EnumSet<ChannelFlag> oldFlags = channel.getFlags();
+        EnumSet<ChannelFlag> newFlags = EnumSet.noneOf(ChannelFlag.class);
+        if (jsonChannel.hasNonNull("flags")) {
+            for (JsonNode flag : jsonChannel.get("flags")) {
+                newFlags.add(ChannelFlag.getByValue(flag.asInt()));
+            }
+        }
```
felldo, line 530: Again, this is a bitfield, not an array

javacord-core/src/main/java/org/javacord/core/util/handler/channel/ChannelUpdateHandler.java
```diff
@@ -421,6 +455,165 @@ private void handleServerForumChannel(JsonNode jsonChannel) {
...
+
+        if (!oldAvailableTags.equals(newAvailableTags)) {
+            channel.setAvailableTags(newAvailableTags);
+            ServerForumChannelChangeForumTagsEvent event =
+                    new ServerForumChannelChangeForumTagsEventImpl(channel, newAvailableTags, oldAvailableTags);
+
+            api.getEventDispatcher().dispatchServerForumChannelChangeForumTagsEvent(
+                    (DispatchQueueSelector) channel.getServer(), channel.getServer(), channel, event);
+        }
```
felldo, line 556: In a previous comment, I said that you also need to modify the tags of the thread channel. For that you should have added a `getActiveThreads` method or something similar to the forum channel depending on what channels we actually get as a helper method, where you have to check the applied tags and remove them if a thread also has them.

felldo (changes requested):
- The name of `getAppliedTags` and `getAvailableTags` was perfectly fine. They describe what you get. But the return type name was wrong in this case. Because when calling `getAppliedTags` why would you get an `AvailableTag`? Therefore rename every `forumTag` variable back to it's original name.
- The ServerForumChannelBuilder is missing.
- The ServerThreadChannelBuilder need to be enhanced to create threads in server forum channels https://discord.com/developers/docs/resources/channel#start-thread-in-forum-or-media-channel . Also it would be nice if there was a method in the `ServerForumChannel` to quickly create a thread with various parameters. As there a fields that are not available for the other kind of threads when creating them, just add a little note to the Javadoc that these are only available for forum/media channels.
- The ServerThreadChannel should also have a method to update the applied tags more easily.
- Please take your time when doing something large like this. There are many mistakes (wrong or missing) which can be easily avoided by looking at the documentation carefully. A review process for something like this also consumes a lot of time, especially if I have to open the documentation by myself and verify every single line for correctness. And please test this PR more, like creating a forum channel and a thread withing it with tags and so on.

Issue #1114, opened 2022-09-01
Feature: Add a MessageReplyListener to Message


feddevanderlist:
At this moment in time you can do this ` message.addReactionAddListener()` to a message.
Now is there a feature called `reply`
![image](https://user-images.githubusercontent.com/23259736/187938958-19b5059e-2d05-4a7e-891b-0eb684b2d5b6.png)

For I message I would like to have the option to follow all replies on a single message.
This should be called `Message#addMessageCreateListener` 
Something like this already exists for User `User#addMessageCreateListener`

PR #1139, opened 2022-10-03
Add the ability to listen for replies to specific messages

Mysterypotatoguy:
## Checklist

## Changelog
- Adds message reply listeners (via `Message#addMessageCreateListener`)

### Breaking Changes
N/A

## Description

Adds the ability to attach message create listeners to messages, which will be fired when the message receives a reply 

Closes: #1114

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/listener/message/MessageReplyListener.java
+++ b/javacord-api/src/main/java/org/javacord/api/listener/message/MessageReplyListener.java
@@ -1,0 +1,25 @@
+package org.javacord.api.listener.message;
+
+import org.javacord.api.event.message.MessageReplyEvent;
+import org.javacord.api.listener.GloballyAttachableListener;
+import org.javacord.api.listener.ObjectAttachableListener;
+import org.javacord.api.listener.channel.TextChannelAttachableListener;
+import org.javacord.api.listener.server.ServerAttachableListener;
+import org.javacord.api.listener.user.UserAttachableListener;
+import org.javacord.api.listener.webhook.WebhookAttachableListener;
+
+/**
+ * This listener listens to message replies.
+ */
+@FunctionalInterface
+public interface MessageReplyListener extends ServerAttachableListener, UserAttachableListener,
+        WebhookAttachableListener, TextChannelAttachableListener, MessageAttachableListener,
+        GloballyAttachableListener, ObjectAttachableListener {
+
+    /**
+     * This method is called when a message is created which replies to another.
+     *
+     * @param event The event.
+     */
+    void onMessageReply(MessageReplyEvent event);
+}
--- a/javacord-core/src/main/java/org/javacord/core/event/message/MessageReplyEventImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/event/message/MessageReplyEventImpl.java
@@ -1,0 +1,29 @@
+package org.javacord.core.event.message;
+
+import org.javacord.api.entity.message.Message;
+import org.javacord.api.event.message.MessageReplyEvent;
+
+/**
+ * The implementation of {@link MessageReplyEvent}.
+ */
+public class MessageReplyEventImpl extends CertainMessageEventImpl implements MessageReplyEvent {
+
+    private final Message referencedMessage;
+
+    /**
+     * Creates a new event instance.
+     *
+     * @param message The created message.
+     * @param referencedMessage The message which is being replied to.
+     */
+    public MessageReplyEventImpl(Message message, Message referencedMessage) {
+        super(message);
+        this.referencedMessage = referencedMessage;
+    }
+
+    @Override
+    public Message getReferencedMessage() {
+        return referencedMessage;
+    }
+
+}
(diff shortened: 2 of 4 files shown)


KILLEliteMaste (commented):
Imo the `MessageCreateListener` doesn't fit name wise for this purpose. We are not listening for MessageCreates but rather for MessageReplies. And adding a `MessageCreateListener` to a message is not that obvious that you are waiting for replies. For that reason I think a new `MessageReplyListener` (can extend from `MessageCreateListener`) would make more sense

KILLEliteMaste:
> Should a `MessageReplyListener` have its own event `MessageReplyEvent` or should it still use the `MessageCreateEvent`?

I think it's own event is better. Then we can also add `getReferencedMessage` and `requestReferencedMessage` to the event like we have for the message content.

Saladoc:
Semantics question for `MessageReplyEvent`: 
If attached to a `User` / `MessageAuthor`, does it 
 * Trigger for messages in Reply to that user / author?
 * Trigger for messages by that user / author?

Mysterypotatoguy:
It doesn't necessarily have to be attachable to `User`/`MessageAuthor`

Vampire:
Why not both?

javacord-api/src/main/java/org/javacord/api/listener/message/MessageReplyListener.java
```diff
@@ -0,0 +1,20 @@
...
+import org.javacord.api.event.message.MessageReplyEvent;
+import org.javacord.api.listener.ObjectAttachableListener;
+import org.javacord.api.listener.user.UserAttachableListener;
+
+/**
+ * This listener listens to message replies.
+ */
+@FunctionalInterface
+public interface MessageReplyListener extends UserAttachableListener, MessageAttachableListener,
+        ObjectAttachableListener {
```
KILLEliteMaste, line 12: Shouldn't the interface extend the same interfaces as the `MessagecreateListener`?

javacord-core/src/main/java/org/javacord/core/event/message/MessageReplyEventImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.javacord.api.entity.message.Message;
+import org.javacord.api.event.message.MessageReplyEvent;
+
+/**
+ * The implementation of {@link MessageReplyEvent}.
+ */
+public class MessageReplyEventImpl extends CertainMessageEventImpl implements MessageReplyEvent {
+
+    private Message referencedMessage;
```
KILLEliteMaste, line 11:
```suggestion
    private final Message referencedMessage;
```

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageCreateHandler.java
```diff
@@ -98,6 +105,20 @@ private void handle(TextChannel channel, JsonNode packet) {
                 author.asUser().orElse(null),
                 author.isWebhook() ? author.getId() : null,
                 event);
+
+        optionalReferencedMessage.ifPresent(referencedMessage -> {
```
KILLEliteMaste, line 109:
```suggestion
        message.getReferencedMessage().ifPresent(referencedMessage -> {
```
And remove the variable assignment in line 93

KILLEliteMaste (changes requested):

javacord-api/src/main/java/org/javacord/api/listener/message/MessageReplyListener.java
Mysterypotatoguy: Not sure, do we really want this being attachable to Servers and TextChannels? The semantics seem very weird to me

Vampire: Why not? You listen to replies happening in a specific channel or server. I cannot see where that would be weird.

javacord-core/src/main/java/org/javacord/core/util/handler/message/MessageCreateHandler.java
```diff
@@ -98,6 +104,31 @@ private void handle(TextChannel channel, JsonNode packet) {
...
+            referencedMessage.getUserAuthor().ifPresent(users::add);
+            message.getUserAuthor().ifPresent(users::add);
+
+            Set<Long> webhookIds =  new HashSet<>();
+            if (message.getAuthor().isWebhook()) {
+                webhookIds.add(message.getAuthor().getId());
+            }
+            if (referencedMessage.getAuthor().isWebhook()) {
+                webhookIds.add(referencedMessage.getAuthor().getId());
+            }
```
KILLEliteMaste, line 120:
```suggestion
            message.getAuthor().getWebhookId().ifPresent(webhookIds::add);
            referencedMessage.getAuthor().getWebhookId().ifPresent(webhookIds::add);
```

KILLEliteMaste (changes requested):

KILLEliteMaste (approved):
LGTM!

PR #1143, opened 2022-10-09
Remove okhttp shutdown calls when disconnecting the api

KILLEliteMaste:
## Checklist

## Changelog
- Internal: Remove OkHttp shutdown calls when disconnecting the DiscordApi

## Description
Earlier versions of OkHttp used non-daemon threads which forced us to forcefully shutdown the OkHttpClient. In version 4.5.1 OkHttp switched to daemon threads.


Vampire (approved):
lgtm

Issue #805, opened 2021-07-07
Concept: Cache module


Bastian:
I've already suggested a design for an external cache module in https://github.com/Javacord/Javacord/issues/454#issuecomment-832957623.
However, since this suggestion was not well received because of the immutability, this issue proposes a mutable external cache that also allows sharing between multiple shards and allows for different data stores like Redis.

Similar to the immutable cache, the mutable cache holds simple immutable and serializable POJOs that do not contain references to other entities except by their ids. 

Since the module is independent of Javacord, we can use another JVM language like Kotlin to reduce the necessary boilerplate.

### Storing data objects

* The DataEntity interface that every data object has to extend:

```kt
interface DataEntity {
    val indices: Array<Index>
}

data class Index(
    val name: String,
    val value: String
)
```

* The data entities itself

```kt
@Serializable
data class ServerData(
    val id: Long,
    val name: String,
    // More fields ...
): DataEntity {
    override val indices: Array<Index>
        get() = arrayOf(Index("ID", id.toString()))
}
```

```kt
@Serializable
data class ServerTextChannelData(
    val id: Long,
    val name: String,
    val serverId: Long,
    // More fields ...
): DataEntity {
    override val indices: Array<Index>
        get() = arrayOf(
            Index("ID", id.toString()),
            Index("SERVER", serverId.toString()),
        )
}
```

* A DataStore that hold the data entities and manages their indices

```kt
interface DataStore {
    fun insert(value: DataEntity)
    fun get(clazz: Class<out DataEntity>, indexName: String, indexValue: String): List<DataEntity>
    fun delete(clazz: Class<out DataEntity>, indexName: String, indexValue: String)
}
```
The implementation of the data store is extremely simple and can look like this for a classic in-memory store:
```kt
class InMemoryDataStore: DataStore {

    val entities = HashMap<String, MutableList<DataEntity>>()
    val reverseIndex = HashMap<DataEntity, Array<String>>()

    override fun insert(value: DataEntity) {
        val indexKeys = value.indices.map {
            value::class.simpleName + "-" + it.name + "-" + it.value
        }.toTypedArray()
        indexKeys.forEach {
            val list = entities[it] ?: mutableListOf()
            list.add(value)
        }
        reverseIndex[value] = indexKeys
    }

    override fun get(clazz: Class<out DataEntity>, indexName: String, indexValue: String): List<DataEntity> {
        return entities[clazz.simpleName + "-" + indexName + "-" + indexValue] ?: listOf()
[...]

Issue #356, opened 2018-08-06
REST only module


Bastian:
Sometimes you don't need websocket events, but just perform REST-actions.
It might be a good idea to move all REST-logic into a more lightweight module.

RealYusufIsmail:
@Bastian would it be ok if i tried to implement this

Saladoc:
You are free to work on as many issues as you like. Bear in mind however that this is a large change that will require a lot of work, review and discussion, much more than the forum channel pr you already are taking care of.

PR #1144, opened 2022-10-09
Add RestRequestListener

felldo:
## Checklist

## Changelog
- Added a new listener `RestRequestListener` which dispatches each time a rest request is executed

## Description
This PR will add a new listener which will be dispatched each time a rest request is executed. This is helpful for i.e. monitoring purposes.
Furthermore this PR will now auto close the Response bodies like the Javadoc of OkHttp states.
In addition some data was moved to the OkHttpResponseImpl from the RestRequestResult

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/event/connection/RestRequestEvent.java
+++ b/javacord-api/src/main/java/org/javacord/api/event/connection/RestRequestEvent.java
@@ -1,0 +1,26 @@
+package org.javacord.api.event.connection;
+
+import org.javacord.api.event.Event;
+import org.javacord.api.util.auth.Request;
+import org.javacord.api.util.auth.Response;
+
+/**
+ * A rest request event.
+ */
+public interface RestRequestEvent extends Event {
+
+    /**
+     * Gets the request that is used to perform the rest request call.
+     *
+     * @return The request.
+     */
+    Request getRequest();
+
+    /**
+     * Gets the response of the rest request call.
+     *
+     * @return The response.
+     */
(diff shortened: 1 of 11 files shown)


Saladoc:
I'm really not sure if I'm a fan of this. Not because I don't see the value of being able to intercept and monitor the rest requests, but because we're mixing in a low-level event with all the high-level ones.

All events passed through Javacords event bus relate to actual state changes on discords side - granted, they may be caused by (bot) user actions but still, the events displayed comes from discord. This would be the first event that happens on our side and also the only one that's on a low technical level rather than representing a state change.

I like the idea, but I am not at all convinced just slapping it on the event dispatcher. Things that deal with the internals of Javacord should go a different way. The idea of specific modules (#356 and #805) has been brought up before and I feel a "Monitoring Module" that can be activated and deactivated on startup as needed and interacted with would be a far more robust and adequate solution.

KILLEliteMaste:
I see what you mean. Usually I would agree on moving it into it's own module but isn't it a bit overkill? While the cache and rest module will be quiet large and encapsulate the code very good like it should be, I don't think there a lot of metrics that we can actually dispatch. We could for example in addition to the rest requests also expose the websocket events (there's already a PR that suggest to add an event for any gateway event #917) but I can't think of anything that might be worth exposing too for monitoring purposes. Even if there are 2 or 3 more possible metrics, it will most likely be just a few lines to actually expose them when doing it like in this PR. 
Also what do you think how should the module look like and how should we interact with it to get the data if we do not dispatch it?

Issue #906, opened 2021-10-26
Add get thread member endpoint


KILLEliteMaste:
https://github.com/discord/discord-api-docs/pull/3998

RealYusufIsmail:
Will do this one

PR #1155, opened 2022-10-18
added get thread member endpoint

RealYusufIsmail:
## Checklist

## Changelog
- Added the ability to get a thread member by their user id.
- Deprecated getThreadMembers in favor of requestThreadMembers

## Description

Closes: https://github.com/Javacord/Javacord/issues/906

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannel.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannel.java
@@ -183,12 +183,42 @@ public interface ServerThreadChannel extends ServerChannel, TextChannel, Mention
     }
 
     /**
+     * Gets a thread member by their id.
+     *
+     * @param userId The id of the user.
+     * @return The thread member.
+     */
+    CompletableFuture<ThreadMember> requestThreadMemberById(long userId);
+
+    /**
+     * Gets a thread member by their id.
+     *
+     * @param userId The id of the user.
+     * @return The thread member.
+     */
+    default CompletableFuture<ThreadMember> requestThreadMemberById(String userId) {
+        return requestThreadMemberById(Long.parseLong(userId));
+    }
+
+    /**
+     * Gets all members in this thread.
+     * Requires the {@link org.javacord.api.entity.intent.Intent#GUILD_MEMBERS} intent.
+     *
+     * @return All members in this thread.
+     * @deprecated Use {@link #requestThreadMembers()} instead.
+     */
+    @Deprecated
+    default CompletableFuture<Set<ThreadMember>> getThreadMembers() {
+        return requestThreadMembers();
+    }
+
+    /**
      * Gets all members in this thread.
      * Requires the {@link org.javacord.api.entity.intent.Intent#GUILD_MEMBERS} intent.
      *
      * @return All members in this thread.
      */
-    CompletableFuture<Set<ThreadMember>> getThreadMembers();
+    CompletableFuture<Set<ThreadMember>> requestThreadMembers();
 
     /**
      * Creates an updater for this thread.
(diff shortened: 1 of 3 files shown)


RealYusufIsmail:
@Saladoc can you review this pr please

javacord-api/src/main/java/org/javacord/api/entity/channel/ThreadMember.java
```diff
@@ -64,5 +64,4 @@ default CompletableFuture<User> requestUser() {
      * @return Any user-thread settings.
      */
     int getFlags();
-
```
Mysterypotatoguy, line 67: Revert line removal

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannel.java
```diff
@@ -182,6 +182,24 @@ default CompletableFuture<Void> leaveThread() {
         return getServer().leaveServerThreadChannel(getId());
     }
 
+    /**
+     * Used to get a thread member by their id.
+     *
+     * @param userId The id of the user.
+     * @return The thread member.
+     */
+    CompletableFuture<ThreadMember> getThreadMember(long userId);
```
Mysterypotatoguy, line 191:
```suggestion
    CompletableFuture<ThreadMember> requestThreadMemberById(long userId);
```

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannel.java
```diff
@@ -182,6 +182,24 @@ default CompletableFuture<Void> leaveThread() {
...
+
+    /**
+     * Used to get a thread member by their id.
+     *
+     * @param userId The id of the user.
+     * @return The thread member.
+     */
+    default CompletableFuture<ThreadMember> getThreadMember(String userId) {
```
Mysterypotatoguy, line 199:
```suggestion
    default CompletableFuture<ThreadMember> requestThreadMemberById(String userId) {
```

Saladoc (changes requested):
Methods that (might) do a rest request under the hood should be named `requestXY`. Methods that return a local value (for example from cache) should be named `getXY`.

Also please keep your method descriptions consistent. We usually describe what a method does (`Gets a Foo from the Bar`), not what the method is used for (`Used to get a Foo from the Bar`)

Mysterypotatoguy (changes requested):
I'd argue `getThreadMembers` may need to be renamed to `requestThreadMembers` too, but this would be a breaking change

Saladoc:
> I'd argue `getThreadMembers` may need to be renamed to `requestThreadMembers` too, but this would be a breaking change

We can rename the method, keep an old one in that calls the new one and deprecate it in favor of the new one.

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerThreadChannel.java
```diff
@@ -182,13 +182,31 @@ default CompletableFuture<Void> leaveThread() {
...
+    }
+
     /**
      * Gets all members in this thread.
      * Requires the {@link org.javacord.api.entity.intent.Intent#GUILD_MEMBERS} intent.
      *
      * @return All members in this thread.
      */
-    CompletableFuture<Set<ThreadMember>> getThreadMembers();
+    CompletableFuture<Set<ThreadMember>> requestThreadMembers();
```
Saladoc, line 209: This is a breaking change. Please restore the old method and deprecate it in favor of the new method.

Saladoc (changes requested):

Saladoc (approved):

Mysterypotatoguy (approved):

PR #1156, opened 2022-10-18
Change get and request sticker string id parse methods to be more consistent

KILLEliteMaste:
## Checklist

### Breaking Changes
- get/requestSticketById(String) methods do no longer throw an exception when the id is being parsed and instead return an empty optional or request a not existing sticker 

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApi.java
@@ -942,7 +942,9 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
         try {
             return getUserById(Long.parseLong(id));
         } catch (NumberFormatException e) {
-            return getUserById(-1);
+            CompletableFuture<User> future = new CompletableFuture<>();
+            future.completeExceptionally(e);
+            return future;
         }
     }
 
@@ -1414,7 +1416,7 @@ public interface DiscordApi extends GloballyAttachableListenerManager {
         try {
             return getStickerById(Long.parseLong(id));
         } catch (NumberFormatException exception) {
-            throw new IllegalArgumentException("The id must be a number!");
+            return Optional.empty();
         }
     }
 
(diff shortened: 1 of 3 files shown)


javacord-api/src/main/java/org/javacord/api/DiscordApi.java
```diff
@@ -1436,7 +1436,7 @@ default CompletableFuture<Sticker> requestStickerById(String id) {
         try {
             return requestStickerById(Long.parseLong(id));
         } catch (NumberFormatException exception) {
-            throw new IllegalArgumentException("The ID must be a number!");
+            return requestStickerById(-1);
```
Saladoc, line 1439:
```suggestion
            CompletableFuture<Sticker> future = new CompletableFuture<>();
            future.completeExceptionally(exception);
            return future;
```
There is no reason to swallow the exception and deprive the user of debugging information.

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -3515,7 +3515,7 @@ default CompletableFuture<Sticker> requestStickerById(String id) {
         try {
             return requestStickerById(Long.parseLong(id));
         } catch (NumberFormatException exception) {
-            throw new IllegalArgumentException("The ID must be a number.");
+            return requestStickerById(-1);
```
Saladoc, line 3518:
```suggestion
            CompletableFuture<Sticker> future = new CompletableFuture<>();
            future.completeExceptionally(exception);
            return future;
```
Same as above. After the switch to Java 11 we can also do `return CompletableFuture.failedFuture(exception);`

Saladoc (changes requested):

javacord-api/src/main/java/org/javacord/api/DiscordApi.java
KILLEliteMaste: Seems there is another inconsistency. `DiscordApi#getUserById` looks the same like this

Saladoc: I'd argue we should return a failed future in both cases. Depending on how many methods have this problem we can create an issue or make the changes all in this PR. If we have a way to pass the exception to the user without breaking the method contract, we should do so and not throw away the information.

Saladoc (approved):

Issue #928, opened 2021-12-28
Add CompletableFutures for Channels, Guilds, ..


MichaelMelzer:
So it looks like there is no way to call Javacord API implementaitons to request channels, guilds, messages and so on by id, even tho the Discord API does provide that functionality?

I get that Javacord tries to cache all of the channels and guilds, but sometimes they are not in cache and the Optional is not present (like in my edge case in #927 )

I'd say it would generally be a great feature to add a CompletableFuture for those objects, in the form of DiscordApi#requestServer or similar

RealYusufIsmail:
Will implement this

Issue #814, opened 2021-07-13
Add the request method for member,guild,role,channel and other


MCArTyR:
It is necessary to add the ability to get an object by ID. I want to check if there is a member on the server, but if the member is not cached, then I cannot do this, so I need a method similar to <Server>.requestMember(id: long) that would return CompletableFuture<Optional<User> that is , if there is such a member on the server, it will return User to me, if not, then null

ShindouMihou:
Hello, before making an issue for stuff like this, please check the documentations or try out your IDE's autocomplete or ask in Discord where most people would most likely be able to help you faster.

To answer this issue, such method already exists under `DiscordApi` called `getUserById` which returns a CompletableFuture.

https://docs.javacord.org/api/v/3.3.2/org/javacord/api/DiscordApi.html#getUserById-long-

Bastian:
I think the problem is, that there's no method to get a member (i.e., a user that has also role information).
We haven an open PR for this feature: https://github.com/Javacord/Javacord/pull/684

For other entities, I don't think that we can provide such methods. I'm not sure, whether or not Discord even offers such endpoints, but if it does, we would run into problems with related entities.
E.g. how would `Role#getServer()` work if we only get the role but not the server? Every entity besides members are always in the cache (for the shard that "owns" them). Javacord is not intended to be used without an active cache (except members).

MCArTyR:
I need Server#requestMemberById method (request) that returns CompletableFuture<Role>,
https://discord.com/developers/docs/resources/guild#get-guild-member
I need SlashCommandInteraction#requestChannel that return CompletableFuture<Optional<TextChannel>> if channel_id is present and Channel object not cached
About `Role#getServer()` you can store server id to Role object.

RealYusufIsmail:
This issue seems similar to #928. I will be implementing this.

PR #1159, opened 2022-10-20
Added CompletableFutures for Channels, Guilds, etc

RealYusufIsmail:
## Checklist

## Changelog
In this pr I have added CompletableFutures for Channels, Guilds and more.

### Breaking Changes
Moved the interface member to the API folder instead of core

## Description
If there is any suggestions can you let me know please

Closes: #928 
Closes: #814

Additions
- [x] request Server
- [x] request channels
- [ ] request sticker
- [x] request member
- [ ] request role


Saladoc:
> In this pr I have added CompletableFutures for Channels, Guilds and more.
`CompletableFuture`s are a part of the standard library, they don't need adding. You probably mean "Added request methods for Channels, ..." 

> Moved the interface member to the API folder instead of core
Absolutely not. Exposing `Member` is a major change that is not as simple as moving the interface. See #1095 for the progress on that change and the cascading api changes.

KILLEliteMaste:
I don't think it is a good idea to add these methods at the moment. There is no purpose for requesting a server or channel, because they are essential for Javacord and are always cached. It would be better if we add these methods once we have a REST module like proposed in this issue #356 .
And like Saladoc mentioned, for exposing the member interface a much bigger refactor is required. Btw. the current development is continued on the `refactor/expose-member` Javacord branch because there is still quiet a bit todo so everyone can contribute with PRs

RealYusufIsmail:
Ah i see, will close this pr then

wasdennnoch:
> are always cached

Not with sharding; it's sometimes useful to fetch guilds from different shards, especially since a lot of Javacord methods require full instances of entities and not just IDs. But yeah, given the large pending member refactor, any other big refactors should probably wait, and the sharding thing might be fixed by a more low-level REST module, depending on how that will be architectured.

PR #1162, opened 2022-10-22
Added support for news channel

### LINKED ISSUE #453: News Channels
So news channels are a thing now. 

The good news: They are basically `ServerTextChannel`s so an easy way of implementing them would be to create an interface extending `ServerTextChannel` that gives a different `ChannelType` and we'd be fully backwards compatible.

The bad news: news channels don't support slow mode, so the slow mode methods would become unnecessary. In the spirit of clarity those would have to be dropped from the news channel interface, meaning that the whole structure would have to be rewritten. If done right we'd still be backwards compatible but would require code to be recompiled.

My suggestion, which I might start implementing later, would be to introduce a new interface called `ServerMessageChannel` which both `ServerTextChannel` and, (consistently named) `ServerNewsChannel` inherit from, only the former of which will retain slow mode specific methods. This would also be a working foundation for implementing store channels, which are even more nebulous in purpose and properties than the elusive news channels.

RealYusufIsmail:
## Checklist

## Changelog

- Added support for news channel
- Removed channel type SERVER_STORE_CHANNEL as it was removed by discord a long time ago
- ServerTextChannel now extends ServerMessageChannel and so does NewsChannel
This is because everything apart from slow mode is the same. Slowmode is in the interface ServerTextChannel

### Breaking Changes
 ServerTextChannel and, (consistently named) ServerNewsChannel now inhert ServerMessageChannel

## Description

Closes: #453 


RealYusufIsmail:
Unable to work on. Most is done just testing and some minor changes are needed so if someone wants to take over let me know.

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerMessageChannel.java
```diff
@@ -0,0 +1,390 @@
...
+import org.javacord.api.entity.Mentionable;
+import org.javacord.api.entity.message.Message;
+import org.javacord.api.entity.server.ArchivedThreads;
+import org.javacord.api.entity.webhook.WebhookBuilder;
+
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public interface ServerMessageChannel extends RegularServerChannel, TextChannel, Mentionable, Categorizable {
```
RealYusufIsmail, line 12: Make it extend TextableRegularServerChannel

PR #1165, opened 2022-10-23
Added NitroType to User

RealYusufIsmail:
## Checklist

## Changelog
- added nitro type to user object

## Description

Closes: NaN

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/user/NitroType.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/user/NitroType.java
@@ -1,0 +1,44 @@
+package org.javacord.api.entity.user;
+
+public enum NitroType {
+    NONE(0),
+    NITRO_CLASSIC(1),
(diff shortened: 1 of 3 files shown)


Mysterypotatoguy:
I'm not sure how you tested this, as the `premium_type` field isn't available on users through the gateway. You would need to use OAuth2 with the `identify` scope to be able to fetch this field

RealYusufIsmail:
> I'm not sure how you tested this, as the `premium_type` field isn't available on users through the gateway. You would need to use OAuth2 with the `identify` scope to be able to fetch this field

How would you do that

Mysterypotatoguy:
You would need to use OAuth2 and authenticate with the user, which is out of scope for Javacord currently

RealYusufIsmail:
> You would need to use OAuth2 and authenticate with the user, which is out of scope for Javacord currently

oh ok

PR #1167, opened 2022-10-24
Change how the delete_message_seconds field is set when banning a user

RealYusufIsmail:
## Checklist

## Changelog
- Change how the delete_message_seconds field is set when banning a user from as query parameter to the request json body

This means that setting a delete message duration will now work

## Description

Closes: NaN


Saladoc (changes requested):
No issues on the code front but please make the commit message and changelog elements more descriptive. Consider the target audience for both.

KILLEliteMaste:
No idea what you were thinking of but the new title / commit message is by far worse than the first you had. It doesn't mention in any way that it changes the way how the `delete_message_seconds` parameter is set.
Please change the commit message to: Change how the `delete_message_seconds` field is set when banning a user from as query parameter to the request json body

PR #1170, opened 2022-10-29
Fixes for ServerVoiceChannel audio connection

KILLEliteMaste:
## Checklist

## Changelog
- Fixed moving the audio connection to the new ServervoiceChannel if the bot gets moved
- Fixed not closing the audio connection when kicking the bot from a ServerVoiceChannel

## Description
This PR fixes some issues with the audio connection. The audio connection will now be closed when kicking the bot from a ServerVoiceChannel and the audio connection will be moved to the new ServerVoiceChannel if the bot gets moved

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/guild/VoiceStateUpdateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/guild/VoiceStateUpdateHandler.java
@@ -29,6 +29,7 @@ import org.javacord.core.util.gateway.PacketHandler;
 import org.javacord.core.util.logging.LoggerUtil;
 
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Handles the voice state update packet.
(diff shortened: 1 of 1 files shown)


Mysterypotatoguy:
It's not documented still but the method for moving between audio channels (even moves initiated by other users) was changed when servers changed away from having set audio regions.
Moves now cause the current websocket to disconnect and require reconnecting to the new channel's websocket address, for which you need to receive both a `VOICE_STATE_UPDATE` and `VOICE_SERVER_UPDATE`
I attempted to rework this a while back but I couldn't fit it into the current architecture

The close-on-kick part looks fine to me

KILLEliteMaste:
> It's not documented still but the method for moving between audio channels (even moves initiated by other users) was changed when servers changed away from having set audio regions.
> Moves now cause the current websocket to disconnect and require reconnecting to the new channel's websocket address, for which you need to receive both a `VOICE_STATE_UPDATE` and `VOICE_SERVER_UPDATE`
> I attempted to rework this a while back but I couldn't fit it into the current architecture

But then my current approach might be better than the broken behavior. At least it works until someone (maybe you?) can rework it properly.

Mysterypotatoguy:
I think it will visually move to the correct channel but the actual audio connection will be dead

KILLEliteMaste:
> I think it will visually move to the correct channel but the actual audio connection will be dead

I tested it and it works pretty good until you move the bot very fast between 2 channels in like <0.5 seconds, then it stops playing. But when you move it afterwards again to a voice channel it starts playing again.

Mysterypotatoguy:
...Interesting. I will have to take another look when I'm free

KILLEliteMaste:
@Mysterypotatoguy are you okay with this or should this PR not be merged until you had another look on it?

Mysterypotatoguy (approved):
I'm okay with this being merged if it fixes a majority of the voice moving problems. Can do another pass for the edge cases later

Issue #1116, opened 2022-09-02
Changes to the SlashCommandInteractionOptionsProvider


KILLEliteMaste:
The SlashCommandInteractionOptionsProvider is an interface that contains methods to get options from a slash command interaction. Both the `SlashCommandInteraction` and `SlashCommandInteractionOption` interface implements it to be able to navigate through the options more conveniently. 
Currently methods like `getOptionLongValueByName` only make sense if you are on the last "level" of the command. That means if the current option is a `SUB_COMMAND` or if the command has no sub command. 
Therefore invoking these methods on the following command: 
`/base group subcommand <arg1>`
they will always return an empty optional if you are currently on the `base` or `group` level. 

To address this issue we should decide what should be done.
On the one side changing the behavior to get the options from `SlashCommandInteraction#getArguments` and renaming the methods would solve it but the API is not as fluent anymore if you are navigating though the options by yourself. Then if you are on the last level the `getOptionLongValueByName` methods would be missing. Then it might be a bit weird to call `getArgumentLongValueByName` although the options of the current option contains the arguments. 
On the other side this change is not that bad because there shouldn't be any reason to get the arguments that way if you can call the convenience method `getArguments` or the new added methods. 
Imo the advantages we get by doing this change is bigger than a bit less fluent API

PR #1171, opened 2022-10-30
Rename SlashCommandInteractionOptionsProvider methods

KILLEliteMaste:
## Checklist

### Breaking Changes
- Renamed SlashCommandOptionsProvider methods that are only usable for arguments of a slash command from `getOption*` to `getArgument*` and changed the implementation from getting the options from the option to getting always the command arguments if there are any

## Description
See issue for more detailed thoughts on this

Closes: #1116 

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,8 +6,8 @@ plugins {
     `java-library`
     `maven-publish`
     signing
-    id("com.github.johnrengelman.shadow") version "2.0.4" apply false
-    id("net.researchgate.release") version "2.7.0" apply false
+    id("com.github.johnrengelman.shadow") version "7.1.2" apply false
+    id("net.researchgate.release") version "3.0.2" apply false
 }
 
 repositories {
--- a/javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteraction.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteraction.java
@@ -1,18 +1,6 @@
 package org.javacord.api.interaction;
 
-import java.util.List;
-
 public interface SlashCommandInteraction extends ApplicationCommandInteraction, SlashCommandInteractionOptionsProvider {
-
-    /**
-     * Gets the arguments of this slash command if there are any.
-     *
-     * <p>This is a shorthand method to avoid checking for Subcommmands or SubcommandGroups
-     * to get the slash command arguments.
-     *
-     * @return The argument options.
-     */
-    List<SlashCommandInteractionOption> getArguments();
 
     /**
      * Gets the full command name of this slash command including the name of the Subcommand and SubcommandGroup.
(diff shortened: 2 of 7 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteraction.java
```diff
@@ -1,19 +1,7 @@
...
 
-    /**
-     * Gets the arguments of this slash command if there are any.
-     *
-     * <p>This is a shorthand method to avoid checking for Subcommmands or SubcommandGroups
-     * to get the slash command arguments.
-     *
-     * @return The argument options.
-     */
-    List<SlashCommandInteractionOption> getArguments();
```
Saladoc, line 15: Is there no way to keep / deprecate this method for another release?

build.gradle.kts
```diff
@@ -6,8 +6,8 @@ plugins {
     `java-library`
     `maven-publish`
     signing
-    id("com.github.johnrengelman.shadow") version "2.0.4" apply false
-    id("net.researchgate.release") version "2.7.0" apply false
+    id("com.github.johnrengelman.shadow") version "7.1.2" apply false
+    id("net.researchgate.release") version "3.0.2" apply false
```
Saladoc, line 10: Sneay version upgrades not mentioned in any commit or pr description - should probably be a separate commit?

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteractionOptionsProvider.java
```diff
@@ -37,129 +47,142 @@ default Optional<SlashCommandInteractionOption> getOptionByName(String name) {
...
     }
 
     /**
-     * Gets the string value of an option having the specified name.
+     * Gets the string representation value of an argument having the specified name.
      *
-     * @param name The name of the option to search for.
-     * @return An Optional with the string value of such an option if it exists; an empty Optional otherwise
+     * @param name The name of the argument to search for.
+     * @return An Optional with the string value of such an argument if it exists; an empty Optional otherwise
```
Saladoc, line 66: Can we get the difference between a string representation and a string value mentioned in the javadoc? The `@return` tag is the same as for `#getStringValueByName(String)` which might be confusing.

Saladoc (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/SlashCommandInteraction.java
KILLEliteMaste: There shouldn't be any noticeable difference for users except when using reflection as it implements the `SlashCommandINteractionOptionsProvider`

build.gradle.kts
KILLEliteMaste: Yeah might be better :) Will do it later if it gets approved

javacord-core/build.gradle.kts
```diff
@@ -2,18 +2,21 @@ import aQute.bnd.version.MavenVersion.parseMavenString
...
     mavenCentral()
+    maven(url="https://m2.dv8tion.net/releases")
 }
 
 dependencies {
     api(project(":javacord-api")) {
         isTransitive = true
     }
 
+    implementation("com.sedmelluq:lavaplayer:1.3.77")
```
Saladoc, line 18: Where is this dependency used? Is this a leftover from testing another branch?

Saladoc (approved):

PR #1172, opened 2022-10-30
add intent to example in README

Lainika:
Intent is required for message content

## Checklist
- [X] I have tested this PR[^1].
- [X] I have read the [contributing guidelines](./CONTRIBUTING.md).

## Changelog
*Documentation only*

## Description

Closes: NaN

### DIFF
--- a/.github/README_template.md
+++ b/.github/README_template.md
@@ -12,7 +12,8 @@ It has rich [documentation](#-documentation) and an [awesome community on Discor
 
 ## 🎉 Basic Usage
 
-The following example logs the bot in and replies to every "!ping" message with "Pong!". 
+The following example logs the bot in and replies to every "!ping" message with "Pong!". Note that message content is a 
+[privileged Intent](https://javacord.org/wiki/basic-tutorials/gateway-intents.html#privileged-intents) and needs to specifically be enabled.
 
 ```java
 public class MyFirstBot {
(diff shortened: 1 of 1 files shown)


feddevanderlist:
I would personally vote for either of the 2 following options instead of this:
`.setAllNonPrivilegedIntents()`
or
`setIntents(Intent.MESSAGE_CONTENT)`

Instead of setting all intents and risking security issues or people not thinking about access.

Saladoc:
First option won't work because message content is privileged. Since privileged intents must be enabled in the application dashboard I don't see a harm in setting all intents. There's not really a way to accidently set them all in both steps.

KILLEliteMaste:
First of all thank you for mentioning this issue we currently have in our example :)

I think it is better if you use `addIntent(Intent.MESSAGE_CONTENT)` instead of setting all intents because the user might not need all privileged intents.

> I would personally vote for either of the 2 following options instead of this

Both options will not work. Saladoc already mention the first reason. The 2nd reason is when you use `setIntent` you will override all intents and then you will miss out on the `GUILD_MESSAGES` to receive messages in the first place

Lainika:
Thank you.
Right that's better, thanks for the advice.

KILLEliteMaste:
@all-contributors please add @Lainika for doc

PR #1175, opened 2022-10-31
Add new Javacord compatible library "Discord Interaction Handler"

KILLEliteMaste:
## Checklist

## Changelog
- Added new Javacord compatible library `Discord Interaction handler` and removed `sdcf4j` because it is deprecated


Saladoc (changes requested):
I feel that there should be something resembling a "neutral" ordering to the command frameworks - i suggest alphabetical, Can you also make it clear that it is not intended to be an exhaustive list?

Saladoc (approved):

PR #1177, opened 2022-10-31
Deprecate ApplicationCommand delete methods

KILLEliteMaste:
## Checklist

## Changelog
- Deprecated `ApplicationCommand#deleteGlobal` and `deleteForServer` in favor of `delete`

## Description
I don't know why this was changed back then to methods that specifically deleted them globally or on a server. But currently it doesn't make sense as an application commands only exists in either one of them if you have i.e. a `SlashCommand` instance. That means `deleteForServer` will never work if you call it on a global command you got with `DiscordApi#getGlobalSlashCommands` because if you create the exact same command as a server command they do not share the same id.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
@@ -79,7 +79,7 @@ public interface ApplicationCommand extends DiscordEntity, Specializable<Applica
     boolean isDisabledByDefault();
 
     /**
-     * Gets whether this command is able to be used in DMs.
+     * Gets whether this command can be used in DMs.
      * Will always return {@code false} for server commands.
      *
      * @return Whether the command is enabled in DMs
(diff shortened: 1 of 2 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
```diff
@@ -112,19 +112,30 @@ public interface ApplicationCommand extends DiscordEntity, Specializable<Applica
      */
     boolean isServerApplicationCommand();
 
+    /**
+     * Deletes this application.
```
Saladoc, line 116:
```suggestion
     * Deletes this application ommand.
```

Saladoc (changes requested):

javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommand.java
```diff
@@ -79,7 +79,7 @@ public interface ApplicationCommand extends DiscordEntity, Specializable<Applica
     boolean isDisabledByDefault();
 
     /**
-     * Gets whether this command is able to be used in DMs.
+     * Gets whether this command is can be used in DMs.
```
Saladoc, line 82:
```suggestion
     * Gets whether this command can be used in DMs.
```

Saladoc (changes requested):

Saladoc (approved):

PR #1179, opened 2022-11-01
Add generate changelog Gradle task

KILLEliteMaste:
## Checklist

## Changelog
Internal changes only


Saladoc (changes requested):
### Tested: 
- [X] Task/Build runs with Java17
- [X] Task/Build runs with Java11
- [X] Output validity

### Identified (potential) problems:
`gradlew generateChangelog > changelog.md` also includes the gradle build output, making it less viable to quickly create a file and then edit it

Suggestions: 
 - accept a parameter that designates an output file
 - also write to `changelog.md` by default and add `changelog.md` to `.gitignore`

---

Title line is output as `# ? Changelog` 
 - Identify possible encoding issue and fix

KILLEliteMaste:
> `gradlew generateChangelog > changelog.md` also includes the gradle build output, making it less viable to quickly create a file and then edit it

You can add the parameter `-q` to log only errors (quiet mode) and then it work perfectly fine.

I don't think we need to add a parameter for the output file. After you have executed the script you can simply copy the file in the root directory or pipe the content to a file you want

> Title line is output as # ? Changelog

This is a windows cmd encoding issue. The generated changelog in the file correctly display this line and piping the output to a file also works fine on Linux

Saladoc (approved):

PR #1182, opened 2022-11-03
Throw an exception when accessing the message content

KILLEliteMaste:
## Checklist

### Breaking Changes
- Throw an exception when the content of a message is accessed without having the `MESSAGE_CONTENT` enabled

## Description
The only purpose of this intent is to receive the content of a message. And since a lot of people are not aware that they have to enable this intent, it is better to directly throw an exception and let the user know that they are missing this intent before they need to ask for help on why this is not working.

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/MessageImpl.java
@@ -6,6 +6,7 @@ import org.javacord.api.entity.DiscordEntity;
 import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.emoji.CustomEmoji;
 import org.javacord.api.entity.emoji.Emoji;
+import org.javacord.api.entity.intent.Intent;
 import org.javacord.api.entity.message.Message;
 import org.javacord.api.entity.message.MessageActivity;
 import org.javacord.api.entity.message.MessageAttachment;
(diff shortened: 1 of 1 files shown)


Mysterypotatoguy:
I'm not 100% sure if this is a good idea. Bots can still receive message content without the intent if they are mentioned within the message

KILLEliteMaste:
> I'm not 100% sure if this is a good idea. Bots can still receive message content without the intent if they are mentioned within the message

You are completely right, I forgot about that. But to circumvent this we can add an additional check `content.isEmpty()` where we can be absolutely sure that the bot is not mentioned

Vampire:
I guess just throwing if the content is actually empty and the intent is missing makes sense.
We actually wanted to sprinkle exceptions everyhwere a method is used that needs an intent that is not present.
It was just not done because of the time pressure to get the support out, but it should definitely be added more widespread imho. :-)

KILLEliteMaste:
> We actually wanted to sprinkle exceptions everyhwere a method is used that needs an intent that is not present.

This can be done in a follow-up PR as for example the `GUILD_MEMBERS` intent will require a larger rewrite of methods due to that a lot of methods check the cache before they i.e. request a user. I just want to get this through because this is the most problematic one users are facing currently

Vampire:
Sure, I didn't mean it should be done here, just second the general tactic

Vampire (approved):
LGTM

Mysterypotatoguy (approved):

PR #1185, opened 2022-11-05
Improve throwing exception for missing MESSAGE_CONTENT intent

KILLEliteMaste:
## Checklist

## Changelog
- Added `Message#canYouReadContent` which checks if you can read the content of a message if you do not have the `MESSAGE_CONTENT` intent enabled to cover special cases like DMs, bot mentions or your own messages


javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -719,10 +721,20 @@ default CompletableFuture<Void> unpin() {
...
+     */
+    boolean canYouReadContent();
+
     /**
      * Gets the content of the message.
      *
      * @return The content of the message.
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
```
Saladoc, line 737:
```suggestion
     * @throws MissingIntentException If you can't get the message content because you are not mentioned in the
     *                                Message, not the author of the message, the message is not a DM, and you
     *                                don't have the {@link Intent#MessageContent} intent set.
```
Existing description reads like an exception is always thrown when the intent is missing.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -756,6 +768,9 @@ default CompletableFuture<Void> unpin() {
      * </ul>
      *
      * @return The readable content of the message.
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
      */
```
Saladoc, line 774: Same suggestion as above.

javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
```diff
@@ -965,6 +980,9 @@ default Optional<CompletableFuture<Message>> requestReferencedMessage() {
      * Gets all channels mentioned in this message.
      *
      * @return All channels mentioned in this message.
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
```
Saladoc, line 985: Same suggestion as above.

javacord-api/src/main/java/org/javacord/api/event/message/CertainMessageEvent.java
```diff
@@ -65,6 +67,9 @@ default List<MessageAttachment> getMessageAttachments() {
      *
      * @return The content of the event's message.
      * @see Message#getContent()
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
```
Saladoc, line 72: Suggestion as above

javacord-api/src/main/java/org/javacord/api/event/message/CertainMessageEvent.java
```diff
@@ -75,6 +80,9 @@ default String getMessageContent() {
      *
      * @return The readable content of the event's message.
      * @see Message#getReadableContent()
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
```
Saladoc, line 85: Suggestion as above

javacord-api/src/main/java/org/javacord/api/event/message/OptionalMessageEvent.java
```diff
@@ -45,6 +47,9 @@ default Optional<List<MessageAttachment>> getMessageAttachments() {
      *
      * @return The content of the event's message.
      * @see Message#getContent()
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
```
Saladoc, line 52: Suggestion as above, but with the addition "if the message is present."

javacord-api/src/main/java/org/javacord/api/event/message/OptionalMessageEvent.java
```diff
@@ -55,6 +60,9 @@ default Optional<String> getMessageContent() {
      *
      * @return The readable content of the event's message.
      * @see Message#getReadableContent()
+     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
+     *                                receive the content of a message except if the message
+     *                                is a DM, you are mentioned, or it's your own message.
```
Saladoc, line 65: Suggestion as above, but with the addition "if the message is present."

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
```diff
@@ -40,7 +40,8 @@ public static WebhookMessageBuilder fromMessage(Message message) {
         WebhookMessageBuilder builder = new WebhookMessageBuilder()
                 .setDisplayAvatar(message.getAuthor().getAvatar())
                 .setDisplayName(message.getAuthor().getDisplayName());
-        builder.getStringBuilder().append(message.getContent());
+
+        builder.getStringBuilder().append(message.canYouReadContent() ? message.getContent() : "");
```
Saladoc, line 44: What's the pro/con of not throwing an exception here too?

If I were to redirect a message via a webhook I'd expect the content to show up on the target as well. Specifically since the description says **which would produce the same text as the given message** - if the text can not be acquired, an exception should be thrown.

Saladoc (changes requested):
Both `CertainMessageEvent` and `OptionalMessageEvent` should also provide shortcut methods for the `canYouReadContent()` method. It does not seem handy to have the methods throwing the exception on the event as well but not the method that allows to perform the check for safe content retrieval.

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
KILLEliteMaste: The issue with this is when you want to copy a message and do not have access to the content, you can not prefill the message builder at all. So instead of only having to set a content you want, you also have to set attachments, embeds and everything else which is the purpose of this method

Saladoc: I'm assuming it's not possible to receive attachments, embeds etc if you can't access the message content. If we have a message with unknowable content (and are not aware), manual filling of a WebhookMessageBuilder would, if not checked properly, throw an exception. `fromMessage` in the proposed form would *not* fill in all content elements correctly (since there is no way to do so with the message content being "unknowable") and then *not* indicate that to the user, which in my book is a major fauxpas.

We could offer an overload method that will only prefill the fields that are accessible, but silently changing the content conflicts with both expectations and the methods own documentation.

KILLEliteMaste: Oh you are up to something. Currently content, embeds, attachments, and components require the intent according to the documentation. This is also missing in the introduction of the message content exception PR. In that regard I would agree with you because these are literally the most essential fields in a message especially if you want to copy it.

javacord-api/src/main/java/org/javacord/api/event/message/OptionalMessageEvent.java
KILLEliteMaste: I don't think this belongs in there and should if added, added above to i.e. `Gets the content of the event's message.`

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
```diff
@@ -41,7 +41,7 @@ public static WebhookMessageBuilder fromMessage(Message message) {
                 .setDisplayAvatar(message.getAuthor().getAvatar())
                 .setDisplayName(message.getAuthor().getDisplayName());
 
-        builder.getStringBuilder().append(message.canYouReadContent() ? message.getContent() : "");
+        builder.getStringBuilder().append(message.getContent());
```
Saladoc, line 44: Method javadoc is missing `@throws` annotation, should link to `Message#getContent` for further explanation.

javacord-api/src/main/java/org/javacord/api/event/message/OptionalMessageEvent.java
```diff
@@ -37,6 +47,14 @@ default Optional<MessageAuthor> getMessageAuthor() {
      *
      * @return All attachments of the event's message.
      * @see Message#getAttachments()
+     * @throws MissingIntentException If not a single of the following requirements is met:
```
Saladoc, line 50:
```suggestion
     * @throws MissingIntentException If the message is present and not a single of the following requirements is met:
```

javacord-api/src/main/java/org/javacord/api/event/message/OptionalMessageEvent.java
```diff
@@ -47,9 +65,14 @@ default Optional<List<MessageAttachment>> getMessageAttachments() {
      *
      * @return The content of the event's message.
      * @see Message#getContent()
-     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
-     *                                receive the content of a message except if the message
-     *                                is a DM, you are mentioned, or it's your own message.
+     * @throws MissingIntentException If not a single of the following requirements is met:
```
Saladoc, line 68:
```suggestion
     * @throws MissingIntentException If the message is present and not a single of the following requirements is met:
```

javacord-api/src/main/java/org/javacord/api/event/message/OptionalMessageEvent.java
```diff
@@ -60,9 +83,14 @@ default Optional<String> getMessageContent() {
      *
      * @return The readable content of the event's message.
      * @see Message#getReadableContent()
-     * @throws MissingIntentException If you are missing the {@link Intent#MESSAGE_CONTENT} intent which is needed to
-     *                                receive the content of a message except if the message
-     *                                is a DM, you are mentioned, or it's your own message.
+     * @throws MissingIntentException If not a single of the following requirements is met:
```
Saladoc, line 86:
```suggestion
     * @throws MissingIntentException If the message is present and not a single of the following requirements is met:
```

Saladoc (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/message/WebhookMessageBuilder.java
```diff
@@ -35,11 +36,13 @@ public class WebhookMessageBuilder {
      *
      * @param message The message to copy.
      * @return A webhook message builder which would produce the same text as the given message.
+     * @throws MissingIntentException See Javadoc of{@link Message#getContent()} for further explanation.
```
Saladoc, line 39:
```suggestion
     * @throws MissingIntentException See Javadoc of {@link Message#getContent()} for further explanation.
```

Saladoc (approved):
lgtm except for one missing space

PR #1204, opened 2023-01-10
Fixes NullPointerException in TextInput when null is passed down from `TextInput`.

ShindouMihou:
## Checklist

## Changelog
- Fixed `NullPointerException` when passing down `null` in `TextInputBuilder#setValue`

## Description
There was a recent encounter with a `NullPointerException` on Beemo that was caused by passing a nullable variable down to `TextInputBuilder().setValue(...)` (which totally accepted it) and tried to convert it to JSON but ended up failing down at that specific line when trying to do `isEmpty`.
(Stacktrace that depicts exact line of error)
image

Closes: NaN

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/TextInputBuilderDelegateImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/message/component/internal/TextInputBuilderDelegateImpl.java
@@ -53,7 +53,7 @@ public class TextInputBuilderDelegateImpl implements TextInputBuilderDelegate {
 
     @Override
     public void setValue(String value) {
-        this.value = value;
+        this.value = value == null ? "" : value;
     }
 
     @Override


KILLEliteMaste:
Currently the value field of a text input is returned as a simple string. And as far as I have seen when receiving an interaction with a text input components it always contains this field.
Now when you allow null values, you can build a TextInput component with nullable `value` field which where the `getValue` method of the TextInput return also null while the method signature is `String` instead of `Optional<String>`. 
We can solve this in 2 ways:
1. When setting the value in the builder check if it is null and exchange it with and empty string
2. Change the method signature for `TextInput#getValue` to Optional<String>

Personally I would go with 1 as it is not a breaking change and as previosuly said I haven't seen the interaction text inputs contain a missing value field so far.

ShindouMihou:
Proceeded with option number one since it sounds the most optimal to me, changed it specifically on the builder delegate since that seems to be where most of the logic should happen.

KILLEliteMaste (approved):
LGTM! Thanks :)

PR #1209, opened 2023-02-02
Let InteractionOriginalResponseUpdater implement Deletable.

### LINKED ISSUE #1203: InteractionOriginalResponseUpdater does not implement Deletable
### Javacord version

3.7.0

### Steps to reproduce

make an instance of InteractionOriginalResponseUpdater
try adding it to an ArrayList<Deletable>

### What is expected?

The instance of InteractionOriginalResponseUpdater was expected to be addable to the AraryList<Deletable>

### What is actually happening?

'add(org.javacord.api.entity.Deletable)' in 'java.util.ArrayList' cannot be applied to '(org.javacord.api.interaction.callback.InteractionOriginalResponseUpdater)'

### Any additional comments?

Since InteractionOriginalResponseUpdater implements delete() in the same way as the interface Deletable imposes, I expected to be able to add it to my ArrayList of Deletables, but there is no super interface of the InteractionOriginalResponseUpdater that implements the Deletable interface.

Saladoc:
## Checklist
- [X] I have tested this PR[^1].
- [X] I have read the [contributing guidelines](https://github.com/Javacord/Javacord/blob/master/.github/CONTRIBUTING.md).

## Changelog
- InteractionOriginalResponseUpdater now implements Deletable

## Description

Closes: #1203 

Notes: Audit reasons being ignored if not supported is made explicit in the documentation. InteractionOriginalResponseUpdater is safe to store and reuse.

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
@@ -3,6 +3,7 @@ package org.javacord.api.entity.message.internal;
 import org.javacord.api.entity.message.Message;
 import org.javacord.api.entity.message.MessageFlag;
 import org.javacord.api.interaction.InteractionBase;
+
 import java.util.EnumSet;
 import java.util.concurrent.CompletableFuture;
 
@@ -29,7 +30,18 @@ public interface InteractionMessageBuilderDelegate extends MessageBuilderBaseDel
      * @param interaction The interaction.
      * @return The completable future when the message has been deleted.
      */
-    CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction);
+    default CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction) {
+        return deleteInitialResponse(interaction, null);
+    }
+
+    /**
+     * Delete the original response message.
+     *
+     * @param interaction The interaction.
+     * @param auditLogMessage The audit log message to attach.
+     * @return The completable future when the message has been deleted.
+     */
+    CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction, String auditLogMessage);
 
     /**
      * Edits the message.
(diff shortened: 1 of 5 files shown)


javacord-core/src/main/java/org/javacord/core/interaction/InteractionOriginalResponseUpdaterImpl.java
```diff
@@ -52,11 +54,21 @@ public CompletableFuture<Message> update() {
...
 
+    @Override
+    public DiscordApi getApi() {
+        return this.interaction.getApi();
+    }
+
     @Override
     public CompletableFuture<Void> delete() {
         return this.delegate.deleteInitialResponse(interaction);
     }
```
KILLEliteMaste, line 65: Can be removed

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
```diff
@@ -31,6 +32,15 @@ public interface InteractionMessageBuilderDelegate extends MessageBuilderBaseDel
      */
     CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction);
```
KILLEliteMaste, line 33: Can be default and you can call `deleteInitialResponse(InteractionBase interaction, String auditLogMessage)` instead

javacord-core/src/main/java/org/javacord/core/entity/message/InteractionMessageBuilderDelegateImpl.java
```diff
@@ -59,6 +60,17 @@ public CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction
                 .execute(result -> null);
     }
 
+    @Override
```
KILLEliteMaste, line 63: The method above is no longer needed if you made the method default

felldo (changes requested):

Saladoc:
I wanted to omit "breaking" api changes but you're right, of those are internal anyway and using null as an audit log reason is safe all up to the `RestRequest` implementation.

javacord-api/src/main/java/org/javacord/api/entity/message/internal/InteractionMessageBuilderDelegate.java
```diff
@@ -29,7 +30,18 @@ public interface InteractionMessageBuilderDelegate extends MessageBuilderBaseDel
      * @param interaction The interaction.
      * @return The completable future when the message has been deleted.
      */
-    CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction);
+    default CompletableFuture<Void> deleteInitialResponse(InteractionBase interaction) {
+        return deleteInitialResponse(interaction, null);
+    }
```
KILLEliteMaste, line 35: Actually, this can be removed since passing null as the reason anywhere else has the same effect as this method currently.
In the `InteractionMessageBuilder` the method overload with the reason is also missing,

felldo (changes requested):

PR #1216, opened 2023-03-02
Fix deprecated IP Discovery UDP packet size

KILLEliteMaste:
## Checklist

## Changelog
- Changed deprecated IP Discovery UDP packet size to 74 which is required from the 15.03.23 onwards. **IMPORTANT** Everyone who is using voice has to upgrade until then otherwise it will no longer work

## Description
image

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
@@ -8,14 +8,12 @@ import org.javacord.core.audio.AudioConnectionImpl;
 import org.javacord.core.entity.server.ServerImpl;
 import org.javacord.core.event.audio.AudioSourceFinishedEventImpl;
 import org.javacord.core.util.logging.LoggerUtil;
-
 import java.io.IOException;
 import java.net.DatagramPacket;
 import java.net.DatagramSocket;
 import java.net.InetSocketAddress;
 import java.net.SocketException;
 import java.nio.ByteBuffer;
-import java.util.Arrays;
 
 public class AudioUdpSocket {
 
@@ -47,8 +45,8 @@ public class AudioUdpSocket {
      * Creates a new audio udp socket.
      *
      * @param connection The audio connection that uses the socket.
-     * @param address The address to connect to.
-     * @param ssrc The ssrc.
+     * @param address    The address to connect to.
+     * @param ssrc       The ssrc.
      * @throws SocketException If the socket could not be opened,
      *                         or the socket could not bind to the specified local port.
      */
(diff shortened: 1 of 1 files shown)


javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
```diff
@@ -79,12 +79,12 @@ public void setSecretKey(byte[] secretKey) {
      * @see <a href="https://discord.com/developers/docs/topics/voice-connections#ip-discovery">Discord Docs</a>
      */
     public InetSocketAddress discoverIp() throws IOException {
-        byte[] buffer = new byte[70];
+        byte[] buffer = new byte[74];
         ByteBuffer.wrap(buffer).putInt(0, ssrc);
```
Mysterypotatoguy, line 83:
```suggestion
        ByteBuffer.wrap(buffer)
            .putShort((short) 1)
            .putShort((short) 70)
            .putInt(ssrc);
```

Ref: https://discord.com/developers/docs/topics/voice-connections#ip-discovery
Apparently this changed in 2019 and we never noticed

javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
```diff
@@ -79,12 +79,12 @@ public void setSecretKey(byte[] secretKey) {
...
         // send the byte array which contains the ssrc
         socket.send(new DatagramPacket(buffer, buffer.length, address));
         // create a new buffer which is used to receive data from discord
-        buffer = new byte[70];
+        buffer = new byte[74];
         socket.receive(new DatagramPacket(buffer, buffer.length));
         // gets the ip of the packet
         String ip = new String(Arrays.copyOfRange(buffer, 3, buffer.length - 2)).trim();
```
Mysterypotatoguy, line 90:
```suggestion
        String ip = new String(buffer, 8, buffer.length - 2).trim();
```

Mysterypotatoguy (changes requested):

javacord-core/src/main/java/org/javacord/core/util/gateway/AudioUdpSocket.java
KILLEliteMaste: Had to change this to `new String(buffer, 8, buffer.length - 10).trim();` otherwise it would throw an error. There was also an issue with the port, which was hardcoded so it would obviously no longer work since the array changed

Mysterypotatoguy: Ah yes, port should be
`int port = ByteBuffer.wrap(new byte[]{buffer[69], buffer[68]}).getShort() & 0xffff;` or
`int port = ByteBuffer.wrap(new byte[]{buffer[buffer.length -1], buffer[buffer.length - 2]}).getShort() & 0xffff;`

My bad on the `String`, the last parameter is a length not an end index

Mysterypotatoguy (approved):

PR #1218, opened 2023-03-18
Cleanup and simplify all (Interaction)MessageBuilders and add support for extended markdown

felldo:
## Checklist

## Changelog
- Housekeeping on all MessageBuilders. There are now separate interfaces for specific methods like affecting the content. Users shouldn't notice anything


felldo:
This PR now also adds support for extended markdown like lists and headers

Issue #1222, opened 2023-03-25
Audio connection can't be set up if command is used in a text in voice channel


RealMarvn:
### Javacord version

3.8.0

### Steps to reproduce

In my play command I get the voiceChannel by checking if the user of a server is in a voice channel:
```java
        user.getConnectedVoiceChannel(server).ifPresentOrElse(voiceChannel -> {
```

After a few checks if I am able to join that channel and if I have enough permissions to speak ,
I try to create an audio connection:

```java
                    voiceChannel.connect().thenAccept(audioConnection -> { 
```

And now my program stops. There is now error or anything I get. It won't run the code with this AudioConnection. It's like the program ist still waiting for retrieving the audioConnection. After that. No application command works. It's like the bot is stuck trying to retrieve it. 
!!!THIS BUG ONLY HAPPENS IF I POST THIS COMMAND IN A VOICE CHANNEL. IN A NORMAL TEXT CHANNEL IT WORKS!!!

### What is expected?

The bot should start playing music. IT should retrieve the audio connection.

### What is actually happening?

Nothing. The Bot starts waiting for the audiConnection

### Any additional comments?

This are my Logs!
This is the startup and I try to connect the bot via my play command. After that It does not respond. And if I try another application command after that, it prints an error as you can see at the bottom.

```Trying to send GET request to https://discord.com/api/v10/gateway/bot
Sent GET request to https://discord.com/api/v10/gateway/bot and received status code 200 with body {"url": "wss://gateway.discord.gg", "shards": 1, "session_start_limit": {"total": 1000, "remaining": 989, "reset_after": 85278845, "max_concurrency": 1}}
Calculated an offset of -841 to the Discord time.
Creating 1 shard
Creating shard 1 of 1
Received HELLO packet
Queued lifecycle frame for sending with priority: WebSocketFrame(FIN=1,RSV1=0,RSV2=0,RSV3=0,Opcode=TEXT,Length=14,Payload="{"op":1,"d":0}")
Sending identify packet
Queued lifecycle frame for sending without priority: WebSocketFrame(FIN=1,RSV1=0,RSV2=0,RSV3=0,Opcode=TEXT,Length=214,Payload="{"op":2,"d":{"token":"Bot **********","compress":true,"large_threshold":250,"properties":{"os":"Mac OS X","browser":"Javacord","device":"Javacord"},"intents":65535}}")
[...]

PR #1223, opened 2023-03-26
Fix interactions in text-in-voice channels

Mysterypotatoguy:
## Checklist

## Changelog
- Fixed AssertionError when receiving interactions from text-in-voice channels

Closes: #1222 

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
@@ -3,6 +3,7 @@ package org.javacord.core.util.handler.interaction;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.logging.log4j.Logger;
 import org.javacord.api.DiscordApi;
+import org.javacord.api.entity.channel.ServerChannel;
 import org.javacord.api.entity.channel.TextChannel;
 import org.javacord.api.entity.message.component.ComponentType;
 import org.javacord.api.event.interaction.AutocompleteCreateEvent;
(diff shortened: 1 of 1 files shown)


javacord-core/src/main/java/org/javacord/core/util/handler/interaction/InteractionCreateHandler.java
```diff
@@ -64,7 +65,9 @@ public void handle(JsonNode packet) {
 
             // Check if this interaction comes from a guild or a DM
             if (packet.hasNonNull("guild_id")) {
-                channel = api.getTextChannelById(channelId).orElse(null);
+                channel = api.getServerChannelById(channelId)
+                        .flatMap(ServerChannel::asTextableRegularServerChannel)
+                        .orElse(null);
```
felldo, line 70: Makes it easier to debug next time
```suggestion
                        .orElseThrow(AssertionError::new);
```

felldo (changes requested):

felldo (commented):
Not exactly sure what I am doing wrong, but if I send a slash command in the voice channel as described in the issue and I do not get any error. And since every TextableRegularServerChannel is a TextChannel, we should be able to get the voice channel

Mysterypotatoguy:
Well that's super odd, I was able to reproduce it earlier and now I cannot 🤔
Perhaps this is some caching issue with text-in-voice channels instead?

felldo:
> Well that's super odd, I was able to reproduce it earlier and now I cannot 🤔 Perhaps this is some caching issue with text-in-voice channels instead?

This would be my first guess, that the channel is not even cached

RealMarvn:
Ok so I have other commands which try to get the voice channel by ID. Those work great most of the time. The only leak is when I try to get the voice channel via getConnectedVoiceChannel. It could be a cache issue. But I have never watched inside Javacord.

PR #1225, opened 2023-03-31
Add NSFW property to Application Command Updater

S3nS3IW00:
## Checklist

## Changelog
- Added NSFW property to Application Command Updater
- Check if defaultMemberPermission in ApplicationCommandImpl is null before accessing it

## Description
A recent PR added NSFW property to Application Command Builder, but it was missing from Updater, so this PR adds it to the Updater as well.

It also fixes a NullPointerException that I discovered accidentaly when I called ApplicationCommandImpl#isDisabledByDefault(), so I added a simple check before checking its emptiness.

Closes: NaN

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
@@ -124,7 +124,7 @@ public abstract class ApplicationCommandUpdater<T extends ApplicationCommand,
     }
 
     /**
-     * Sets the slash command as nsfw.
+     * Sets whether the command is nsfw.
      *
      * @param nsfw Whether the command is nsfw.
      * @return The current instance in order to chain call methods.
--- a/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
+++ b/javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
@@ -124,6 +124,17 @@ public abstract class ApplicationCommandUpdater<T extends ApplicationCommand,
     }
 
     /**
+     * Sets the slash command as nsfw.
+     *
+     * @param nsfw Whether the command is nsfw.
+     * @return The current instance in order to chain call methods.
+     */
+    public B setNsfw(boolean nsfw) {
+        delegate.setNsfw(nsfw);
+        return (B) this;
+    }
+
+    /**
      * Updates a global application command.
      * When used to update multiple global application commands at once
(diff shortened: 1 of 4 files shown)


javacord-api/src/main/java/org/javacord/api/interaction/ApplicationCommandUpdater.java
```diff
@@ -123,6 +123,17 @@ public B setEnabledInDms(boolean enabledInDms) {
         return (B) this;
     }
 
+    /**
+     * Sets the slash command as nsfw.
```
felldo, line 127:
```suggestion
     * Sets whether the command is nsfw.
```

javacord-api/src/main/java/org/javacord/api/interaction/internal/ApplicationCommandUpdaterDelegate.java
```diff
@@ -68,6 +68,13 @@
      */
     void setEnabledInDms(boolean enabledInDms);
 
+    /**
+     * Sets the slash command as nsfw.
```
felldo, line 72:
```suggestion
     * Sets whether the command is nsfw.
```

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandImpl.java
```diff
@@ -111,7 +111,7 @@ public Optional<EnumSet<PermissionType>> getDefaultRequiredPermissions() {
 
     @Override
     public boolean isDisabledByDefault() {
-        return defaultMemberPermission.isEmpty();
+        return defaultMemberPermission != null && defaultMemberPermission.isEmpty();
```
felldo, line 114: Can't mark it, but please revert this change and instead change the initial assignment of null to this variable to `EnumSet.noneOf(PermissionType.class)` (line 58)

javacord-core/src/main/java/org/javacord/core/interaction/ApplicationCommandUpdaterDelegateImpl.java
```diff
@@ -27,6 +27,8 @@
 
     protected Boolean dmPermission = null;
 
+    protected Boolean nsfw = false;
```
felldo, line 30: Set this to null and add a null check when adding this to the body like it's done with `dmPermission` this ensures in case Discord decides to change the default value (which is unlikely), that we are reflecting the change too.

felldo (changes requested):

felldo (approved):
Thank you, LGTM!

Issue #1227, opened 2023-04-03
Message.removeEmbed() doesn't do anything


memecodes:
### Javacord version

3,8,0

### Steps to reproduce

Send Message with embed, then use Message.removeEmbed().get();
No exception is thrown and the message embeds stay.

### What is expected?

Embeds from the message to be removed, or an error to throw

### What is actually happening?

Returns the same message, not removing any embeds.

### Any additional comments?

Thanks to vampire for helping me with it :# 
![image](https://user-images.githubusercontent.com/87757515/229581165-0695e267-95b4-4c7c-9366-d25f38a94a29.png)

PR #1228, opened 2023-04-03
Fix wrong implementation for Message#removeEmbeds

feddevanderlist:
## Checklist

## Changelog
- Fixed Message.removeEmbed() doesn't do anything #1227 

## Description
When calling the `Message.removeEmbeds()` nothing happened not even an exception/error.
Now the embeds get deleted when the method is called

Closes: #1227 

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/message/Message.java
@@ -509,7 +509,7 @@ public interface Message extends DiscordEntity, Deletable, Comparable<Message>, 
      * @return A future to check if the removal was successful.
      */
     default CompletableFuture<Message> removeEmbed() {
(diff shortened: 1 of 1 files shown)


felldo (commented):
Please write a more descriptive commit message and fill out the PR description properly (your checkboxes and a bullet point for the changelog), otherwise it will not be picked up by our changelog generator

felldo (commented):
Your commit message is still the same. Something like `Fix wrong implementation for Message#removeEmbeds` is sufficient. The changelog in the PR is only for the users. There, you write something like the first bullet point you did. The 2nd bullet points isn't really interesting for any user to read, as they just want to know what has been fixed and not how.

Vampire:
Nope, still what he said last, your commit message didn't change

feddevanderlist:
Sometimes I just need to read the right way :)
I am blaming it on that it's late

felldo (approved):

PR #1233, opened 2023-04-07
Improved the way to test your code

RealYusufIsmail:
## Checklist

## Changelog
- Added a sample class to make it easy to copy and test.
- Added a folder where the test class can be created and classes apart from SampleTestBot are ignored 

## Description

Closes: NaN

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -17,7 +17,7 @@ While there are no strict rules for commit messages, it is strongly recommended 
 When you create pull requests, make sure your changes work by testing them.
 If your PR is untested, please mark it as a draft.
 There are 2 easy ways to test your changes:
-1. Create a Test file in the Javacord project `./javacord-core/src/main/java/org/javacord/core/Testfile.java` and put your code in a main method(Be careful to **NOT** add the file to your VCS and commit it!). Then you just need to log in with a bot, and you can verify if your changes are working properly by using your changes.
+1. Create a Test class with any name in the Javacord core test folder.  `./javacord-core/src/test/java/org/javacord/testing`. A sample class can be seen in the class SampleTestBot in the same folder, please don't edit. Then you just need to log in with a bot, and you can verify if your changes are working properly by using your changes.
 2. Create a composite build with Gradle.
    1. Create a new Gradle project in the same directory as the Javacord project.
    2. Add `includeBuild("../Javacord")` to you `settings.gradle` in your created project.
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@
 
 # Other
 CHANGELOG.md
+
+# Test files
+/javacord-core/src/test/java/org/javacord/testing/
--- a/javacord-core/src/test/java/org/javacord/testing/SampleTestBot.java
+++ b/javacord-core/src/test/java/org/javacord/testing/SampleTestBot.java
@@ -1,0 +1,37 @@
+package org.javacord.testing;
+
+import org.javacord.api.DiscordApi;
+import org.javacord.api.DiscordApiBuilder;
+import org.javacord.api.entity.message.MessageFlag;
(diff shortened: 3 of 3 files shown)


felldo:
Did you test your recommended way to get started testing own changes?
When I do your what you wroteI get an exception 
`java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7`

RealYusufIsmail:
> Did you test your recommended way to get started testing own changes? When I do your what you wroteI get an exception `java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7`

That does not seem an issue with the test

RealYusufIsmail:
> Did you test your recommended way to get started testing own changes? When I do your what you wroteI get an exception `java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7`

can you tell me what I did

RealYusufIsmail:
> > I ran the build and it worked
> 
> I just did what you described. Copied the class in the same directory, right-clicked the file and executed it. Not sure what I would be doing wrong. To verify if it's only my fault, please try to execute a bot in the same way as I did.

I mean I don't have a bot token at hand right now

RealYusufIsmail:
but did you change the class name and set a token

felldo:
I also ran the normal tests and it also threw an error. For some reason it seems like gradle doesn't like running with Java 17 in Javacord. When I switched to 11 it worked fine. 
And regarding executing the test class, it seems suboptimal that the DEBUG logs are shown in the console, so the logger configuration has to be adjusted or better said seperated so each test directory has their own configuration.

RealYusufIsmail:
Will see if I am able to do the latter part of your message

RealYusufIsmail:
> I also ran the normal tests and it also threw an error. For some reason it seems like gradle doesn't like running with Java 17 in Javacord. When I switched to 11 it worked fine. 
> 
> And regarding executing the test class, it seems suboptimal that the DEBUG logs are shown in the console, so the logger configuration has to be adjusted or better said seperated so each test directory has their own configuration. 

Would it be alright to add a .xml file for logging like https://github.com/YDWK/YDWK/blob/master/src/test/resources/logback.xml in the test folder. It says what logging you want and where from.

felldo:
> Would it be alright to add a .xml file for logging like https://github.com/YDWK/YDWK/blob/master/src/test/resources/logback.xml in the test folder. It says what logging you want and where from.

We already have a logging file and this will most likely not work because you will also affect the logging level for the normal tests, that's why I said the configuration has to be separated somehow.

RealYusufIsmail:
> > Would it be alright to add a .xml file for logging like https://github.com/YDWK/YDWK/blob/master/src/test/resources/logback.xml in the test folder. It says what logging you want and where from.
> 
> 
> 
> We already have a logging file and this will most likely not work because you will also affect the logging level for the normal tests, that's why I said the configuration has to be separated somehow.
> 
> 

Can we leave that for a separate pr

felldo:
> Can we leave that for a separate pr

But then what's the point of this PR if it is not really suitable for getting started with testing? Then, I think using the current way to create a test class is easier than having to deal with the logging configuration.

RealYusufIsmail:
> > Can we leave that for a separate pr
> 
> But then what's the point of this PR if it is not really suitable for getting started with testing? Then, I think using the current way to create a test class is easier than having to deal with the logging configuration.

but this serves as a template for a test class and to know where to test. Makes it easer.

RealYusufIsmail:
> > > Can we leave that for a separate pr
> > 
> > 
> > But then what's the point of this PR if it is not really suitable for getting started with testing? Then, I think using the current way to create a test class is easier than having to deal with the logging configuration.
> 
> but this serves as a template for a test class and to know where to test. Makes it easer.

I could look at how to make the logging work

PR #1251, opened 2023-07-16
Fixed a cache corruption issue when receiving ephemeral messages in text-in-voice channels and introduce TextableRegularServerChannel as entity

bassner:
## Checklist

## Changelog
- Check for TextableRegularServerChannels instead of pure ServerTextChannels when receiving ephemeral messages
- Upgrade TextableRegularServerChannel to be a fully fledged intermediate entity, with builders and everything

### Breaking Changes
- Rename `ServerTextChannel#updateNsfwFlag` to `updateNswf`

## Description

When receiving an ephemeral messages, Javacord was checking if the channel is on a server rather than a DM by looking for ServerTextChannels and ServerThreadChannels. However, ephemeral messages can also be received in voice channels. As Javacord was not checking for that, it created a private channel in the cache with the ID of the voice channel, which corrupted the cache and made further interactions or messages impossible to handle for that specific channel.

Steps to reproduce without this PR:

1. Create a server voice channel
2. Let a bot send a message with a button into that channel.
3. Add a listener so that the bot responds with an ephemeral messages if the button is clicked.
4. Click the button and observe that it works for the first time.
5. Click the button again and see that it doesn't work anymore + see the exception in logs
6. Send a message as a user into the channel and see in the logs how Javacord fails to handle the message

This is potentially the actual fix for the problem of #1223.

Closes: #1222 
Replaces: #1223 


felldo (commented):
I saw that we currently use `getServerTextChannelById` in `WebhooksUpdateHandler` as well. As the voice channels can now have webhooks because of the text channel, would you mind changing this in this PR too?

bassner:
This actually turns out to not be that easy. I first have to convert `TextableRegularServerChannel` into a fully fledged entity sitting in between voice/text channels and `RegularServerChannel` in the entity hierarchy, with interfaces, updater etc. I'm looking into it but it's quite complex

javacord-core/src/main/java/org/javacord/core/entity/channel/TextableRegularServerChannelBuilderDelegateImpl.java
```diff
@@ -0,0 +1,63 @@
+package org.javacord.core.entity.channel;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.javacord.api.entity.channel.ChannelCategory;
+import org.javacord.api.entity.channel.internal.ServerTextChannelBuilderDelegate;
+import org.javacord.api.entity.channel.internal.TextableRegularServerChannelBuilderDelegate;
+import org.javacord.core.entity.server.ServerImpl;
+
+/**
+ * The implementation of {@link ServerTextChannelBuilderDelegate}.
```
felldo, line 10:
```suggestion
 * The implementation of {@link TextableRegularServerChannelBuilderDelegate}.
```

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerVoiceChannel.java
```diff
@@ -297,31 +297,6 @@ default CompletableFuture<Void> removeUserLimit() {
...
-     *
-     * @return A future to check if the update was successful.
-     */
-    default CompletableFuture<Void> removeCategory() {
-        return createUpdater().removeCategory().update();
-    }
-
     /**
      * Updates the nsfw flag of the channel.
      *
```
felldo, line 302: Same here, `updateNsfwFlag` should be in the `TextableRegularServerChannel`?

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerTextChannel.java
```diff
@@ -10,7 +10,7 @@
 /**
  * This class represents a server text channel.
  */
-public interface ServerTextChannel extends RegularServerChannel, TextableRegularServerChannel,
+public interface ServerTextChannel extends TextableRegularServerChannel,
         ServerTextChannelAttachableListenerManager {
 
     @Override
```
felldo, line 16: `getMentionTag` can be removed as it's already present in `TextableRegularServerChannel`

felldo (changes requested):
Because the `TextableRegularServerChannel` is now implemented like the other types, methods to the `DiscordApi` should be added to get the channel so you don't have to map the server channel like you do it now.

You put some more fields like the `parentId` in the `TextableRegularServerChannel`. Name wise I think this doesn't really fit in there. But for now I think it is okay

javacord-api/src/main/java/org/javacord/api/entity/channel/ServerTextChannel.java
RealMarvn: Is there anything new on this pr after your commit?

felldo: What do you mean?

felldo: This is not my PR and I just suggested some changes -> haven't commited anything

RealMarvn:
I still have issues testing it. There is still no #setCategory or #addPermissionOverwrite in for TextChannel builders.

bassner:
> I still have issues testing it. There is still no #setCategory or #addPermissionOverwrite in for TextChannel builders.

image

It works, tho. I guess there's still something wrong with importing this snapshot into your project

javacord-api/src/main/java/org/javacord/api/entity/channel/ChannelType.java
```diff
@@ -7,20 +7,20 @@
...
+    PRIVATE_CHANNEL(1, true, true, false, false, false),
+    SERVER_VOICE_CHANNEL(2, false, true, true, true, true),
+    GROUP_CHANNEL(3, true, true, false, false, false),
+    CHANNEL_CATEGORY(4, false, false, true, true, false),
+    SERVER_NEWS_CHANNEL(5, true, false, true, true, true),
+    SERVER_STORE_CHANNEL(6, true, false, true, true, true),
+    SERVER_NEWS_THREAD(10, true, false, true, false, false),
+    SERVER_PUBLIC_THREAD(11, true, false, true, false, false),
+    SERVER_PRIVATE_THREAD(12, true, false, true, false, false),
+    SERVER_STAGE_VOICE_CHANNEL(13, false, true, true, true, false),
```
felldo, line 20:
```suggestion
    SERVER_STAGE_VOICE_CHANNEL(13, false, true, true, true, true),
```

javacord-api/src/main/java/org/javacord/api/entity/channel/TextableRegularServerChannelBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+        this.delegate = delegate;
+    }
+
+    /**
+     * Sets the category of the channel.
+     *
+     * @param category The category of the channel.
+     * @return The current instance in order to chain call methods.
+     */
+    public TextableRegularServerChannelBuilder<T> setCategory(ChannelCategory category) {
```
felldo, line 33: The same for `setSlowmodeDelayInSeconds`
```suggestion
    public T setCategory(ChannelCategory category) {
```

javacord-api/src/main/java/org/javacord/api/entity/channel/TextableRegularServerChannelUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+        this(DelegateFactory.createTextableRegularServerChannelUpdaterDelegate(channel));
+    }
+
+    /**
+     * Queues the category to be updated.
+     *
+     * @param category The new category of the channel.
+     * @return The current instance in order to chain call methods.
+     */
+    public TextableRegularServerChannelUpdater<T> setCategory(ChannelCategory category) {
```
felldo, line 39: The same for the other methods
```suggestion
    public T setCategory(ChannelCategory category) {
```

felldo (changes requested):

felldo (approved):
LGTM!

PR #1253, opened 2023-08-01
Use sets instead of lists for user roles collections in server updater

bassner:
## Checklist

## Changelog
- replaced internal collection type for user roles in the server updater with sets

## Description

if you add the same role twice to the server updater, Javacord will currently send it twice, causing this error:

```
{
   "message":"Invalid Form Body",
   "code":50035,
   "errors":{
      "roles":{
         "2":{
            "_errors":[
               {
                  "code":"SET_TYPE_ALREADY_CONTAINS_VALUE",
                  "message":"The set already contains this value"
               }
            ]
         }
      }
   }
}
```

Closes: NaN

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/server/ServerUpdaterDelegateImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/server/ServerUpdaterDelegateImpl.java
@@ -37,6 +37,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 /**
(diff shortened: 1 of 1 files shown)


javacord-core/src/main/java/org/javacord/core/entity/server/ServerUpdaterDelegateImpl.java
```diff
@@ -516,31 +516,31 @@ public void reorderRoles(List<Role> roles) {
 
     @Override
     public void addRoleToUser(User user, Role role) {
-        Collection<Role> userRoles = this.userRoles.computeIfAbsent(user, u -> new ArrayList<>(server.getRoles(u)));
+        Collection<Role> userRoles = this.userRoles.computeIfAbsent(user, u -> new HashSet<>(server.getRoles(u)));
```
felldo, line 519:
```suggestion
        Set<Role> userRoles = this.userRoles.computeIfAbsent(user, u -> new HashSet<>(server.getRoles(u)));
```
^ Same for the methods below

The userRoles variable declaration can also be changed to 
`private final Map<User, Set<Role>> userRoles`

felldo (changes requested):

felldo (approved):
LGTM!
Please rebase into a single commit + merge your PR

PR #1259, opened 2023-08-18
Added a way to set bot activity and custom status when creating bot

RealYusufIsmail:
## Checklist

## Changelog
- Currently only activity can be set after bot has started.

## Description
Added ability to set activity as it starts.
```java
String token = "your token";

        DiscordApi api = new DiscordApiBuilder().setToken(token)
                .setActivity(BotActivity.playing("with Javacord")).login().join();

or for example
new DiscordApiBuilder().setToken(token)
                .setActivity(BotActivity.custom("with Javacord")).login().join();

or for example
new DiscordApiBuilder().setToken(token)
                .setActivity(BotActivity.playing("JavaCord", "in IDE")).login().join();

```

image

image

image

Closes: NaN

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
+++ b/javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
@@ -2,6 +2,7 @@ package org.javacord.core.util.gateway;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.neovisionaries.ws.client.ProxySettings;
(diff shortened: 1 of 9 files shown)


RealYusufIsmail:
image

RealYusufIsmail:
No idea why the main is failing throwing some weird error

RealYusufIsmail:
image

```java
public class Test {
        String token = "";

        new DiscordApiBuilder().setToken(token)
                .setActivity(BotActivity.custom("with Javacord")).login().join();
    }
}
```

RealYusufIsmail:
image


```java
public class Test {
        String token = "";

        new DiscordApiBuilder().setToken(token)
                .setActivity(BotActivity.playing("JavaCord", "in IDE")).login().join();
    }
}

```

felldo:
Imho this enhancement introduces a bad practice because when you set the activity on the builder, you will send `SHARDS` requests to set the activity even though only one request is required to set the activity for the whole application.
This is just my opionion on this and I would personally not merge this.

If this is something the community or other maintainers approve, I would be fine with this as it is not severe because it only happens on startup.

javacord-core/src/main/java/org/javacord/core/util/gateway/DiscordWebSocketAdapter.java
```diff
@@ -939,6 +966,7 @@ public void updateStatus() {
                 return Optional.of(type);
```
NoahvdAa, line 966: Shouldn't this check above be removed? Else you can only set the custom status at startup

RealYusufIsmail: Implemented the new changes but I need to test it

PR #1260, opened 2023-08-24
Migrated to new user name system

### LINKED ISSUE #1245: Replacement/just removal and deprecation of getDiscriminatedName
### What feature are you proposing?

`getDiscriminatedName` is no longer desirable as users who've migrated to the new username system will have no discriminator (most likely represented as 0000). I personally don't think changing the behaviour of the method is a good idea as the name implies that it will have a discriminator!

The method could have a replacement which drops #0000... however at some point both `getDiscriminatedName` and this method will become useless

### Do you have any additional material for your feature proposal?

_No response_

RealYusufIsmail:
## Checklist

## Changelog
- Add getDisplayName which can be null.
- Added its respective listener.

### Breaking Changes
```getDiscriminator``` is now Deprecated

## Description

Closes: #1245 

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java
@@ -48,6 +48,7 @@ public class UserImpl implements User, InternalUserAttachableListenerManager {
     private final Long id;
     private final String name;
     private final String discriminator;
+    private final String globalName;
     private final String avatarHash;
     private EnumSet<UserFlag> userFlags = EnumSet.noneOf(UserFlag.class);
     private final boolean bot;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: javacord-core/src/main/java/org/javacord/core/entity/user/UserImpl.java [hunk 1]
```diff
@@ -156,7 +160,12 @@ public UserImpl replacePartialUserData(JsonNode partialUserJson) {
             avatarHash = partialUserJson.get("avatar").asText();
         }
 
-        return new UserImpl(api, id, name, discriminator, avatarHash, bot, member);
+        String displayName = this.globalName;
+        if (partialUserJson.hasNonNull("global_name")) {
+            displayName = partialUserJson.get("global_name").asText();
+        }
```
wasdennnoch, line 166: `displayName` is never used. Did you mean to update `getDisplayName()`?

RealYusufIsmail: yeah my bad.

TheKodeToad:
I don't think we'll ever see discriminators fully be removed, so deprecating it is not a good idea

PR #1264, opened 2023-09-03
Fixed typos

StrajnarFilip:


javacord-api/src/main/java/org/javacord/api/util/event/ListenerManager.java
```diff
@@ -32,15 +32,15 @@ public interface ListenerManager<T> {
     T getListener();
 
     /**
-     * Gets the class of the object, the listener was added to.
+     * Gets class of the object the listener was added to.
```
Vampire, line 35: Removing the "the" is not correct.
And if removing the comma is correct, please also remove it in the `@returns`.

javacord-api/src/main/java/org/javacord/api/util/event/ListenerManager.java
```diff
@@ -32,15 +32,15 @@ public interface ListenerManager<T> {
...
      * For global listeners, it returns an empty {@code Optional}.
      *
      * @return The class of the object, the listener was added to.
      */
     Optional<Class<?>> getAssignedObjectClass();
 
     /**
-     * Gets the id of the object, the listener as added to.
+     * Gets id of the object the listener was added to.
```
Vampire, line 43: Same as above

Vampire (changes requested):

StrajnarFilip:
Thank you for your feedback, I'm creating another PR (branch `patch1-fixes`) with requested changes.

Vampire:
Why? o_O
Next time just force-push the branch after you improved the commit.

StrajnarFilip:
I'm really sorry for the inconvenience, I didn't know which way would be preferred. Note taken for potential future commits.
👌

Vampire:
I know no single project that would prefer closing a PR and creating a new one with requested changes. That doesn't make much sense, looses all context and review comments, and is a handling nightmare.

Issue #1268, opened 2023-09-09
Parameters are ignored when requesting archived threads


End45:
### Javacord version

3.8.0

### Steps to reproduce

- Create a channel with at least 3 public archived threads
- Call `ServerTextChannel.getPublicArchivedThreads(2)` to retrieve the 2 that were archived most recently

### What is expected?

The method returns 2 archived threads

### What is actually happening?

The method returns more than 2 threads

### Any additional comments?

Calls to the following methods have their parameters ignored and are equivalent to calling the parameterless version of the method:
- `ServerTextChannel.getPublicArchivedThreads(long before)`
- `ServerTextChannel.getPublicArchivedThreads(int limit)`
- `ServerTextChannel.getPublicArchivedThreads(Long before, Integer limit)`
- `ServerTextChannel.getPrivateArchivedThreads(long before)`
- `ServerTextChannel.getPrivateArchivedThreads(int limit)`
- `ServerTextChannel.getPrivateArchivedThreads(Long before, Integer limit)`
- `ServerTextChannel.getJoinedPrivateArchivedThreads(long before)`
- `ServerTextChannel.getJoinedPrivateArchivedThreads(int limit)`
- `ServerTextChannel.getJoinedPrivateArchivedThreads(Long before, Integer limit)`

Why?

If we check the API docs, the [API endpoints to retrieve archived threads](https://discord.com/developers/docs/resources/channel#list-public-archived-threads) take the following **query** parameters:
- before?	- ISO8601 timestamp - returns threads archived before this timestamp
- limit? - integer - optional maximum number of threads to return

However, the implementation of these methods places the following parameters on the request **body**:
- before - long - Get archived threads before the thread with this id
- limit - int - The maximum amount of archived threads

https://github.com/Javacord/Javacord/blob/b1ce43cd618fb1306e25929f63d6ab8958a33496/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerTextChannelImpl.java#L61-L74

Not only the first parameter has the wrong format, but they are ignored by the API since they are not query parameters. This results in the Discord API replying with the default response, which lists the most X recent public/private archived threads (I don't know what the exact amount is).

I tried moving the parameters to the querystring while debugging the problem, with code like this:
```java
new RestRequest<ArchivedThreads>(getApi(), RestMethod.GET, RestEndpoint.LIST_PUBLIC_ARCHIVED_THREADS)
    .setUrlParameters(getIdAsString())
[...]

PR #1269, opened 2023-09-09
Fix archived thread getters

End45:
## Checklist

### Breaking Changes
- Fixed not working parameters for `getPublicArchivedThreads`, `getPrivateArchivedThreads`, `getJoinedPrivateArchivedThreads` and changed from requiring a channel ID to an Instant

## Description
This PR changes the signature and behavior of the following methods:
- `ServerTextChannel.getPublicArchivedThreads(long before)`
- `ServerTextChannel.getPublicArchivedThreads(Long before, Integer limit)`
- `ServerTextChannel.getPrivateArchivedThreads(long before)`
- `ServerTextChannel.getPrivateArchivedThreads(Long before, Integer limit)`

It also changes the behavior of the following methods:
- `ServerTextChannel.getPublicArchivedThreads(int limit)`
- `ServerTextChannel.getPrivateArchivedThreads(int limit)`
- `ServerTextChannel.getJoinedPrivateArchivedThreads(long before)`
- `ServerTextChannel.getJoinedPrivateArchivedThreads(int limit)`
- `ServerTextChannel.getJoinedPrivateArchivedThreads(Long before, Integer limit)`

Although they never worked properly in the first place, so I don't know if that counts.

Closes: #1268

### DIFF
--- a/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerTextChannelImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/channel/ServerTextChannelImpl.java
@@ -13,6 +13,9 @@ import org.javacord.core.util.rest.RestMethod;
 import org.javacord.core.util.rest.RestRequest;
 
 import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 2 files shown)


javacord-core/src/main/java/org/javacord/core/entity/channel/ServerTextChannelImpl.java
```diff
@@ -59,51 +58,47 @@ public void setTopic(String topic) {
 
     @Override
     public CompletableFuture<ArchivedThreads> getPublicArchivedThreads(Long before, Integer limit) {
```
felldo, line 60: Because this is not a channel id anymore, but a timestamp, this has to be changed too. Use an `Instant` as a parameter. See https://github.com/Javacord/Javacord/blob/b1ce43cd618fb1306e25929f63d6ab8958a33496/javacord-core/src/main/java/org/javacord/core/entity/server/ServerUpdaterDelegateImpl.java#L613-L619 on how to convert this to the appropriate format. This applied to all affected methods

felldo (changes requested):

felldo (approved):
Thanks for the PR, LGTM! :)

felldo:
Btw. the changelog in the PR description is our actual changelog. Write something (which I already did) which we can use in our changelog to quickly inform users about the changes.

felldo:
@all-contributors please add @End45 to code

PR #1273, opened 2023-09-14
Fix race condition where task queue might be empty and remove now unnecessary null-check

Vampire:
## Checklist


wasdennnoch (approved):

Vampire:
This should also fix some event scheduling stalls .

If you look at
https://github.com/Javacord/Javacord/blob/2e32d382c154bdbf78f5cc4757244dd713b3d402/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java#L254-L257
There the `runningListeners.isEmpty()` check will always lead to the `return` if the early-out from the removed `null` check kicked in, as also the `runningListeners.remove(finalQueueSelector)` at
https://github.com/Javacord/Javacord/blob/master/javacord-core/src/main/java/org/javacord/core/util/event/EventDispatcherBase.java#L299
was skipped.

felldo (approved):
I can confirm that this PR fixes the race condition effectively causing a NPE, when polling the queue. This bug also caused a memory leak, where the queue would fill up with all events but never dispatching and thus removing them.

PR #1285, opened 2023-10-20
Expose User ID in AuditLogEntry

wasdennnoch:
## Checklist

## Changelog
- Added getter for User ID in AuditLogEntry

## Description
Previously, only `getUser()` was exposed, which fetches the full user object from Discord. If you only need to know the user's ID, for example to check if it's your own client ID, this would result in an unnecessary network call. So let's expose the raw ID!

Closes: NaN


felldo (approved):

felldo:
Please rebase your branch :)

wasdennnoch:
Rebased 👍

Issue #1249, opened 2023-07-13
Add the message author id to message reaction events


felldo:
### Do you have any additional material for your feature proposal?

https://github.com/discord/discord-api-docs/pull/6102
https://github.com/discord/discord-api-docs/pull/6296

kezc:
I'll work on this!

PR #1289, opened 2023-11-11
Add message author id to ReactionAddEvent 

kezc:
## Checklist

## Changelog
- Added message author id to message reaction event

## Description

Closes: https://github.com/Javacord/Javacord/issues/1249

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
+++ b/javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
@@ -1,5 +1,6 @@
 package org.javacord.api.event.message.reaction;
 
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -19,6 +20,6 @@ public interface ReactionAddEvent extends SingleReactionEvent {
      *
      * @return The message author id.
      */
-    long getMessageAuthorId();
+    Optional<Long> getMessageAuthorId();
 
 }
--- a/javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
+++ b/javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
@@ -14,4 +14,11 @@ public interface ReactionAddEvent extends SingleReactionEvent {
      */
     CompletableFuture<Void> removeReaction();
 
+    /**
+     * Gets the message author id.
+     *
+     * @return The message author id.
+     */
+    long getMessageAuthorId();
+
 }
--- a/javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
@@ -15,6 +15,11 @@ import java.util.concurrent.CompletableFuture;
 public class ReactionAddEventImpl extends SingleReactionEventImpl implements ReactionAddEvent {
 
     /**
+     * The id of the user that sent the message.
+     */
+    private final long messageAuthorId;
+
+    /**
      * Creates a new reaction add event.
      *
      * @param api The discord api instance.
@@ -23,10 +28,12 @@ public class ReactionAddEventImpl extends SingleReactionEventImpl implements Rea
      * @param emoji The emoji.
      * @param userId The id of the user who added the reaction.
      * @param member The member if it happened in a server.
+     * @param messageAuthorId The id of the user that sent the message
      */
-    public ReactionAddEventImpl(
-            DiscordApi api, long messageId, TextChannel channel, Emoji emoji, long userId, Member member) {
+    public ReactionAddEventImpl(DiscordApi api, long messageId, TextChannel channel, Emoji emoji,
+                                long userId, Member member, long messageAuthorId) {
         super(api, messageId, channel, emoji, userId);
+        this.messageAuthorId = messageAuthorId;
     }
 
     @Override
@@ -34,4 +41,8 @@ public class ReactionAddEventImpl extends SingleReactionEventImpl implements Rea
         return Reaction.removeUser(getApi(), getChannel().getId(), getMessageId(), getEmoji(), getUserId());
     }
 
+    @Override
+    public long getMessageAuthorId() {
+        return messageAuthorId;
+    }
 }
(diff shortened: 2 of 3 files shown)


javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
```diff
@@ -14,4 +14,11 @@ public interface ReactionAddEvent extends SingleReactionEvent {
      */
     CompletableFuture<Void> removeReaction();
 
+    /**
+     * Gets the message author id.
+     *
+     * @return The reaction.
```
RealYusufIsmail, line 20:
```suggestion
     * @return The message author id.
```

javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
```diff
@@ -23,15 +28,25 @@ public class ReactionAddEventImpl extends SingleReactionEventImpl implements Rea
...
     @Override
     public CompletableFuture<Void> removeReaction() {
         return Reaction.removeUser(getApi(), getChannel().getId(), getMessageId(), getEmoji(), getUserId());
     }
 
+    /**
+     * Returns the id of the user that sent the message.
+     *
+     * @return The id of the user that sent the message.
+     */
```
RealYusufIsmail, line 48: Java doc is jot required in impls

RealYusufIsmail (changes requested):

javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
Vampire: That's non-sense.
JavaDoc always makes sense everywhere.
It makes most sense or should always be there on public methods in the public API.
But JavaDoc is always and everywhere better to be there than not to be there.
Thus you should really not motivate people to leave it out, especially when they already added it.

RealYusufIsmail: Ok I apologia

kezc: Anyway, thanks @RealYusufIsmail  for pointing it out as I was forced to add it by linter and I've started wondering why. Now I see that `@Override` annotation is missing.

Vampire: Ah, sorry @RealYusufIsmail, I misread your comment as "you don't need to add JavaDoc to core classes", which are our "implementation" of the logic.

That you don't need it on methods that implement an interface if you properly inherit the JavaDoc is true of course.

kezc: Ok, I've removed the unnecessary JavaDoc.

RealYusufIsmail: > classes

Not a problem.

RealYusufIsmail (approved):
LGTM. Please squash you commits and one of the contrubters will merge your pr if they approve.

javacord-core/src/main/java/org/javacord/core/util/handler/message/reaction/MessageReactionAddHandler.java
```diff
@@ -45,6 +45,7 @@ public void handle(JsonNode packet) {
         long channelId = packet.get("channel_id").asLong();
         long messageId = packet.get("message_id").asLong();
         long userId = packet.get("user_id").asLong();
+        long messageAuthorId = packet.get("message_author_id").asLong();
```
felldo, line 48:
```suggestion
        Long messageAuthorId = packet.hasNonNull("message_author_id") ? packet.get("message_author_id").asLong() : null;
```

javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
```diff
@@ -14,6 +14,11 @@
  */
 public class ReactionAddEventImpl extends SingleReactionEventImpl implements ReactionAddEvent {
 
+    /**
+     * The id of the user that sent the message.
+     */
+    private final long messageAuthorId;
```
felldo, line 20:
```suggestion
    private final Long messageAuthorId;
```

javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
```diff
@@ -23,15 +28,21 @@ public class ReactionAddEventImpl extends SingleReactionEventImpl implements Rea
      * @param emoji The emoji.
      * @param userId The id of the user who added the reaction.
      * @param member The member if it happened in a server.
+     * @param messageAuthorId The id of the user that sent the message
      */
-    public ReactionAddEventImpl(
-            DiscordApi api, long messageId, TextChannel channel, Emoji emoji, long userId, Member member) {
+    public ReactionAddEventImpl(DiscordApi api, long messageId, TextChannel channel, Emoji emoji,
+                                long userId, Member member, long messageAuthorId) {
```
felldo, line 34:
```suggestion
                                long userId, Member member, Long messageAuthorId) {
```

javacord-core/src/main/java/org/javacord/core/event/message/reaction/ReactionAddEventImpl.java
```diff
@@ -23,15 +28,21 @@ public class ReactionAddEventImpl extends SingleReactionEventImpl implements Rea
...
     }
 
     @Override
     public CompletableFuture<Void> removeReaction() {
         return Reaction.removeUser(getApi(), getChannel().getId(), getMessageId(), getEmoji(), getUserId());
     }
 
+    @Override
+    public long getMessageAuthorId() {
+        return messageAuthorId;
```
felldo, line 46:
```suggestion
    public Optional<Long> getMessageAuthorId() {
        return Optional.ofNullable(messageAuthorId);
```

javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
```diff
@@ -14,4 +14,11 @@ public interface ReactionAddEvent extends SingleReactionEvent {
      */
     CompletableFuture<Void> removeReaction();
 
+    /**
+     * Gets the message author id.
+     *
+     * @return The message author id.
+     */
+    long getMessageAuthorId();
```
felldo, line 22:
```suggestion
    Optional<Long> getMessageAuthorId();
```
This is an optional field and should return an `Optional<Long>`

felldo (changes requested):

javacord-api/src/main/java/org/javacord/api/event/message/reaction/ReactionAddEvent.java
kezc: Good point, thank you!

felldo (approved):
LGTM! Thanks!

Issue #1248, opened 2023-07-13
Add attachment flags


felldo:
### Do you have any additional material for your feature proposal?

https://github.com/discord/discord-api-docs/pull/6272

mikolaj-skrzypczak:
Issued a pull request :)

PR #1292, opened 2023-11-16
Add attachment flags

mikolaj-skrzypczak:
## Checklist

## Changelog
- Added attachment flags

### Breaking Changes
None

## Description

Closes: #1248

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlag.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlag.java
@@ -1,0 +1,49 @@
+package org.javacord.api.entity;
+
+public enum AttachmentFlag {
+    /**
+     * This attachment has been edited using the remix feature on mobile.
+     */
+    IS_REMIX(1 << 2),
+
+    /**
+     * The unknown message flag type.
+     */
+    UNKNOWN(-1);
+
+    private final int id;
+
+    /**
+     * Creates a new attachment flags type.
+     *
+     * @param id attachment flags id.
+     */
+    AttachmentFlag(final int id) {
+        this.id = id;
+    }
+
+    /**
+     * Gets the id of the attachment flags type.
+     *
+     * @return The id of the attachment flags type.
+     */
+    public int getId() {
+        return this.id;
+    }
+
+    /**
+     * Gets the attashment flags type by its id.
+     *
+     * @param id The id of the attachment flag type.
+     * @return The attachment flag type with the given id or {@link AttachmentFlag#UNKNOWN} if unknown id.
+     */
+    public static AttachmentFlag getFlagTypeById(final int id) {
+        for (final AttachmentFlag value : values()) {
+            if (value.getId() == id) {
+                return value;
+            }
+        }
+        return UNKNOWN;
+    }
+
+}
--- a/javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
+++ b/javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.logging.log4j.Logger;
 import org.javacord.api.DiscordApi;
 import org.javacord.api.entity.Attachment;
+import org.javacord.api.entity.AttachmentFlag;
 import org.javacord.api.entity.DiscordEntity;
 import org.javacord.core.util.FileContainer;
 import org.javacord.core.util.logging.LoggerUtil;
@@ -16,6 +17,7 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Base64;
+import java.util.EnumSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
@@ -76,7 +78,7 @@ public class AttachmentImpl implements Attachment {
      * Whether this attachment is ephemeral.
      */
     private final Boolean ephemeral;
-    
+
     /**
      * The duration of the audio file.
      */
@@ -86,6 +88,11 @@ public class AttachmentImpl implements Attachment {
      * The base64 encoded bytearray representing a sampled waveform.
      */
     private final String waveform;
+
+    /**
+     * Attachment flags computed from the bitfield.
+     */
+    private final EnumSet<AttachmentFlag> attachmentFlags = EnumSet.noneOf(AttachmentFlag.class);
 
     /**
      * Creates a new attachment.
@@ -106,6 +113,14 @@ public class AttachmentImpl implements Attachment {
         ephemeral = data.hasNonNull("ephemeral") ? data.get("ephemeral").asBoolean() : null;
         durationSeconds = data.hasNonNull("duration_secs") ? data.get("duration_secs").doubleValue() : null;
         waveform = data.hasNonNull("waveform") ? data.get("waveform").asText() : null;
+        if (data.has("flags")) {
+            int flags = data.get("flags").asInt();
+            for (AttachmentFlag flag : AttachmentFlag.values()) {
+                if ((flag.getId() & flags) == flag.getId()) {
+                    attachmentFlags.add(flag);
+                }
+            }
+        }
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -106,6 +112,8 @@ public AttachmentImpl(DiscordApi api, final JsonNode data) {
         ephemeral = data.hasNonNull("ephemeral") ? data.get("ephemeral").asBoolean() : null;
         durationSeconds = data.hasNonNull("duration_secs") ? data.get("duration_secs").doubleValue() : null;
         waveform = data.hasNonNull("waveform") ? data.get("waveform").asText() : null;
+        Integer bitfield = data.hasNonNull("flags") ? data.get("flags").asInt() : null;
+        flags = bitfield != null ? AttachmentFlags.getFlagTypeByBitfield(bitfield) : null;
```
RealYusufIsmail, line 116: can these not be combined together, rather than having a separate variable

mikolaj-skrzypczak: I guess I could change it to:
```java
flags = data.hasNonNull("flags")
                ? AttachmentFlags.getFlagTypeByBitfield(data.get("flags").asInt())
                : null;
```
Would that be a satisfying solution for you?

RealYusufIsmail: yeah that works

RealYusufIsmail (approved):
LGTM. Please squash your commits then wait for a review so they can check then merge

mikolaj-skrzypczak:
Is there any chance for those changes to be merged any time soon? I kind of need it to pass one task in one of the courses in my studies 😅

javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlags.java
```diff
@@ -0,0 +1,49 @@
...
+     * This attachment has been edited using the remix feature on mobile.
+     */
+    IS_REMIX(1 << 2),
+
+    /**
+     * The unknown message flag type.
+     */
+    UNKNOWN(-1);
+
+    private final int bitfield;
```
felldo, line 14: This is not a bitfield, but only the value of the flag. The bitfield would be something like `010` where the `1` corresponds to a specific flag like `IS_REMIX`. And `0` would mean that another potentual flag is not applied for this attachment. So each single digit corresponds alway to one flag.
So please rename it and everywhere else in your newly added code :)
```suggestion
    private final int id;
```

javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlags.java
```diff
@@ -0,0 +1,49 @@
...
+    AttachmentFlags(final int bitfield) {
+        this.bitfield = bitfield;
+    }
+
+    /**
+     * Gets the bitfield of the attachment flags type.
+     *
+     * @return The bitfield of the attachment flags type.
+     */
+    public int getBitfield() {
```
felldo, line 30:
```suggestion
    public int getId() {
```

javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlags.java
```diff
@@ -0,0 +1,49 @@
...
+        return this.bitfield;
+    }
+
+    /**
+     * Gets the attashment flags type by its bitfield.
+     *
+     * @param bitfield The bitfield of the attachment flag type.
+     * @return The attachment flag type with the given id or {@link AttachmentFlags#UNKNOWN} if unknown bitfield.
+     */
+    public static AttachmentFlags getFlagTypeByBitfield(final int bitfield) {
```
felldo, line 40:
```suggestion
    public static AttachmentFlags getFlagTypeById(final int id) {
```

javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -106,6 +112,9 @@ public AttachmentImpl(DiscordApi api, final JsonNode data) {
         ephemeral = data.hasNonNull("ephemeral") ? data.get("ephemeral").asBoolean() : null;
         durationSeconds = data.hasNonNull("duration_secs") ? data.get("duration_secs").doubleValue() : null;
         waveform = data.hasNonNull("waveform") ? data.get("waveform").asText() : null;
+        flags = data.hasNonNull("flags")
+                ? AttachmentFlags.getFlagTypeByBitfield(data.get("flags").asInt())
+                : null;
```
felldo, line 117: This is not the correct way to get flags. Please have a look at this, on how to get the flags from a bitfield
https://github.com/Javacord/Javacord/blob/971ea44e5d671d380382e0617fcaeaad070c8d02/javacord-core/src/main/java/org/javacord/core/entity/activity/ActivityImpl.java#L82-L89

javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlags.java
```diff
@@ -0,0 +1,49 @@
+package org.javacord.api.entity;
+
+public enum AttachmentFlags {
```
felldo, line 3:
```suggestion
public enum AttachmentFlag {
```

javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -87,6 +88,11 @@ public class AttachmentImpl implements Attachment {
      */
     private final String waveform;
 
+    /**
+     * Attachment flags combined as a bitfield.
+     */
+    private final AttachmentFlags flags;
```
felldo, line 94:
```suggestion
    private final EnumSet<AttachmentFlag> attachmentFlags = EnumSet.noneOf(AttachmentFlag.class);
```

javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -87,6 +88,11 @@ public class AttachmentImpl implements Attachment {
      */
     private final String waveform;
 
+    /**
+     * Attachment flags combined as a bitfield.
```
felldo, line 92:
```suggestion
     * Attachment flags computed from the bitfield.
```

javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
```diff
@@ -183,6 +192,11 @@ public Optional<String> getWaveFormBase64() {
         return Optional.ofNullable(waveform);
     }
 
+    @Override
+    public Optional<AttachmentFlags> getFlags() {
```
felldo, line 196:
```suggestion
    public Optional<EnumSet<AttachmentFlag>> getFlags() {
```

felldo (changes requested):
Sorry for the late response, but here we go! I've outlined some issues with your code which would need to be fixed first before we can merge your PR :)

javacord-core/src/main/java/org/javacord/core/entity/AttachmentImpl.java
Mysterypotatoguy: I think this can just be an `EnumSet<AttachmentFlag>` with an empty set when there are no flags

felldo: > I think this can just be an `EnumSet<AttachmentFlag>` with an empty set when there are no flags

You are right. In this case it wouldn't make sense to return an optional, as there is no information that we loose when we remove it

mikolaj-skrzypczak:
@felldo thank you for the review :)
I think that my latest commit adressed the highlighted issues

felldo (changes requested):

javacord-api/src/main/java/org/javacord/api/entity/AttachmentFlag.java
```diff
@@ -0,0 +1,49 @@
...
+     * The unknown message flag type.
+     */
+    UNKNOWN(-1);
+
+    private final int id;
+
+    /**
+     * Creates a new attachment flags type.
+     *
+     * @param id attachment flags combined as a bitfield.
```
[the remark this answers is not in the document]
mikolaj-skrzypczak: 🤦, fixed

felldo (approved):
LGTM! Thanks!

Issue #527, opened 2020-04-21
Default allowed metions


Bastian:
As a follow-up of https://github.com/Javacord/Javacord/pull/525 we should add an option to set a default `AllowedMentions` object, e.g. if a bot should never mention `@everyone`.

PR #1293, opened 2023-12-19
Default allowed mentions

TheKodeToad:
## Checklist

## Changelog
- Added default allowed mentions to the `DiscordApi`

## Description
I used a slightly hacky approach of passing around DiscordApi. I hope it's okay!

Closes: #527

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
+++ b/javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
@@ -1,6 +1,8 @@
 package org.javacord.api;
 
 import org.javacord.api.entity.intent.Intent;
+import org.javacord.api.entity.message.MessageBuilder;
+import org.javacord.api.entity.message.mention.AllowedMentions;
 import org.javacord.api.event.server.ServerBecomesAvailableEvent;
 import org.javacord.api.internal.DiscordApiBuilderDelegate;
 import org.javacord.api.listener.ChainableGloballyAttachableListenerManager;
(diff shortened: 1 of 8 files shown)


TheKodeToad:
I tested a bot but I didn't test `./gradlew build` :)

javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
```diff
@@ -450,6 +452,28 @@ public DiscordApiBuilder setAllIntentsWhere(Predicate<Intent> condition) {
...
+
+    /**
+     * Controls who will be mentioned if mentions exist in a message.
+     * This is overriden with {@link MessageBuilder#setAllowedMentions(AllowedMentions)}
+     *
+     * @return The mention object.
+     */
+    public Optional<AllowedMentions> getAllowedMentions() {
+        return delegate.getAllowedMentions();
+    }
```
felldo, line 475: We usually don't have getters in the `DiscordApiBuilder`
```suggestion
```

javacord-api/src/main/java/org/javacord/api/internal/DiscordApiBuilderDelegate.java
```diff
@@ -177,6 +179,22 @@ public interface DiscordApiBuilderDelegate {
...
+    void setAllowedMentions(AllowedMentions allowedMentions);
+
+    /**
+     * Controls who will be mentioned if mentions exist in a message.
+     * This is overriden with {@link MessageBuilder#setAllowedMentions(AllowedMentions)}
+     *
+     * @return The mention object.
+     */
+    Optional<AllowedMentions> getAllowedMentions();
+
```
felldo, line 197:
```suggestion
```

felldo (changes requested):
Looks good so far, but:
* Please rename every field/method/Javadoc so it says `defaultAllowedMentions`.
* You have added the DiscordApi to a few methods, please add it as the first parameter

javacord-api/src/main/java/org/javacord/api/DiscordApiBuilder.java
TheKodeToad: There are getters for e.g. bot token in the builder 🤔

felldo: There are a few exceptions for like the bot token you mentioned or also the total shards. But if you look at the other 30 setters, none of them has getters. Not sure about the exact use case of the current getters, but I doubt getting the allowed mentioned on the builder instance is going to provide any benefits

TheKodeToad:
Hmm, the constructor which doesn't use a websocket doesn't have a defaultAllowedMentions parameter... is that an issue? It's internal anyway, right?

felldo:
> Hmm, the constructor which doesn't use a websocket doesn't have a defaultAllowedMentions parameter... is that an issue? It's internal anyway, right?

What do you mean? Can you show me which line?

TheKodeToad:
> What do you mean? Can you show me which line?

https://github.com/TheKodeToad/Javacord/blob/default-allowed-mentions/javacord-core/src/main/java/org/javacord/core/DiscordApiImpl.java#L422

felldo:
There are a lot of other things the constructor doesn't have, so it's okay

TheKodeToad:
This PR should be ready then! hopefully I didn't miss anything...

felldo (approved):
LGTM! Thanks!

PR #1301, opened 2024-03-19
Add Bulk Ban Support

wasdennnoch:
## Checklist

## Changelog
- Added bulk ban support
- Fixed NPE in default ban method

## Description

- https://github.com/discord/discord-api-docs/pull/6720
- Also fixes an NPE when calling `Server#banUser(String)`; this method would previously pass `null` as the `Duration` to the other overloads, which would then attempt to call `duration.getSeconds()`.

I'm not 100% confident that the `BulkBanResponse` I used is the way you generally want to approach such cases where an endpoint returns a special object that doesn't occur anywhere else, and I also wasn't quite sure where to put it; for now, I've put it next to the `Ban`  classes representing a single ban. Feedback appreciated.

Closes: NaN

### DIFF
--- a/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
+++ b/javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
@@ -67,6 +67,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import java.util.stream.LongStream;
 
 /**
  * The class represents a Discord server, sometimes also called guild.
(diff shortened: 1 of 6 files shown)


javacord-core/src/main/java/org/javacord/core/entity/server/BulkBanResponseImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+public class BulkBanResponseImpl implements BulkBanResponse {
+
+    /**
+     * The user ids that were banned successfully.
+     */
+    private final Set<Long> bannedUserIds = new LinkedHashSet<>();
```
felldo, line 14: Is there a specific reason you used a `LinkedHashSet`?

javacord-core/src/main/java/org/javacord/core/entity/server/BulkBanResponseImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Collection<Long> getBannedUserIds() {
+        return bannedUserIds;
+    }
+
+    @Override
+    public Collection<Long> getFailedUserIds() {
+        return failedUserIds;
+    }
```
felldo, line 39: Either you should return a unmodifiable copy of the collection or you assign an unmodifiable collection right away to the class fields

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2282,6 +2283,55 @@ default CompletableFuture<Void> banUser(long userId, Duration duration, String r
...
 
+    /**
+     * Bans the given user from the server.
+     *
+     * @param userIds The ids of the users to ban.
+     * @return A future returning details about the bulk ban.
+     */
+    default CompletableFuture<BulkBanResponse> bulkBanUsers(long... userIds) {
+        return bulkBanUsers(
+                LongStream.of(userIds).boxed().map(Long::toUnsignedString).collect(Collectors.toList()),
```
felldo, line 2294:
```suggestion
                LongStream.of(userIds).mapToObj(Long::toUnsignedString).collect(Collectors.toList()),
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2282,6 +2283,55 @@ default CompletableFuture<Void> banUser(long userId, Duration duration, String r
...
+    }
+
+    /**
+     * Bans the given user from the server.
+     *
+     * @param userIds The ids of the users to ban.
+     * @return A future returning details about the bulk ban.
+     */
+    default CompletableFuture<BulkBanResponse> bulkBanUsers(String... userIds) {
+        return bulkBanUsers(Arrays.stream(userIds).collect(Collectors.toList()), 0, TimeUnit.SECONDS, null);
```
felldo, line 2305: Seems a bit overly complex? :D
```suggestion
        return bulkBanUsers(Arrays.asList(userIds), 0, TimeUnit.SECONDS, null);
```

javacord-api/src/main/java/org/javacord/api/entity/server/Server.java
```diff
@@ -2282,6 +2283,55 @@ default CompletableFuture<Void> banUser(long userId, Duration duration, String r
         return banUser(Long.toUnsignedString(userId), duration, reason);
     }
 
+    /**
+     * Bans the given user from the server.
```
felldo, line 2287: Change this and the Javadoc from the other newly added methods
```suggestion
     * Bans the given users from the server.
```

felldo (changes requested):
You missed to add the status code
image

javacord-core/src/main/java/org/javacord/core/entity/server/BulkBanResponseImpl.java
wasdennnoch: I used it to preserve the order the API returns the IDs in. It was more of a potential nice-to-have and probably doesn't really matter, the API afaik just returns them in the same order you sent them in. But to be honest I also don't really see the use case for order preservation, so I've changed it to a normal HashSet for a tiny bit of performance gain.

